PR #14, opened 2024-03-21
hook up validation::block_connected tracepoint

0xB10C:
https://github.com/bitcoin/bitcoin/blob/master/doc/tracing.md#context-validation


0xB10C:
There seems to be an issue with the duration.

0xB10C:
> There seems to be an issue with the duration.

See also: https://github.com/bitcoin/bitcoin/pull/29877

I'm assuming a 25.x or older release to be used - i.e. passing nanoseconds. Internally, we convert to microseconds as the nanosecond precision isn't really needed.

PR #22, opened 2024-06-20
fix: avoid stack overflows with "huge" P2P messages

0xB10C:
On some systems, allocating an array for a "HUGE" P2P message (4 MB) payload failed due to it being larger than the rust default stack size of 2 MB.

To fix this, use a vector (allocated on the heap) instead of an array. This also allows some memory savings for "huge" messages that aren't actually 4MB in size as we can use the metadata size instead.

A test for huge messages is introduced to add test coverage for this bug.


0xB10C:
I've tested it on the affected system. Merging.

PR #29, opened 2024-08-08
remove rlimit bumping in extractor

0xB10C:
As noticed by @i-am-yuvi recently, bumping the rlimit is not required (anymore).

It turns out that, since libbpf commit https://github.com/libbpf/libbpf/commit/216eaa760edf06f3521d5d49a419515607c1f38e, the rlimit bump in downstream applications isn't needed anymore as libbpf handles it, if needed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -699,7 +699,6 @@ version = "0.1.0"
 dependencies = [
  "libbpf-cargo",
  "libbpf-rs",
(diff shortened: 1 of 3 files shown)


i-am-yuvi:
I don't see any problem with the rlimit. It's been more than 20 hours that peer-observer is running without any issue!

PR #30, opened 2024-08-08
Use clap for command line arguments

0xB10C:
Fixes #28.

The clap-based command line arguments not provide a `--help`, `--version`, and usage hints. It's also explicit what's needed and what's used as default.

For example, the extractor help now shows:

```
$ extractor --help
The peer-observer extractor Hooks into a Bitcoin Core binary with tracepoints and extracts events from it into a nanomsg PUB-SUB queue

Usage: extractor [OPTIONS] --bitcoind-path <BITCOIND_PATH>

Options:
  -a, --address <ADDRESS>              TCP socket the nanomsg publisher binds on [default: tcp://127.0.0.1:8883]
  -b, --bitcoind-path <BITCOIND_PATH>  Path to the Bitcoin Core (bitcoind) binary that should be hooked into
      --no-p2pmsg-tracepoints          Controls if the p2p message tracepoints should be hooked into
      --no-connection-tracepoints      Controls if the connection tracepoints should be hooked into
      --no-mempool-tracepoints         Controls if the mempool tracepoints should be hooked into
      --no-validation-tracepoints      Controls if the validation tracepoints should be hooked into
      --addrman-tracepoints            Controls if the addrman tracepoints should be hooked into. These may not have been PRed to Bitcoin Core yet
  -h, --help                           Print help
  -V, --version                        Print version
```

PR #45, opened 2024-08-26
Spy detector in Peer-observer

### LINKED ISSUE #43: Automaticall Detecting Spy Peers
Currently, there is no feature to detect spy peers/nodes, which has been discussed [here](https://github.com/0xB10C/peer-observer/issues/13#issuecomment-1999385505).

Some important anomalies to consider are:

- When INV is sent to the peer, but the peer doesn't send GETDATA to our node
- Spy nodes never send us INV for transactions we sent

We only handle INV, GETDATA, and TX p2p messages. We need to maintain a shared state – one entry (IP address + Port) for each connection with the number of INV, GETDATA, and TX sent and received. Additionally, spy peers will close the connection, so we also need to account for handling closed connections. Once all this is implemented, we can have some stats on normal/spy peers/nodes.

One approach could be to find the ratio of INV/GETDATA for each peer, but there might be other heuristics to detect peer identity.

A reference implementation has been done at the following URL: https://github.com/i-am-yuvi/peer-observer/tree/spy-detection.

i-am-yuvi:
Fixes #43 

This creates a tool, Spy-Detector that:
- prints stats of every peer(inv, tx, getdata) at every 10 secs
- for connection being closed their stats are printed and removed from shared state


0xB10C (changes requested):
INV and GetData messages can contain e.g. blocks too and not only transactions. You currently use the msg type string from the Metadata, but it's better to match on the message payload. You can filter out block INV and GetData messages there.

Cargo.toml
```diff
@@ -7,5 +7,5 @@ members = [
   "tools/logger",
   "tools/websocket",
   "tools/metrics",
-  "tools/connectivity-check",
+  "tools/connectivity-check", "tools/spy-detector",
```
0xB10C, line 10: needs a new line

tools/spy-detector/src/main.rs
```diff
@@ -0,0 +1,158 @@
...
+use nng::{Protocol, Socket};
+
+use shared::event_msg;
+use shared::event_msg::event_msg::Event;
+use shared::net_msg;
+use shared::prost::Message;
+use simple_logger::SimpleLogger;
+
+const ADDRESS: &str = "tcp://127.0.0.1:8883";
+const STATS_INTERVAL: Duration = Duration::from_secs(120); // Duration(seconds) to print stats of peers
```
0xB10C, line 16: we should make these configurable through command line arguments with clap. See #30 for an example

i-am-yuvi: Correct!! we can pass some arguments using clap to configure that!!

i-am-yuvi:
> INV and GetData messages can contain e.g. blocks too and not only transactions. You currently use the msg type string from the Metadata, but it's better to match on the message payload. You can filter out block INV and GetData messages there.

Agree!!

i-am-yuvi:
> INV and GetData messages can contain e.g. blocks too and not only transactions. You currently use the msg type string from the Metadata, but it's better to match on the message payload. You can filter out block INV and GetData messages there.

So basically INVs can contain blocks too, my question is:

- spy nodes can relay block inv but not transaction inv?
- can INV messages contain both Block and Tx at the same time?

0xB10C:
> So basically INVs can contain blocks too, my question is:

See https://docs.rs/bitcoin/latest/bitcoin/p2p/message_blockdata/enum.Inventory.html

We want to handle `Inventory::Transaction` and `Inventory::WTx` (protobuf definitions here https://github.com/0xB10C/peer-observer/blob/712a51c4d435ba91cc9fd404965eb507ef7edd56/protobuf/proto-types/primitive.proto#L49-L52) for INVs. We might need to handle `Inventory::WitnessTransaction` (https://github.com/0xB10C/peer-observer/blob/712a51c4d435ba91cc9fd404965eb507ef7edd56/protobuf/proto-types/primitive.proto#L54) too for GET_DATA. 

See also https://github.com/bitcoin/bitcoin/blob/5abb9b1af49be9024f95fa2f777285c531785d85/src/protocol.h#L473-L490 and the comments there.

> * spy nodes can relay block inv but not transaction inv?

I think we can ignore block invs for now. Here, we are only interested in TX / WTX invs.

> * can INV messages contain both Block and Tx at the same time?

Yes, that's possible, but not too common. Best would be to iterate over an INV message and increment the counter for each transaction/witness_transaction in the INV message.

i-am-yuvi:
> > So basically INVs can contain blocks too, my question is:
> 
> See https://docs.rs/bitcoin/latest/bitcoin/p2p/message_blockdata/enum.Inventory.html
> 
> We want to handle `Inventory::Transaction` and `Inventory::WTx` (protobuf definitions here
> 
> https://github.com/0xB10C/peer-observer/blob/712a51c4d435ba91cc9fd404965eb507ef7edd56/protobuf/proto-types/primitive.proto#L49-L52
> 
> ) for INVs. We might need to handle `Inventory::WitnessTransaction` (
> https://github.com/0xB10C/peer-observer/blob/712a51c4d435ba91cc9fd404965eb507ef7edd56/protobuf/proto-types/primitive.proto#L54
> 
> ) too for GET_DATA.
> See also https://github.com/bitcoin/bitcoin/blob/5abb9b1af49be9024f95fa2f777285c531785d85/src/protocol.h#L473-L490 and the comments there.
> 
> > * spy nodes can relay block inv but not transaction inv?
> 
> I think we can ignore block invs for now. Here, we are only interested in TX / WTX invs.
> 
> > * can INV messages contain both Block and Tx at the same time?
> 
> Yes, that's possible, but not too common. Best would be to iterate over an INV message and increment the counter for each transaction/witness_transaction in the INV message.

great!! Thanks for sharing!

i-am-yuvi:
I think we can define primitive for Getdata too! @0xB10C ?

0xB10C:
We already have one!

i-am-yuvi:
> We already have one!

Okay got it!!

0xB10C:
Hey, are you still working on this?

i-am-yuvi:
> Hey, are you still working on this?

Yes, I am working!!

i-am-yuvi:
Need to resolve conflicts and squash commits!!

i-am-yuvi:
Continued in #57 , I'm closing this for now!

i-am-yuvi:
Continued in #57  , I'm closing this for now!

Issue #46, opened 2024-09-02
Improve logging for all tools


0xB10C:
Currently, most tools and the extractor don't have proper logging. We might want to:

- [ ] move the `simple_logger` dependency into `shared` to have only one place to upgrade and to be sure we're only using one version across all tools
- [ ] replace all `println!` macros with their respective `log:<level>!` macros
- [ ] insert `SimpleLogger::new().init()` at startup
- [ ]  (optionally) make logging levels configurable through clap
- ...

PR #50, opened 2024-09-02
Improve logging of tools and extractor

0xB10C:
flowing up on https://github.com/0xB10C/peer-observer/pull/42

from #46: 

- [x] move the `simple_logger` dependency into `shared` to have only one place to upgrade and to be sure we're only using one version across all tools
- [x] replace all `println!` macros with their respective `log:<level>!` macros
- [x] insert `SimpleLogger::new().init()` at startup
- [x]  make logging levels configurable through clap

Closes https://github.com/0xB10C/peer-observer/issues/46

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -496,7 +496,6 @@ dependencies = [
  "crossbeam",
  "csv",
  "lazy_static",
- "log",
  "nng",
  "prometheus",
  "rand",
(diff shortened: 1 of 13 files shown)


i-am-yuvi (approved):
Tested ACK!! 🔥

0xB10C:
Thanks for the detailed review @i-am-yuvi!

PR #57, opened 2024-11-04
add: spy-detector

### LINKED ISSUE #43: Automaticall Detecting Spy Peers

yuvicc:
Fixes #43 

This creates a tool, Spy-Detector that:
- prints stats of every peer(inv, tx, getdata) at every 10 secs
- for the connection being closed their stats are printed and removed from the shared state

### DIFF
--- a/tools/spy-detector/Cargo.toml
+++ b/tools/spy-detector/Cargo.toml
@@ -4,8 +4,5 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
-nng = "1.0.1"
 dashmap = "6.1.0"
-chrono = "0.4.38"
 shared = { path = "../../shared" }
-log = "0.4.22"
--- a/tools/spy-detector/Cargo.toml
+++ b/tools/spy-detector/Cargo.toml
@@ -1,0 +1,11 @@
+[package]
+name = "spy-detector"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+nng = "1.0.1"
+dashmap = "6.1.0"
+chrono = "0.4.38"
+shared = { path = "../../shared" }
+log = "0.4.22"
--- a/tools/spy-detector/Cargo.toml
+++ b/tools/spy-detector/Cargo.toml
@@ -4,5 +4,4 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
-dashmap = "6.1.0"
 shared = { path = "../../shared" }
--- a/tools/spy-detector/src/main.rs
+++ b/tools/spy-detector/src/main.rs
@@ -1,4 +1,3 @@
-use dashmap::DashMap;
 use shared::clap;
 use shared::clap::Parser;
 use shared::event_msg;
@@ -10,14 +9,9 @@ use shared::nng::options::Options;
 use shared::nng::{Protocol, Socket};
 use shared::prost::Message;
 use shared::simple_logger;
-use std::sync::atomic;
-use std::sync::atomic::AtomicU32;
-use std::sync::Arc;
+use std::collections::HashMap;
 use std::time::Duration;
 use std::time::Instant;
-
-// const ADDRESS: &str = "tcp://127.0.0.1:8883";
-// const STATS_INTERVAL: Duration = Duration::from_secs(120); // Duration(seconds) to print stats of peers
 
 const LOG_TARGET: &str = "main";
 
@@ -34,7 +28,7 @@ struct Args {
 
     /// Duration (in seconds) to print stats of peers
     #[arg(short = 'i', long, default_value = "120")]
-    stats_interval: u64,
+    interval: u64,
 
     /// The log level the tool should run with. Valid log levels
     /// are "trace", "debug", "info", "warn", "error". See https://docs.rs/log/latest/log/enum.Level.html
@@ -44,20 +38,19 @@ struct Args {
 
 #[derive(Debug, Default)]
 struct PeerStats {
-    inv_tx_received: AtomicU32,            // TX(INV) received by the peer
-    inv_tx_sent: AtomicU32,                // TX(INV) sent by the peer
-    inv_wtx_sent: AtomicU32,               // WTX(INV) received by the peer
-    inv_wtx_received: AtomicU32,           // WTX(INV) received by the peer
-    inv_witnesstx_received: AtomicU32,     // WitnessTX(INV) received by the peer
-    inv_witnesstx_sent: AtomicU32,         // WitnessTX(INV) sent by the peer
-    getdata_witnesstx_sent: AtomicU32,     // WitnessTX(GETDATA) sent by the peer
-    getdata_witnesstx_received: AtomicU32, // WitnessTX(GETDATA) received by the peer
-    tx_sent: AtomicU32,                    // TX sent by the peer
-    tx_received: AtomicU32,                // TX received by the peer
-    last_activity: Option<Instant>,        // Last activity time of the peer
+    inv_tx_received: u32,            // TX(INV) received by the peer
+    inv_tx_sent: u32,                // TX(INV) sent by the peer
+    inv_wtx_sent: u32,               // WTX(INV) received by the peer
+    inv_wtx_received: u32,           // WTX(INV) received by the peer
+    inv_witnesstx_received: u32,     // WitnessTX(INV) received by the peer
+    inv_witnesstx_sent: u32,         // WitnessTX(INV) sent by the peer
+    getdata_witnesstx_sent: u32,     // WitnessTX(GETDATA) sent by the peer
+    getdata_witnesstx_received: u32, // WitnessTX(GETDATA) received by the peer
+    tx_sent: u32,                    // TX sent by the peer
+    tx_received: u32,                // TX received by the peer
 }
 
-type PeerMap = Arc<DashMap<String, PeerStats>>;
+type PeerMap = HashMap<String, PeerStats>;
 
 fn main() {
     let args = Args::parse();
@@ -65,11 +58,9 @@ fn main() {
     //to-do: use threshold at appropriate location
     let threshold = &args.threshold;
     let address = &args.address;
-    let stats_interval = Duration::from_secs(args.stats_interval);
+    let stats_interval = Duration::from_secs(args.interval);
 
     simple_logger::init_with_level(args.log_level).unwrap();
-
-    log::info!(target: LOG_TARGET, "Starting spy-detector...",);
 
     log::info!(target: LOG_TARGET, "Starting spy-detector...",);
 
@@ -79,27 +70,22 @@ fn main() {
     let all_topics = vec![];
     sub.set_opt::<Subscribe>(all_topics).unwrap();
 
-    let peer_map: PeerMap = Arc::new(DashMap::new());
+    let mut peer_map: PeerMap = HashMap::new();
 
-    // Spawn a thread for periodic stats display - 2 mins
-    let display_peer_map = peer_map.clone();
-    std::thread::spawn(move || {
-        let mut last_display = Instant::now();
-        loop {
-            let now = Instant::now();
-            if now.duration_since(last_display) >= stats_interval {
-                display_all_stats(&display_peer_map);
-                last_display = now;
-            }
-            std::thread::sleep(Duration::from_secs(1));
-        }
-    });
+    let mut last_stats_display = Instant::now();
 
     log::info!(target: LOG_TARGET, "Spy-detector started",);
 
     loop {
         let msg = sub.recv().unwrap();
         let message = event_msg::EventMsg::decode(msg.as_slice()).unwrap().event;
+
+        // Check if it's time to display stats
+        let now = Instant::now();
+        if now.duration_since(last_stats_display) >= stats_interval {
+            display_all_stats(&peer_map);
+            last_stats_display = now;
+        }
 
         if let Some(event) = message {
             match event {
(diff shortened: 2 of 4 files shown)


i-am-yuvi:
Can you review this @0xB10C ?

tools/spy-detector/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "spy-detector"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+nng = "1.0.1"
```
0xB10C, line 7: no need to have `nng` as a dependency here. You can use it via `shared`.

Drop this 
```suggestion
```
and use it similar to

https://github.com/0xB10C/peer-observer/blob/8a445f71608bfd968c1231fbfa5b5864f06ab53c/tools/websocket/src/main.rs#L11-L12

tools/spy-detector/Cargo.toml
```diff
@@ -0,0 +1,11 @@
...
+dashmap = "6.1.0"
+chrono = "0.4.38"
+shared = { path = "../../shared" }
+log = "0.4.22"
```
0xB10C, line 11: This is also already in `shared`. You can use it with

https://github.com/0xB10C/peer-observer/blob/8a445f71608bfd968c1231fbfa5b5864f06ab53c/tools/websocket/src/main.rs#L9

tools/spy-detector/src/main.rs
```diff
@@ -0,0 +1,262 @@
...
+use shared::net_msg;
+use shared::prost::Message;
+use shared::simple_logger;
+use std::sync::atomic;
+use std::sync::atomic::AtomicU32;
+use std::sync::Arc;
+use std::time::Duration;
+use std::time::Instant;
+
+const ADDRESS: &str = "tcp://127.0.0.1:8883";
```
0xB10C, line 18: This should be a clap argument. See e.g.

https://github.com/0xB10C/peer-observer/blob/8a445f71608bfd968c1231fbfa5b5864f06ab53c/tools/websocket/src/main.rs#L22-L24

tools/spy-detector/src/main.rs
```diff
@@ -0,0 +1,262 @@
...
+const STATS_INTERVAL: Duration = Duration::from_secs(10); // Duration(seconds) to print stats of peers
```
0xB10C, line 19: probably also good to have this as an argument

tools/spy-detector/src/main.rs
```diff
@@ -0,0 +1,262 @@
...
+
+const LOG_TARGET: &str = "main";
+
+#[derive(Parser, Debug)]
+#[command(version, about, long_about=None)]
+struct Args {
+    /// set the threshold for spy detection (default value is 5)
```
0xB10C, line 26: It would be good to explain what the threshold affects. 

And explain: When would we want to increase or lower it?

tools/spy-detector/src/main.rs
```diff
@@ -0,0 +1,262 @@
...
+    last_activity: Option<Instant>,        // Last activity time of the peer
+}
+
+type PeerMap = Arc<DashMap<String, PeerStats>>;
+
+fn main() {
+    let args = Args::parse();
+
+    //to-do: use threshold at appropriate location
+    let threshold = &args.threshold;
```
0xB10C, line 57: Probably good to address this todo?

tools/spy-detector/src/main.rs
```diff
@@ -0,0 +1,262 @@
...
+
+    simple_logger::init_with_level(args.log_level).unwrap();
+
+    log::info!(target: LOG_TARGET, "Starting spy-detector...",);
+
+    log::info!(target: LOG_TARGET, "Starting spy-detector...",);
```
0xB10C, line 63: Duplicate

tools/spy-detector/src/main.rs
0xB10C: IIRC we discussed printing every 2 minutes at one point - do you have a reason for using 10s now?

tools/spy-detector/src/main.rs
```diff
@@ -0,0 +1,262 @@
...
+
+    let sub = Socket::new(Protocol::Sub0).unwrap();
+    sub.dial(ADDRESS).unwrap();
+
+    let all_topics = vec![];
+    sub.set_opt::<Subscribe>(all_topics).unwrap();
+
+    let peer_map: PeerMap = Arc::new(DashMap::new());
+
+    // Spawn a thread for periodic stats display - 2 mins
```
0xB10C, line 73:
```suggestion
    // Spawn a thread for periodic stats display
```

tools/spy-detector/src/main.rs
```diff
@@ -0,0 +1,262 @@
...
+    let display_peer_map = peer_map.clone();
+    std::thread::spawn(move || {
+        let mut last_display = Instant::now();
+        loop {
+            let now = Instant::now();
+            if now.duration_since(last_display) >= STATS_INTERVAL {
+                display_all_stats(&display_peer_map);
+                last_display = now;
+            }
+            std::thread::sleep(Duration::from_secs(1));
```
0xB10C, line 83: I think you could also sleep for `STATS_INTERVAL`. No need to wake up every second and check.

tools/spy-detector/src/main.rs
```diff
@@ -0,0 +1,262 @@
...
+        if let Some(event) = message {
+            match event {
+                Event::Msg(msg) => {
+                    let msg_type: u32;
+
+                    if msg.meta.inbound {
+                        msg_type = 0;
+                    } else {
+                        msg_type = 1;
+                    };
```
0xB10C, line 102: Why do you re encode the `msg.meta.inbound` boolean as an `u32`? You only ever check if it's ` == 0`. This should stay a boolean!

tools/spy-detector/src/main.rs
```diff
@@ -0,0 +1,262 @@
...
+                                process_tx_msg(&peer_map, msg_type, peer_id);
+                            }
+                            _ => {}
+                        }
+                    }
+                }
+                Event::Conn(c) => {
+                    if let Some(event) = c.event {
+                        process_connection_event(&peer_map, &event.to_string());
+                        //println!("{}", event);
```
0xB10C, line 126:
```suggestion
```

tools/spy-detector/src/main.rs
```diff
@@ -0,0 +1,262 @@
...
+            println!("Connection closed for peer: {}", peer_id);
+            print_peer_stats(peer_id, &stats.1);
+        }
+    }
+}
+
+fn print_peer_stats(peer_addr: &str, stats: &PeerStats) {
+    println!(
+        "[{}] Peer ID {} stats:\n  INV TX sent: {:?}\n  INV TX received: {:?}\n  INV WTX received: {:?}\n INV WitnessTX received: {:?}\n GetDATA WitnessTX received: {:?}\n  GETDATA WitnessTX received: {:?}\n  Tx sent: {:?}\n  Tx received: {:?}\n Last Activity: {:?}",
+        chrono::Local::now().format("%Y-%m-%d %H:%M:%S"),
```
0xB10C, line 233: I don't think you need to print the time here

tools/spy-detector/src/main.rs
```diff
@@ -0,0 +1,262 @@
...
+    if event.starts_with("closed") {
+        let peer_id = event.split(' ').nth(1).unwrap_or("");
+        if let Some(stats) = peer_map.remove(peer_id) {
+            println!("Connection closed for peer: {}", peer_id);
+            print_peer_stats(peer_id, &stats.1);
+        }
+    }
+}
+
+fn print_peer_stats(peer_addr: &str, stats: &PeerStats) {
```
0xB10C, line 230: Instead of having a print_peer_stats functions, it might be better to implement `std::fmt::Display` for `PeerStats`.

tools/spy-detector/src/main.rs
```diff
@@ -0,0 +1,262 @@
...
+    let map = peer_map;
+    println!(
+        "\n===== Stats for all peers ({}): =====",
+        chrono::Local::now().format("%Y-%m-%d %H:%M:%S")
+    );
+    for item in map.iter() {
+        let peer_id = item.key();
+        let stats = item.value();
+
+        //println!("Key: {}\n", peer_id);
```
0xB10C, line 257:
```suggestion
```

tools/spy-detector/src/main.rs
0xB10C: I just saw that it's not used. Feel free to remove it

tools/spy-detector/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "spy-detector"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+nng = "1.0.1"
+dashmap = "6.1.0"
+chrono = "0.4.38"
```
0xB10C, line 9: I don't think we need to log the timestamps (as noted below) - I think we can drop chrono.

tools/spy-detector/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "spy-detector"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+nng = "1.0.1"
+dashmap = "6.1.0"
```
0xB10C, line 8: I see that you are using DashMap ("DashMap is an implementation of a concurrent associative array/hashmap in Rust."). You `std::thread::spawn` a thread to print the stats periodically, and the main thread collects the stats.   

I think you could print in the main thread by checking `STATS_INTERVAL` after a message arrived, which would reduce the overall complexity of the code and allows us to drop DashMap in favor of a `std::collection::HashMap`. I prefer to keep the dependency footprint minimal.

tools/spy-detector/src/main.rs
```diff
@@ -0,0 +1,262 @@
...
+                display_all_stats(&display_peer_map);
+                last_display = now;
+            }
+            std::thread::sleep(Duration::from_secs(1));
+        }
+    });
+
+    log::info!(target: LOG_TARGET, "Spy-detector started",);
+
+    loop {
```
0xB10C, line 89: To do the printing in the main thread, you cold do it here similar to:

```suggestion
    loop {
        if Instant::now().duration_since(last_display) >= STATS_INTERVAL {
            display_all_stats(&display_peer_map);
            last_display = now;
        }
    
```

tools/spy-detector/src/main.rs
```diff
@@ -0,0 +1,262 @@
...
+        stats.tx_received,
+        stats.last_activity
+    );
+}
+
+fn display_all_stats(peer_map: &PeerMap) {
+    let map = peer_map;
+    println!(
+        "\n===== Stats for all peers ({}): =====",
+        chrono::Local::now().format("%Y-%m-%d %H:%M:%S")
```
0xB10C, line 251: I don't hink we need the timestamp as we usually only care about the most recent one.

0xB10C (changes requested):
Thank you!

This adds the stats collection and printing part of the spy-collector. However, it doesn't tell us anything about which peers are spys. I think this should be a part of the tool before we merge it. I don't know if you have the time to add it though.

Left some code comments below.

tools/spy-detector/Cargo.toml
i-am-yuvi: Okay, done!

tools/spy-detector/Cargo.toml
i-am-yuvi: okay

tools/spy-detector/src/main.rs
i-am-yuvi: Okay

tools/spy-detector/src/main.rs
i-am-yuvi: Done!!

tools/spy-detector/src/main.rs
i-am-yuvi: done!!

tools/spy-detector/Cargo.toml
i-am-yuvi: correct!!

tools/spy-detector/src/main.rs
i-am-yuvi: Okay, removing timestamps for now!!

i-am-yuvi:
> Thank you!
> 
> This adds the stats collection and printing part of the spy-collector. However, it doesn't tell us anything about which peers are spys. I think this should be a part of the tool before we merge it. I don't know if you have the time to add it though.
> 
> Left some code comments below.

I am analyzing the ratio of inv/getdata that may indicate a spy node. Will push that when ready!

tools/spy-detector/src/main.rs
i-am-yuvi: Threshold is the ratio set by the user to indicate spy nodes; we will need this for detecting spies.

tools/spy-detector/Cargo.toml
i-am-yuvi: Correct, I'll implement that asap!

i-am-yuvi: In that case, we can remove Arc and only use HashMap, since Arc provides immutable shared access to multiple threads.

tools/spy-detector/src/main.rs
0xB10C: Yes, I mean as a comment in the code.

i-am-yuvi: Okay, done!

Maxzy000 (approved):

i-am-yuvi:
@Maxzy000 any suggestion you would like to add??

Maxzy000:
No how can I help you

0xB10C:
closing this for now due to inactivity. feel free to pick it up again at some point and open a new PR when it's ready for review

Issue #58, opened 2024-11-06
Support PID / PID file to attach tracepoints to bitcoind


0xB10C:
With https://github.com/bitcoin/bitcoin/pull/26593, we will only be able to attach to a bitcoind instance via a PID. The peer-observer-extractor should support this:
- either, pass a PID via command line
- or, read a bitcoind PID file

Note that the peer-observer-extractor might need to be restarted after bitcoind is restarted as the PID changes.

0xB10C:
While implementing PID file reading in e.g. systemd, I noticed it's probably cleaner to support this in the extractor itself. Doing this in #64

PR #61, opened 2024-11-07
add: cli arg to specify a bitcoind PID

0xB10C:
Partly addresses #58

This is optional for now but will become required if  bitcoin/bitcoin/26593 is merged.


0xB10C:
I've tested that this indeed works well with #26593. Existing setups will continue to work.

PR #106, opened 2025-01-24
Add python tool to record `getblocktxn` msgs

0xB10C:
Recording for https://delvingbitcoin.org/t/stats-on-compact-block-reconstructions/1052/21?u=0xb10c


tools/record-getblocktxn-py/record-getblocktxn.py
```diff
@@ -0,0 +1,40 @@
+import asyncio
+from nats.aio.client import Client as NATS
+import json
+import sys
+sys.path.append("../../protobuf/python-types")
+
+import event_msg_pb2
+
+def event_to_row(event):
```
0xB10C, line 9: Actually use this to write to a CSV file

PR #134, opened 2025-03-13
fix: Improve error handling

nassersaazi:
- return Result from functions, using the ? operator to propagate errors up the call stack

- Ensure that all potential error points are handled consistently


tools/connectivity-check/Cargo.toml
```diff
@@ -14,7 +14,7 @@ prometheus = "0.13.3"
 lazy_static = "1.5.0"
 csv = "1.2.1"
 serde = { version = "1.0", features = ["derive"] }
-
+anyhow = "1.0"
```
0xB10C, line 17: Hm, I would prefer to handle the errors with the error type of the Rust standard library instead of introducing a dependency on `anyhow`. I don't think using `anyhow` is required here.

See for example https://github.com/0xB10C/peer-observer/blob/90c23c5e67bf04692ca0dc7af4b2bed14150698b/extractor/src/error.rs#L9-L17

Cargo.lock
```diff
@@ -1,6 +1,6 @@
 # This file is automatically @generated by Cargo.
 # It is not intended for manual editing.
-version = 3
+version = 4
```
0xB10C, line 3: Also, seems like upgrading this doesn't (yet) work with the Nixos CI.

PR #138, opened 2025-04-07
Fix: Add filters to logger tool

### LINKED ISSUE #48: Implement filter for `logger` tool
The logger tool is quite spammy currently as it prints all events (p2p message, connections, addrman (if enabled), validation, and mempool). It would be good to have a feature where we can pass a command line argument to enable only the desired event types. E.g. passing `--log-p2pmsgs` would only log p2p messages. Passing `--log-p2pmsgs --log-connections` would log p2p messages and p2p connections.

nassersaazi:
Fixes #48

The logger tool is quite spammy currently as it prints all events
 (p2p message, connections, addrman (if enabled), validation, and mempool).
This PR adds a filter where we can pass a command line argument to enable only the desired event types. For example, passing --filter mempool or -f mempool would only log $Mempool messages.

### DIFF
--- a/tools/logger/src/main.rs
+++ b/tools/logger/src/main.rs
@@ -8,7 +8,11 @@ use shared::prost::Message;
 use shared::simple_logger;
 use shared::{clap, nats};
 
-/// A peer-observer tool that logs all received event messages
+/// A peer-observer tool that logs all received event messages.
+/// By default, all events are shown. This can be a lot. Events can be
+/// filtered by type. For example, --messages only shows P2P messages.
+/// Using `--messages --connections` together showns both P2P messages
+/// and connections.
 #[derive(Parser, Debug)]
 #[command(version, about, long_about = None)]
 struct Args {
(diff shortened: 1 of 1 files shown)


tools/logger/src/main.rs
```diff
@@ -47,26 +51,27 @@ fn main() {
...
                         "$Mempool {}", m.event.unwrap()
                     };
                 }
-                Event::Validation(v) => {
+                Event::Validation(v) if args.filter == "all" || args.filter == "validation" => {
                     log::info! {
                         "+Validation {}", v.event.unwrap()
                     };
                 }
+                _ => {}
```
0xB10C, line 74: Previously, the compiler would error if we forgot to add a new event to the logger. With this change, this will be silently ignored. I think it might be better to check inside the match like e.g. `=> { if args.filter == "all" || args.filter == "validation" { log::info!(...) } }`.

0xB10C:
Nice. I've tested this with `--filter=mempool` and also had a look at the docs.
```
$ cargo run --bin logger -- --help
A peer-observer tool that logs all received event messages

Usage: logger [OPTIONS]

Options:
  -n, --nats-address <NATS_ADDRESS>  The NATS server address the tool should connect and subscribe to [default: 127.0.0.1:4222]
  -l, --log-level <LOG_LEVEL>        The log level the tool should run on. Events are logged with the INFO log level. Valid log levels are "trace", "debug", "info", "warn", "error". See https://docs.rs/log/latest/log/enum.Level.html [default: DEBUG]
  -f, --filter <FILTER>              Filter events by type (e.g., "mempool", "conn", "all") [default: all] [possible values: all, mempool, conn, addrman, validation]
  -h, --help                         Print help
  -V, --version                      Print version
```

0xB10C:
Pushed a minor fixup and squashed. Thanks again. Looks good!

```
A peer-observer tool that logs all received event messages. By default, all events are shown. This can be a lot. Events can be filtered by type. For example, --messages only shows P2P messages. Using `--messages --connections` together showns both P2P messages and connections

Usage: logger [OPTIONS]

Options:
  -n, --nats-address <NATS_ADDRESS>  The NATS server address the tool should connect and subscribe to [default: 127.0.0.1:4222]
  -l, --log-level <LOG_LEVEL>        The log level the tool should run on. Events are logged with the INFO log level. Valid log levels are "trace", "debug", "info", "warn", "error". See https://docs.rs/log/latest/log/enum.Level.html [default: DEBUG]
      --messages                     If passed, show P2P message events
      --connections                  If passed, show P2P connection events
      --addrman                      If passed, show addrman events
      --mempool                      If passed, show mempool events
      --validation                   If passed, show validation events
  -h, --help                         Print help
  -V, --version                      Print version
```

Issue #188, opened 2025-07-28
Grafana graphics


edilmedeiros:
How updated are the dashboards in https://github.com/0xB10C/peer-observer/tree/master/tools/metrics/dashboards?
I'm trying to load them using Grafana v12.2.0-16557133545 (910454c84c) but I only get empty graphs.

Example using `messages.json`.
<img width="1582" height="1125" alt="Image" src="https://github.com/user-attachments/assets/e9c2c952-e8ca-4a3b-8960-164370b7b334" />

The following is a snippet of the metrics endpoint, it seems to be working ok. 
I can see them on Prometheus.

```
# HELP peerobserver_conn_inbound Number of inbound connections.
# TYPE peerobserver_conn_inbound counter
peerobserver_conn_inbound 1
# HELP peerobserver_conn_inbound_current Number of currently open inbound connections.
# TYPE peerobserver_conn_inbound_current gauge
peerobserver_conn_inbound_current 1
# HELP peerobserver_conn_inbound_network Number of inbound connections by network.
# TYPE peerobserver_conn_inbound_network counter
peerobserver_conn_inbound_network{network="0"} 1
# HELP peerobserver_conn_inbound_subnet Number of inbound connections by subnet (where applicable).
# TYPE peerobserver_conn_inbound_subnet counter
peerobserver_conn_inbound_subnet{subnet="172.18.0.0"} 1
# HELP peerobserver_p2p_feefilter_feerate Number of feefilter messages received and sent by feerate.
# TYPE peerobserver_p2p_feefilter_feerate counter
peerobserver_p2p_feefilter_feerate{direction="inbound",feerate="1000"} 1
peerobserver_p2p_feefilter_feerate{direction="outbound",feerate="1000"} 1
peerobserver_p2p_feefilter_feerate{direction="outbound",feerate="9170997"} 1
# HELP peerobserver_p2p_message_bytes Number of P2P network messages bytes send or received.
# TYPE peerobserver_p2p_message_bytes counter
peerobserver_p2p_message_bytes{connection_type="1",direction="inbound",message="block"} 246107
peerobserver_p2p_message_bytes{connection_type="1",direction="inbound",message="cmpctblock"} 39260
peerobserver_p2p_message_bytes{connection_type="1",direction="inbound",message="feefilter"} 8
peerobserver_p2p_message_bytes{connection_type="1",direction="inbound",message="getaddr"} 0
[...]

PR #193, opened 2025-08-01
Update namespaces in grafana dashboard files

m4ycon:
While I was trying to setup metrics tool to work with prometheus and grafana, I was trying to use the `tools/metrics/dashboards/**` files to see some data and as described in metric's readme it was outdated. It seems that was a simple change from namespace that caused it, from changing `networkobserver` to `peerobserver` (https://github.com/0xB10C/peer-observer/commit/65d2801d33e17e96038f8e82aada65812f08b23e). Also, grafana was "alerting" for some visualizations in `connections` and `mempool`, just edited and hitted a save that seemed to fix, especially `mempool` it worked very well.

Grafana v12.1.0 was used for updating `connections` and `mempool` files.

I don't know if it fixes for all cases but many of them seems to be working right, can't confirm for them all because problably I didn't get my node running enough time to get some types of data, it was for like 10 to 20 minutes in mainnet.

I didn't saw #188 until today, and by your comment (@0xB10C) seems that you will replace these dashboards with your current ones, do you think this PR still holds some value?

The below image shows a slice of `messages.json` working properly.
image


0xB10C:
Correct, renaming from `networkobserver` to `peerobserver` should mostly do the trick.

> I didn't saw https://github.com/0xB10C/peer-observer/issues/188 until today, and by your comment (@0xB10C) seems that you will replace these dashboards with your current ones, do you think this PR still holds some value?

I plan to add my current set of dashboards soon (next week). But having a working version of the old dashboards in the git history doesn't hurt, especially since it's a simple change. But no need to spend more time on getting the old ones to work.

0xB10C:
Thank you!

Issue #198, opened 2025-08-01
metrics tool: add more metrics based on the getpeerinfo RPC event data


0xB10C:
Some initial metrics were added in #191:

https://github.com/0xB10C/peer-observer/blob/8a49c622e21a14a08f4d40df8f16bfaf5830e874/tools/metrics/src/main.rs#L114-L120

but we should add more. Ideas are:

- median ping time across all connected peers
- average ping time across all connected peers
- median min_ping time across all connected peers
- average min_ping time across all connected peers
- count of peers with a time_offset (their clock compared to our clock) larger than 10s into the future or 10s into the past
- count of peers we consider to be BIP152 high bandwidth ``bip152_hb_to``, and count of peers that consider us to be BIP152 high bandwidth `bip152_hb_from`
- count of peers (while we have this via the connection tracepoints already, good to have this metric alone here too if someone isn't using the tracepoints)
- count of `addr_relay_enabled` peers
- count of `v1` peers (`transport_protocol_type`)
- count of `v2` peers (`transport_protocol_type`)
- count of peers by network
- count of peers that have inflight blocks, and count of distinct inflight blocks
- count of peers by connection_type
- count by version (not subversion, version like 70001)
- ...

0xB10C:
- count of peers with a ping_await higher than 5s

0xB10C:
Started working on this in https://github.com/0xB10C/peer-observer/pull/205

PR #205, opened 2025-08-04
Add more metrics based on data from the `getpeerinfo` RPC

0xB10C:
Addresses #198:
- [x] median ping time across all connected peers
- [x] average ping time across all connected peers
- [x] median min_ping time across all connected peers
- [x] average min_ping time across all connected peers
- [x] count of peers with a time_offset (their clock compared to our clock) larger than 10s into the future or 10s into the past
- [x] count of peers we consider to be BIP152 high bandwidth ``bip152_hb_to``, and count of peers that consider us to be BIP152 high bandwidth `bip152_hb_from`
- [x] count of peers (while we have this via the connection tracepoints already, good to have this metric alone here too if someone isn't using the tracepoints)
- [x] count of `addr_relay_enabled` peers
- [x] count of `v1` peers (`transport_protocol_type`)
- [x] count of `v2` peers (`transport_protocol_type`)
- [x] count of peers by network
- [x] count of peers that have inflight blocks, and count of distinct inflight blocks
- [x] count of peers by connection_type
- [x] count by version (not subversion, version like 70001)
- [x] for non-zero AS fields, count the number of connections per AS
- [x] count of peers with a ping_await higher than 5s

### DIFF
--- a/tools/metrics/src/main.rs
+++ b/tools/metrics/src/main.rs
@@ -110,6 +110,7 @@ fn main() {
                 let mut peers_by_network: BTreeMap<&str, i64> = BTreeMap::new();
                 let mut peers_by_connection_type: BTreeMap<&str, i64> = BTreeMap::new();
                 let mut peers_by_protocol_version: BTreeMap<u32, i64> = BTreeMap::new();
+                let mut peers_by_asn: BTreeMap<u32, i64> = BTreeMap::new();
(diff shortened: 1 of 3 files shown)


0xB10C:
I've been testing this and set up a dashboards for these new metrics. Looks good so far.

PR #215, opened 2025-08-08
add: prometheus metric for LinkingLion traffic

### LINKED ISSUE #142: add LinkingLion traffic measuring in metrics tool
A small tool to measure the traffic between us and LinkingLion IPs & us and non-LinkingLion IPs would be helpful to find out how much bandwidth LinkingLion uses. Just summing up the bytes is probably enough.

https://b10c.me/observations/06-linkinglion/

I think gmax meant LinkingLion in his delvingpost https://delvingbitcoin.org/t/bitcoin-node-p2p-traffic-analysis/1490/3:

> There are a number of very long term abusers that are probably wasting a majority of the network’s total non-ibd bandwidth usage on identifiable networks.
> It might be useful to break out those networks in any further reporting for them because they have different implications. E.g. they won’t benefit from erlay gains… and probably the only way to mitigate them is to start shipping a list of subnets that are relegated to blocks only by default or similar.

I'm not convinced that LinkingLion has a major impact on node traffic, but happy to stand corrected by empirical results.

0xB10C:
Instead of having an extra tool as discussed in #142, do it in the `metrics` tool.

closes #142

Needs some dashboards..

### DIFF
--- a/tools/metrics/src/main.rs
+++ b/tools/metrics/src/main.rs
@@ -264,6 +264,18 @@ fn main() {
             .with(&labels)
             .inc_by(msg.meta.size);
 
+        if util::is_on_linkinglion_banlist(&ip) {
+            let mut labels_ll = HashMap::<&str, &str>::new();
+            labels_ll.insert(metrics::LABEL_P2P_MSG_TYPE, &msg.meta.command);
+            labels_ll.insert(metrics::LABEL_P2P_DIRECTION, &direction);
+            metrics::P2P_MESSAGE_COUNT_LINKINGLION
+                .with(&labels_ll)
+                .inc();
+            metrics::P2P_MESSAGE_BYTES_LINKINGLION
+                .with(&labels_ll)
+                .inc_by(msg.meta.size);
+        }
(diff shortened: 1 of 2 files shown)


tools/metrics/src/main.rs
```diff
@@ -264,6 +264,18 @@ fn main() {
...
             .inc_by(msg.meta.size);
 
+        if util::is_on_linkinglion_banlist(&ip) {
+            let mut labels_ll = HashMap::<&str, &str>::new();
+            labels_ll.insert(metrics::LABEL_P2P_MSG_TYPE, &msg.meta.command);
+            labels_ll.insert(metrics::LABEL_P2P_DIRECTION, &direction);
+            metrics::P2P_MESSAGE_COUNT_LINKINGLION
+                .with(&labels_ll)
+                .inc();
+            metrics::P2P_MESSAGE_BYTES
```
0xB10C, line 274:
```suggestion
            metrics::P2P_MESSAGE_BYTES_LINKINGLION
```

Issue #26, opened 2024-07-23
Test tools


0xB10C:
By recording and then replaying some of the nanomsg/protobuf communication that the extractor publishes, we can write regression test for the tools in the CI. This would be useful to have!

See also: https://github.com/0xB10C/peer-observer/pull/25#issuecomment-2246349226

0xB10C:
An alternative to a NATS JetStream based approach could be a testing-extractor that publishes a fixed set of messages into a NATS server and clients read from it. This might be easier to implement and makes it easier to add new test cases since we don't need to recreate a NATS JetStream file to read from. A workflow for this would be similar to:

- Spin up a NATS server in GH Actions (using Docker or binary).
- Wait until NATS is ready.
- Run a tool / consumer.
- Run the testing-extractor / publisher (it should send a few test messages).
- Assert/test that the tool output is as expected

A GH actions workflow to run a NATS server could look similar to:

```yaml
name: Integration Test

on: [push, pull_request]

jobs:
  test:
    runs-on: ubuntu-latest
    services:
      nats:
        image: nats:latest
        ports:
          - 4222:4222
        options: >-
          --health-cmd "nats-server -v"
          --health-interval 5s
          --health-timeout 10s
          --health-retries 5

    steps:
...
```

0xB10C:
I moved the ebpf extractor into `extractors/`, this allows us to have multiple extractors in this directory. For example, a `testing-extractor`.

0xB10C:
The big thing missing here is the ebpf-extractor integration tests. These are probably a bit harder to do than the others because they also require root privileges to run. That's fine for the CI, but not ideal for local development.

0xB10C:
(only saw this after my comment in #380, so reposting it here:

from https://github.com/peer-observer/peer-observer/pull/380#issuecomment-4053460603

> I don't want to recommend devs to run them locally due to their privilege requirements (especially if they aren't touching the ebpf-extractor in their code changes). This means introducing another feature flag for it, and not being able to run with --all-features locally. We can and should run them in CI. Maybe we can also implement a "skip", if the user does not have the right priviledges...

But def Concept ACK on adding tests for the ebpf-extractor in some way!

PR #229, opened 2025-08-19
Integration tests for websocket tool

0xB10C:
As part of #26 this adds integration tests (similar to these in #222) for the `websocket` tool.


0xB10C:
WIP as I want to add more cases with e.g.:
- multiple clients
- disconnecting clients

PR #233, opened 2025-08-22
Fix missed events in metrics and websocket tool

0xB10C:
Previously, `tokio::select!` would always busy-wait in the receiving loop for both tools. As a hack, a 50ms sleep was implemented to avoid busy waiting. However, the root cause for the busy-wait was that the `shutdown_tx` belonging to the `shutdown_rx` was already dropped (not in tests, so this wasn't possible to see in tests). `select!` would most of the time match on the shutdown and we wouldn't process all NATS events. 

This is fixed by actually using `shutdown_tx` which avoids it being droped. At the same time, Ctrl+C handling is implemented.


0xB10C:
These were introduced with #229 and #222

PR #235, opened 2025-08-25
Add metric for large inv sets (>35) sent to peers

### LINKED ISSUE #220: metrics tool: keep track of tx INVs with more than 35 entries (DoS due to inv-to-send sets growing too large)
As a fix for https://bitcoincore.org/en/2024/10/08/disclose-large-inv-to-send/, https://github.com/bitcoin/bitcoin/pull/27610 made the number of tx INV entries dynamic. It would be good to track how many INVs with more than 35 entries we send to peers.

related: https://b10c.me/observations/15-inv-to-send-queue/

jurraca:
Adds a metric to track inv sets larger than 35 called `p2p_invs_outbound_large`.

closes #220

### DIFF
--- a/tools/metrics/src/metrics.rs
+++ b/tools/metrics/src/metrics.rs
@@ -194,6 +194,7 @@ pub struct Metrics {
     pub p2p_inv_entries_histogram: HistogramVec,
     pub p2p_invs_homogeneous: IntCounterVec,
     pub p2p_invs_heterogeneous: IntCounterVec,
+    pub p2p_invs_outbound_large: IntCounter,
     pub p2p_ping_subnet: IntCounterVec,
     pub p2p_oldping_subnet: IntCounterVec,
     pub p2p_version_subnet: IntCounterVec,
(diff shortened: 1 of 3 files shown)


tools/metrics/src/metrics.rs
```diff
@@ -285,6 +286,7 @@ impl Metrics {
         hv!(p2p_inv_entries_histogram, "Histogram number of entries contained in an INV message.", BUCKETS_INV_SIZE, [LABEL_P2P_DIRECTION], registry);
         icv!(p2p_invs_homogeneous, "Number of homogeneous INV entries sent and received.", [LABEL_P2P_DIRECTION], registry);
         icv!(p2p_invs_heterogeneous, "Number of heterogeneous INV entries sent and received.", [LABEL_P2P_DIRECTION], registry);
+        icv!(p2p_invs_outbound_large, "Number of outbound INV messages with more than 35 entries.", [LABEL_P2P_DIRECTION], registry);
```
0xB10C, line 289: I think we don't need the `LABEL_P2P_DIRECTION` since we already have outbound in the name and filter for it.

0xB10C: And could mention https://bitcoincore.org/en/2024/10/08/disclose-large-inv-to-send/ and Improve performance of p2p inv to send queues bitcoin/bitcoin#27610 in the help text too.

jurraca:
Updated, removing the label and turning the metric from an `IntCounterVec` to an `IntCounter`. Also handles WTx as suggested above and updates the test accordingly. 

Question: a peer can either send `Transaction` or `Wtx` but not both in the same Inv ?

PR #237, opened 2025-08-28
add: dashboard for number of invs with >35 items

0xB10C:
This adds a simple dashboard showing the number of transaction INVs with more than 35 entries. This is a follow-up to #235 and generally related to:

- https://bitcoincore.org/en/2024/10/08/disclose-large-inv-to-send/
- https://github.com/bitcoin/bitcoin/pull/27610
- https://b10c.me/observations/15-inv-to-send-queue/

Completes #220

### DIFF
--- a/tools/metrics/dashboards/playlist/inv-with-more-than-35-tx-entries.json
+++ b/tools/metrics/dashboards/playlist/inv-with-more-than-35-tx-entries.json
@@ -1,0 +1,113 @@
+{
+  "annotations": {
+    "list": [
+      {
+        "builtIn": 1,
+        "datasource": {
+          "type": "grafana",
+          "uid": "-- Grafana --"
+        },
+        "enable": true,
+        "hide": true,
+        "iconColor": "rgba(0, 211, 255, 1)",
+        "name": "Annotations & Alerts",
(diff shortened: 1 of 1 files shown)


0xB10C:
I think I know why this is happening: 

- bob knows about a lot of sub 1-sat/vbyte transactions that its peers don't know about
- bob puts these into the inv-to-send sets to relay them to the peers
- the inv-to-send sets grow rather large, so more than 35 tx are send per inv
- since inv-to-send sets are processed in order sorted by highest feerate first, the sub 1 sat/vbyte feerate transactions are processed last
- so first large INVs with high feerate tx are send
- once the sub 1 sat/vbyte feerates are reached, we notice that these are below the peers feerate filter and drop them from the set.

The relevant code for this is here https://github.com/bitcoin/bitcoin/blob/6ca6f3b37b992591726bd13b494369bee3bd6468/src/net_processing.cpp#L5787-L5790

0xB10C:
The timing of large invs correlated with a lot of transactions being added to the mempool on node "bob".
image
 The spikes in transactions added to mempool are only present on bob, so they must be sub 1sat/vbyte transactions.
image

Issue #141, opened 2025-04-10
Implement more extractors


0xB10C:
With NATS (https://github.com/0xB10C/peer-observer/issues/56, #76), we can have multiple extractors pushing events into the message queue. Here are some ideas for future extractors that could be good to implement:

- [x] RPC extractor: in a certain interval, fetch RPC data and publish it into the NATS message queue. For example, `getpeerinfo` can provide useful information about our peers that we currently don't have via the tracepoints. **#191**
- [x] P2P-extractor: Connects to the localhost node via the P2P interface and e.g. frequently pings it and records the ping-pong duration to measure backlog (https://github.com/0xB10C/peer-observer/pull/256)
- [x] Log extractor: parse the log and publish interesting log messages (maybe already parsed and pre-processed) into the NATS message queue. We don't have tracepoints for everything. While the log interface might not be as stable, why not use it if we can. Additionally, the debug log is a machine-to-human interface. Ideally, we'd use machine-to-machine interfaces. (https://github.com/0xB10C/peer-observer/pull/272)
- [x] IPC-extractor: Since Bitcoin Core v30.0, the first IPC interface was introduced. A `TracingEvents` IPC-interface could be introduced which, if low-overhead, would solve some of the tracing interface pain points. See https://github.com/bitcoin-core/libmultiprocess/issues/185 and https://github.com/bitcoin/bitcoin/pull/32898. Also, see https://github.com/peer-observer/peer-observer/issues/370 for an ipc-extractor that we could build to test a Bitcoin Core PR. https://github.com/peer-observer/peer-observer/pull/379
- [ ] call-stack extractor: https://github.com/peer-observer/peer-observer/issues/391
- [ ] TCP extractor: not a high priority, see https://github.com/peer-observer/peer-observer/issues/141#issuecomment-4119199272
- [ ] ZMQ extractor: not as important

edilmedeiros:
> For the RPC extractor, it probably makes sense to use https://crates.io/crates/jsonrpc-core - however it does not provide a proper transport. Maybe the "for non-production use only" transport is enough for us.

Looks like it was superseded by https://github.com/paritytech/jsonrpsee.

Also, `rust-bitcoin` includes an active project to provide types for the RPC interface here https://crates.io/crates/corepc-types (pretty new library, not sure if it covers the whole RPC right now, but is probably a good starting point).

0xB10C:
Implementing an RPC extractor in #191

0xB10C:
[...]

PR #256, opened 2025-09-16
add: p2p-extractor (initially keeps track of ping-pong durations)

0xB10C:
This adds a p2p-extractor. The extractor listens on a TCP socket and accepts inbound connections from a Bitcoin node. A Bitcoin node configured with `--addnode=ip:port` will connect (and reconnect) to the extractor. Additionally, the connection will count as a manual connection and not reduce the number of normal outbound (and not inbounds) the node makes. The extractor does only support P2Pv1 for now. This means, the node and the extractor must have the same network magic. The node will open a v2 connection to the p2p-extractor at first, which will be closed. The node will then retry with a v1 connection which then should succeed. 

Currently, the p2p-extractor only sends a ping and measures the time it takes for the node to respond to a pong. When done on localhost (to reduce network latency), this measures the time it takes for the node to process all other network messages that arrived before it. In a way, we measure the Bitcoin node backlog here.

Additional measurements and events can be extracted in the future. Some ideas are in: https://github.com/0xB10C/peer-observer/issues/263

part of #141 
closes https://github.com/0xB10C/peer-observer/issues/212

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1464,6 +1464,14 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ff011a302c396a5197692431fc1948019154afc178baf7d8e37367442a4601cf"
 
 [[package]]
+name = "p2p-extractor"
+version = "0.1.0"
+dependencies = [
+ "shared",
+ "tokio-util",
+]
+
+[[package]]
 name = "parking_lot"
 version = "0.12.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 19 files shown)


0xB10C:
This still needs:

- [ ] A README.md
- [ ] metrics tool hooked up to the PingDuration event https://github.com/0xB10C/peer-observer/pull/256/files#diff-88ba1cf58a01e06348b30a2b23a924db1bfccc049bd9f99556fc39343e4b0f34R563

0xB10C:
I've been running this for a few days and it seems to work as intended.

PR #262, opened 2025-09-18
nix: update clang 14 -> 21

0xB10C:
On nixos-unstable:
 
    error: clang_14 has been removed, as it is unmaintained and obsolete

### DIFF
--- a/.github/workflows/rust-nix.yml
+++ b/.github/workflows/rust-nix.yml
@@ -20,13 +20,13 @@ jobs:
       with:
         nix_path: nixpkgs=channel:nixos-25.05
     - name: Build
-      run: nix-shell --command "cargo build --all-features"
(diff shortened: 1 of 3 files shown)


0xB10C:
Hm, so the Nix build works fine on NixOS with the kernel headers but doesn't work on Ubuntu. It still uses the Ubuntu kernel headers as revealed by `/usr/include/linux/errno.h:1:10`

0xB10C:
It was a fight, but

```patch
-      export KERNEL_HEADERS=$(nix eval --raw nixpkgs#linuxHeaders)/include
+      export KERNEL_HEADERS=${pkgs.linuxHeaders}/include
```

did the trick in the Ubuntu-base nix-shell build.

PR #272, opened 2025-09-24
Implement log extractor

m4ycon:


0xB10C:
Nice, looks good for a start! Sorry for the conflicts, I was refactoring some parts yesterday. Let me know if you need help rebasing it.

extractors/log/src/lib.rs
```diff
@@ -0,0 +1,112 @@
...
+    }
+}
+
+pub async fn run(args: Args, mut shutdown_rx: watch::Receiver<bool>) -> Result<(), RuntimeError> {
+    log::debug!("Connecting to NATS server at {}..", args.nats_address);
+    let nats_client = async_nats::connect(&args.nats_address).await?;
+    log::info!("Connected to NATS server at {}", &args.nats_address);
+
+    log::info!("Opening bitcoind log pipe...");
+    let file = File::open("bitcoind_pipe").await?;
```
0xB10C, line 49: You'll probably get to this at a later point: we'd probably want the location of the pipe to be configurable via an `Args` entry.

protobuf/log_extractor.proto
```diff
@@ -0,0 +1,23 @@
...
+package log_extractor;
+
+message LogEvent {
+  required uint64 timestamp = 1;
+  required string category = 2; // e.g., "validation", "net", etc.
+  oneof event {
+    UnknownLogMessage unknown_log_message = 3;
+    BlockConnectedLog block_connected_log = 4;
+  }
+}
```
m4ycon, line 12: What do you think about storing the "raw message" for known logs? Or when we gonna implement some known log, we get everything that is a variable in that string? I ask this because btc-core could change messages and maybe its interesting to store what it was at that moment🤔 

Also this is a mistery to me, on how to organize the LogMatcher to support many versions of btc-core, maybe many "case" matches from different versions of that log there, idk.

extractors/log/tests/integration.rs
```diff
@@ -0,0 +1,143 @@
...
+                            assert!(unknown_log_message.raw_message.len() > 0);
+                        }
+                        _ => (),
+                    }
+                }
+            }
+            _ => panic!("unexpected event {:?}", event),
+        };
+
+        assert!(received_events, "Did not receive any log events");
```
m4ycon, line 140: this test is just a POC hehe, probably there will be more tests related to string format... using bcli or another node to trigger something...

m4ycon:
Don't know why but ubuntu action throw at build with a unused value that it is used, and the nix one did build correctly ¯\\\_(ツ)_/¯

0xB10C:
I haven't forgotten about this and plan to get to it as soon as I have time!

> Don't know why but ubuntu action throw at build with a unused value that it is used, and the nix one did build correctly

I'll have a look!


Also, one thing that came to my mind was that I run my nodes with `logtimemicros=1` (see https://github.com/0xB10C/peer-observer-infra-library/blob/2c235ce06d594dd0731d4d78cc064bb78e1fc44e/modules/node/node.nix#L163) for higher precision timestamps in the logs. I think it makes sense to support this here too (or at least test that it's supported and doesn't break).

m4ycon:
> Also, one thing that came to my mind was that I run my nodes with `logtimemicros=1` (see https://github.com/0xB10C/peer-observer-infra-library/blob/2c235ce06d594dd0731d4d78cc064bb78e1fc44e/modules/node/node.nix#L163) for higher precision timestamps in the logs. I think it makes sense to support this here too (or at least test that it's supported and doesn't break).

Added support for this option, it was really going to not match with the regex that I made.

```proto
  // log_extractor.proto
  required uint64 log_timestamp = 1; // unix timestamp in seconds
  required uint32 log_timestamp_micro = 2; // microseconds part of the timestamp
```

I separatted in another field on proto because it doesn't have a `uint128` as returned by `OffsetDateTime::unix_timestamp_nanos` to store a microsecond unix time (or nano).

shared/src/log_matchers.rs
```diff
@@ -218,4 +237,49 @@ mod tests {
...
+    #[test]
+    fn test_log_matcher_with_broken_timestamp() {
+        let log = "2025--17T23:52:01.358911Z [validation] Random message";
+        let log_event = parse_log_event(log);
+
+        assert_eq!(log_event.log_timestamp, 0);
+        assert_eq!(log_event.category, LogDebugCategory::Unknown as i32);
+
+        if let Some(Event::UnknownLogMessage(unknown_log)) = log_event.event {
+            assert_eq!(unknown_log.raw_message, "");
```
m4ycon: empty `raw_message` 🤔

PR #312, opened 2025-12-17
Add `getchaintxstats` RPC to rpc-extractor, expose metrics

deadmanoz:
DRAFT REASON: This PR requires [rust-bitcoin/corepc#430](https://github.com/rust-bitcoin/corepc/pull/430) which fixes the `tx_rate` field type from `i64` to `f64` to match Bitcoin Core's actual return type. So this is currently draft - a small change to `rpc_extractor.rs` is required once upstream changes are merged (to remove `tx_rate` values being truncated to integers).

Add support for the `getchaintxstats` RPC to the `rpc-extractor` and expose the data as Prometheus metrics. This RPC allows for the below new metrics.

The `getchaintxstats` RPC computes statistics about the total number and rate of transactions in the chain. We use the Bitcoin Core defaults:
- **Window size**: ~1 month of blocks
- **End block**: Current chain tip

See: [Bitcoin Core RPC documentation](https://github.com/bitcoin/bitcoin/blob/master/src/rpc/blockchain.cpp#L1758)

| Metric | Description |
|--------|-------------|
| `rpc_chaintxstats_tx_count` | Total transactions in chain up to window end |
| `rpc_chaintxstats_window_final_block_height` | Height of final block in window |
| `rpc_chaintxstats_window_block_count` | Window size in blocks |
| `rpc_chaintxstats_window_tx_count` | Transactions in the window |
| `rpc_chaintxstats_window_interval` | Window duration in seconds |
| `rpc_chaintxstats_tx_rate` | Average tx/second in window (float) |

Note: `window_final_block_hash` is included in the protobuf event but not exposed as a Prometheus metric (block hashes are strings, not numeric values). I don't think a hash/string is a meaningful metric.

Changes include:
- Add `getchaintxstats` RPC polling to rpc-extractor with `--disable-getchaintxstats` flag
- Add `ChainTxStats` protobuf message definition
- Add Prometheus metrics for chain transaction statistics
- Add integration test for the new RPC
- Fix typo: 'quering' → 'querying' in help text

### DIFF
--- a/extractors/rpc/tests/integration.rs
+++ b/extractors/rpc/tests/integration.rs
@@ -169,32 +169,41 @@ async fn test_integration_rpc_getpeerinfo() {
 async fn test_integration_rpc_getmempoolinfo() {
     println!("test that we receive getmempoolinfo RPC events");
 
-    check(true, false, true, true, true, true, true, |event| match event {
-        PeerObserverEvent::RpcExtractor(r) => {
-            if let Some(ref e) = r.rpc_event {
-                match e {
-                    MempoolInfo(info) => {
-                        assert_eq!(info.loaded, true);
-                        assert_eq!(info.size, 0);
-                        assert_eq!(info.usage, 0);
-                        assert_eq!(info.bytes, 0);
-                        assert_eq!(info.total_fee, 0.0);
-                        assert_eq!(info.max_mempool, 300000000);
-                        // These will change between v29 and v30, so don't hardcode something here.
-                        assert!(info.mempoolminfee > 0.0);
-                        assert!(info.minrelaytxfee > 0.0);
-                        assert!(info.incrementalrelayfee > 0.0);
+    check(
+        true,
+        false,
+        true,
+        true,
+        true,
+        true,
+        true,
+        |event| match event {
+            PeerObserverEvent::RpcExtractor(r) => {
+                if let Some(ref e) = r.rpc_event {
+                    match e {
+                        MempoolInfo(info) => {
+                            assert_eq!(info.loaded, true);
+                            assert_eq!(info.size, 0);
+                            assert_eq!(info.usage, 0);
+                            assert_eq!(info.bytes, 0);
+                            assert_eq!(info.total_fee, 0.0);
+                            assert_eq!(info.max_mempool, 300000000);
+                            // These will change between v29 and v30, so don't hardcode something here.
+                            assert!(info.mempoolminfee > 0.0);
+                            assert!(info.minrelaytxfee > 0.0);
+                            assert!(info.incrementalrelayfee > 0.0);
 
-                        assert_eq!(info.unbroadcastcount, 0);
-                        assert_eq!(info.fullrbf, true);
-                        return;
+                            assert_eq!(info.unbroadcastcount, 0);
+                            assert_eq!(info.fullrbf, true);
+                            return;
+                        }
+                        _ => panic!("unexpected RPC data {:?}", r.rpc_event),
                     }
-                    _ => panic!("unexpected RPC data {:?}", r.rpc_event),
                 }
             }
-        }
-        _ => panic!("unexpected event {:?}", event),
-    })
+            _ => panic!("unexpected event {:?}", event),
+        },
+    )
     .await;
 }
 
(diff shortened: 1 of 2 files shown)


0xB10C:
Looks good to me! Could add integration tests for:

- the new RPC being supported in `extractors/rpc/tests/integration.rs`
- the new metrics added in `tools/metrics/tests/integration.rs`

feel free to add them in the same commit as the new feature 

I think https://github.com/rust-bitcoin/corepc/pull/430 isn't far from being merged. Going to wait until that's merged until I set this up for testing

GuiSchet:
yes, RPC fetchs every 10s, maybe other solution could be check last height

deadmanoz:
> I think this RPC is fetched every 10s, right? Not sure if it's needed that frequently. It will return the same result at the same height anyway? Could add a factor of 60 or so, so that it's only fetched every 10 minutes..

Yeah, fair point. 

I think having a slower rate (e.g. the multiplier) rather than some event (e.g. block hash/height change) is the way to go, we have a more deterministic data source. So I'll propose this!

deadmanoz:
Confirmed queries are ~10min apart. Currently erroring because of upstream bug:

```log
2025-12-18T03:51:04.519Z ERROR [rpc_extractor] Could not fetch and publish 'getchaintxstats': RPC error: JSON-RPC error: JSON decode error: invalid type: floating point `5.04260536630741`, expected i64 at line 1 column 272
...
2025-12-18T04:01:04.525Z ERROR [rpc_extractor] Could not fetch and publish 'getchaintxstats': RPC error: JSON-RPC error: JSON decode error: invalid type: floating point `5.04260536630741`, expected i64 at line 1 column 272
```

Metrics are available, again currently empty due to bug:

```bash
❯ curl -s http://127.0.0.1:8282/metrics | grep chaintxstats
# HELP peerobserver_rpc_chaintxstats_tx_count Total number of transactions in the chain up to the final block in the window.
# TYPE peerobserver_rpc_chaintxstats_tx_count gauge
peerobserver_rpc_chaintxstats_tx_count 0
# HELP peerobserver_rpc_chaintxstats_tx_rate Average rate of transactions per second in the window.
# TYPE peerobserver_rpc_chaintxstats_tx_rate gauge
peerobserver_rpc_chaintxstats_tx_rate 0
# HELP peerobserver_rpc_chaintxstats_window_block_count Size of the window in number of blocks.
# TYPE peerobserver_rpc_chaintxstats_window_block_count gauge
peerobserver_rpc_chaintxstats_window_block_count 0
# HELP peerobserver_rpc_chaintxstats_window_final_block_height Height of the final block in the window.
# TYPE peerobserver_rpc_chaintxstats_window_final_block_height gauge
peerobserver_rpc_chaintxstats_window_final_block_height 0
# HELP peerobserver_rpc_chaintxstats_window_interval Elapsed time in the window in seconds.
# TYPE peerobserver_rpc_chaintxstats_window_interval gauge
peerobserver_rpc_chaintxstats_window_interval 0
# HELP peerobserver_rpc_chaintxstats_window_tx_count Number of transactions in the window.
# TYPE peerobserver_rpc_chaintxstats_window_tx_count gauge
peerobserver_rpc_chaintxstats_window_tx_count 0
```

0xB10C:
Did a review pass on c98d84ae2909fadd4c7190f0d3588dcbcf24ab5d and code looked good. I'm afraid you have to rebase on #309 now.

shared/src/protobuf/rpc_extractor.rs
```diff
@@ -232,3 +234,32 @@ impl From<RPCAddrManInfoNetwork> for AddrManInfoNetwork {
...
+            window_final_block_hash: stats.window_final_block_hash,
+            window_final_block_height: stats.window_final_block_height,
+            window_block_count: stats.window_block_count,
+            window_tx_count: stats.window_tx_count,
+            window_interval: stats.window_interval,
+            // Note: Bitcoin Core returns txrate as a float,
+            // but corepc deserialises it as i64!
+            // The following will be changed once the upstream PR is merged:
+            // https://github.com/rust-bitcoin/corepc/pull/430
+            tx_rate: stats.tx_rate.map(|r| r as f64),
```
0xB10C: This comment can be removed now?

extractors/rpc/tests/integration.rs
```diff
@@ -300,3 +304,38 @@ async fn test_integration_rpc_getaddrmaninfo() {
...
+                        assert!(stats.window_block_count >= 0);
+
+                        // Optional fields - only present if window_block_count > 0
+                        if stats.window_block_count > 0 {
+                            assert!(stats.window_tx_count.is_some());
+                            assert!(stats.window_interval.is_some());
+                            // And tx_rate is only present if window_interval > 0
+                            if stats.window_interval.unwrap_or(0) > 0 {
+                                assert!(stats.tx_rate.is_some());
+                            }
```
0xB10C: I was wondering if we hit this in the test?

deadmanoz:
FWIW here's a visualisation of the `tx_rate` metric (Grafana dashboard whipped up to check)

Screenshot 2026-01-06 at 5 12 35 pm

Is the build error transient? Or did they pull 30.0 and 30.1 because of the legacy wallet bug?

0xB10C:
Yeah, I think we can ignore the Ubuntu job for now. Could also add in our own binary again

0xB10C (approved):
LGTM!

PR #318, opened 2025-12-23
clean: fix `cargo clippy` warnings automatically and manually

0xB10C:
This addresses `cargo clippy` warnings both by running with `--fix` and addressing the remaining ones manually. This should allow us to add a CI check for this eventually.


0xB10C:
draft since not all manual warnings have been addressed

PR #326, opened 2025-12-30
add: lint job to CI with cargo clippy

0xB10C:
following up on #318 and #325


0xB10C:
Going with a Ubuntu lint job now since I couldn't get the NixOS one to work in CI and can't reproduce the failure locally.

Issue #201, opened 2025-08-01
rpc-extractor: offer RPC fetch timings and errors via a prometheus server


0xB10C:
#191 logs RPC fetch errors, but these can easily be missed. Since we use Prometheus in tools already, we can use the `metricserver.rs` from `shared` in the rpc-extractor too.

- metric for RPC fetch errors
- metric for NATS publish errors
- metric for RPC fetch times (milliseconds it took for us to get the result from an RPC). this is a good indication of a node being overloaded
- ...

0xB10C:
Feel free to open a followup issue (or PR)!

PR #339, opened 2026-01-19
feat: add prometheus metrics and integration tests to rpc-extractor

GuiSchet:
PR Description:
## Summary
This PR implements Prometheus metrics for the `rpc-extractor` and adds a robust suite of integration tests. 
## Key Changes
**Metrics Instrumentation**: Added Prometheus metrics to track RPC fetch durations and error rates (both for RPC calls and NATS publishing).
**Configuration**: Introduced a new `--prometheus-address` CLI argument (defaulting to `0.0.0.0:8284`) to serve the metrics.
**Enhanced Testing**:
    - Added unit tests for the metrics registration and logic.
    - Implemented 14 integration tests.
    - Verified that all enabled RPC methods (e.g., `getpeerinfo`, `getmempoolinfo`, `uptime`, etc.) correctly publish events.
    - Verified that Prometheus correctly records durations and increments error counters on failures (unreachable host, invalid auth).
   
   ## Related Issues
   #201

### DIFF
--- a/extractors/rpc/tests/common/mod.rs
+++ b/extractors/rpc/tests/common/mod.rs
@@ -1,0 +1,170 @@
+use shared::{
+    corepc_node,
+    log::{self, info},
+    nats_util::NatsArgs,
+    simple_logger::SimpleLogger,
+};
+
+use std::net::TcpListener;
+use std::sync::Once;
+
+use rpc_extractor::Args;
+
+/// Get an available port for the metrics server.
+pub fn get_available_port() -> u16 {
+    TcpListener::bind("127.0.0.1:0")
+        .expect("Failed to bind to port 0")
+        .local_addr()
+        .expect("Failed to get local addr")
+        .port()
+}
+
+static INIT: Once = Once::new();
+
+// 1 second query interval for fast tests
+pub const QUERY_INTERVAL_SECONDS: u64 = 1;
+
+pub fn setup() {
+    INIT.call_once(|| {
+        SimpleLogger::new()
+            .with_level(log::LevelFilter::Trace)
+            .init()
+            .unwrap();
+    });
+}
+
+#[derive(Default)]
+pub struct EnabledRPCsInTest {
+    pub getpeerinfo: bool,
+    pub getmempoolinfo: bool,
+    pub uptime: bool,
+    pub getnettotals: bool,
+    pub getmemoryinfo: bool,
+    pub getaddrmaninfo: bool,
+    pub getchaintxstats: bool,
+    pub getnetworkinfo: bool,
+    pub getblockchaininfo: bool,
+    pub getorphantxs: bool,
+    pub getrawaddrman: bool,
+}
+
+#[allow(dead_code)]
+impl EnabledRPCsInTest {
+    /// Returns an instance with all RPC methods enabled.
+    pub fn all() -> Self {
+        Self {
+            getpeerinfo: true,
+            getmempoolinfo: true,
+            uptime: true,
+            getnettotals: true,
+            getmemoryinfo: true,
+            getaddrmaninfo: true,
+            getchaintxstats: true,
+            getnetworkinfo: true,
+            getblockchaininfo: true,
+            getorphantxs: true,
+            getrawaddrman: true,
+        }
+    }
+
+    /// Returns the names of the enabled RPC methods.
+    /// no need to update a separate list when adding new RPCs.
+    pub fn enabled_methods(&self) -> Vec<&'static str> {
+        let mut methods = Vec::new();
+        if self.getpeerinfo {
+            methods.push("getpeerinfo");
+        }
+        if self.getmempoolinfo {
+            methods.push("getmempoolinfo");
+        }
+        if self.uptime {
+            methods.push("uptime");
+        }
+        if self.getnettotals {
+            methods.push("getnettotals");
+        }
+        if self.getmemoryinfo {
+            methods.push("getmemoryinfo");
+        }
+        if self.getaddrmaninfo {
+            methods.push("getaddrmaninfo");
+        }
+        if self.getchaintxstats {
+            methods.push("getchaintxstats");
+        }
+        if self.getnetworkinfo {
+            methods.push("getnetworkinfo");
+        }
+        if self.getblockchaininfo {
+            methods.push("getblockchaininfo");
+        }
+        if self.getorphantxs {
+            methods.push("getorphantxs");
+        }
+        if self.getrawaddrman {
+            methods.push("getrawaddrman");
+        }
+        methods
+    }
+}
+
+pub fn make_test_args(
+    nats_port: u16,
+    rpc_url: String,
+    cookie_file: String,
+    prometheus_address: String,
+    rpcs: EnabledRPCsInTest,
+) -> Args {
+    Args::new(
+        NatsArgs {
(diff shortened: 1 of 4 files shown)


extractors/rpc/src/lib.rs
```diff
@@ -267,27 +282,54 @@ pub async fn run(args: Args, mut shutdown_rx: watch::Receiver<bool>) -> Result<(
...
 }
 
+fn measure_rpc_call<T, E, F>(method_name: &str, metrics: &Metrics, f: F) -> Result<T, E>
+where
+    F: FnOnce() -> Result<T, E>,
+{
+    let _timer = metrics
+        .rpc_fetch_duration
+        .with_label_values(&[method_name])
+        .start_timer();
```
0xB10C: doesn't this also need `timer.stop_and_record()`?

https://docs.rs/prometheus/latest/prometheus/struct.HistogramTimer.html#method.stop_and_record ?

extractors/rpc/Cargo.toml
```diff
@@ -6,6 +6,9 @@ edition = "2024"
 [dependencies]
 shared = { path = "../../shared" }
 
+[dev-dependencies]
+ureq = "2.9"
+
```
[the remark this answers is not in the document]
deadmanoz: There are no longer any dev-dependencies, so this can be removed?

extractors/rpc/src/lib.rs
deadmanoz: Any reason this is not instrumented?

shared/src/testing/metrics_fetcher.rs
```diff
@@ -0,0 +1,76 @@
...
+    stream.read_to_end(&mut response)?;
+    Ok(String::from_utf8_lossy(&response).to_string())
+}
+
+/// Convenience wrapper that fetches from the root path "/".
+pub fn fetch_metrics_root(port: u16) -> Result<String, std::io::Error> {
+    fetch_metrics(port, "/")
+}
+
+/// Extracts the count value from a histogram metric.
```
deadmanoz: Is there much difference between these 2 methods?

one uses `starts_with`, the other uses `contains`

Seems like we could consolidate into a single method?

shared/src/testing/metrics_fetcher.rs
```diff
@@ -0,0 +1,76 @@
...
+
+/// Extracts the value of a counter metric with a specific label.
+///
+/// Searches for a line starting with `{metric_name}{rpc_method="{label_value}"}` and
+/// returns the numeric value at the end of that line.
+///
+/// # Panics
+/// Panics if the metric is not found or the value cannot be parsed.
+pub fn get_counter_value(metrics_raw: &str, metric_name: &str, label_value: &str) -> u64 {
+    let search_pattern = format!("{}{{rpc_method=\"{}\"}}", metric_name, label_value);
```
0xB10C: These are only usable for RPC-extractor specific code with `rpc_method`

extractors/rpc/tests/metrics_integration.rs
```diff
@@ -0,0 +1,416 @@
...
+    node1_conf.p2p = corepc_node::P2P::Yes;
+    let node1 = setup_node(node1_conf);
+
+    // node2 connects to node1
+    let mut node2_conf = corepc_node::Conf::default();
+    node2_conf.p2p = node1.p2p_connect(true).unwrap();
+    let node2 = setup_node(node2_conf);
+
+    (node1, node2)
+}
```
0xB10C: This seems to be duplicate with the code in integration.rs? How about having a lib.rs for this?

extractors/rpc/tests/metrics_integration.rs
```diff
@@ -0,0 +1,310 @@
...
+use shared::{
+    testing::{
+        metrics_fetcher::{fetch_metrics, get_metric_value},
+        nats_server::NatsServerForTesting,
+    },
+    tokio::{self, sync::watch},
+};
+
+#[tokio::test]
+async fn test_integration_metrics_server_basic() {
```
deadmanoz: Can you please provide a brief comment for each test describing what it does/what it's purpose is?

deadmanoz: e.g. the sequence that happens within

extractors/rpc/tests/common/mod.rs
```diff
@@ -0,0 +1,110 @@
...
+    pub getnettotals: bool,
+    pub getmemoryinfo: bool,
+    pub getaddrmaninfo: bool,
+    pub getchaintxstats: bool,
+    pub getnetworkinfo: bool,
+    pub getblockchaininfo: bool,
+    pub getorphantxs: bool,
+    pub getrawaddrman: bool,
+}
+
```
deadmanoz: I think you should add the following here:

```
  impl EnabledRPCsInTest {
      pub fn all() -> Self {
          Self {
              getpeerinfo: true,
              getmempoolinfo: true,
              uptime: true,
              getnettotals: true,
              getmemoryinfo: true,
              getaddrmaninfo: true,
              getchaintxstats: true,
              getnetworkinfo: true,
              getblockchaininfo: true,
              getorphantxs: true,
              getrawaddrman: true,
          }
      }

      pub fn enabled_methods(&self) -> Vec<&'static str> {
          let mut methods = Vec::new();
          if self.getpeerinfo { methods.push("getpeerinfo"); }
          if self.getmempoolinfo { methods.push("getmempoolinfo"); }
          if self.uptime { methods.push("uptime"); }
          if self.getnettotals { methods.push("getnettotals"); }
          if self.getmemoryinfo { methods.push("getmemoryinfo"); }
          if self.getaddrmaninfo { methods.push("getaddrmaninfo"); }
          if self.getchaintxstats { methods.push("getchaintxstats"); }
          if self.getnetworkinfo { methods.push("getnetworkinfo"); }
          if self.getblockchaininfo { methods.push("getblockchaininfo"); }
          if self.getorphantxs { methods.push("getorphantxs"); }
          if self.getrawaddrman { methods.push("getrawaddrman"); }
          methods
      }
  }
```

This would mean that one place needs updating as new RPCs are added. We don't need to update the tests too, they get it by virtue of using `EnabledRPCsInTest::all()` and EnabledRPCsInTest::enabled_methods()`!!

deadmanoz (changes requested):
Almost there, just a few more things to consider/address!

shared/src/testing/metrics_fetcher.rs
```diff
@@ -0,0 +1,61 @@
...
+    stream.write_all(request.as_bytes())?;
+    stream.flush()?;
+
+    let mut response = Vec::new();
+    stream.read_to_end(&mut response)?;
+    Ok(String::from_utf8_lossy(&response).to_string())
+}
+
+/// Convenience wrapper that fetches from the root path "/".
+pub fn fetch_metrics_root(port: u16) -> Result<String, std::io::Error> {
```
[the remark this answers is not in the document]
0xB10C: I did that on purpose. It just serves metrics on all paths. On `/`, `/metrics`, `/abc`, all should return the same thing. I don't want to parse the path in there.

That said, probably a comment about this would have helped to explain it :)

deadmanoz:
Nice, thanks @GuiSchet. I think the only thing that remains is the redundant `fetch_metrics_root`, have left a comment. It's not blocking but maybe is worth tightening

PR #346, opened 2026-01-30
change: fetch `getrawaddrman` less frequently

0xB10C:
and make the interval configurable.

Also, this changed it from default 600s to 120s.


0xB10C:
With fetching `getrawaddrman` every 10s, I noticed significantly higher resource usage (CPU and RAM) on some nodes. We don't need it every 10s. Every 120s is fine too.

PR #358, opened 2026-02-13
fix(p2p): use OS-assigned port to fix intermittent test failures

deadmanoz:
## Summary

Use port 0 (OS-assigned) for the p2p-extractor TCP listener in integration tests instead of manually picking random ephemeral ports. 

https://docs.rs/tokio/latest/tokio/net/struct.TcpListener.html#method.bind

> "Binding with a port number of 0 will request that the OS assigns a port to this listener. The port allocated can be queried via the `local_addr` method."

This is race-free by design, there's no window between choosing a port and binding it where another process could grab it, so this eliminates the race condition that caused intermittent "Address already in use" failures (#316). 

The `run()` function now accepts an optional `oneshot::Sender<SocketAddr>` that reports the actual bound address back to the caller. This also replaces the previous 2-second sleep with a proper synchronisation barrier via `tokio::select!`, making tests both faster and more reliable. If the extractor fails before binding (e.g. NATS connection error), the real error is surfaced instead of a generic channel-closed panic.

Fixes #316.

## Testing

The fix was validated in my local testing setup with a before/after test script. To reliably reproduce the race condition on the old code, `setup()` was patched to return the same port for all parallel tests (changing `fetch_add` to `load`), forcing every test to compete for a single port:

```bash
sed -i 's/\.fetch_add(1, Ordering::SeqCst)/.load(Ordering::SeqCst)/' \
    extractors/p2p/tests/integration.rs
```

Both phases ran the integration test suite 5 times with `--test-threads=4`:

| | AddrInUse failures | Other failures |
|---|---|---|
| **Old code** (patched, same port) | 5/5 | — |
| **Fixed code** (OS-assigned) | 0/5 | 1/5 (split INV messages as per #357, another fix incoming!) |

PR #365, opened 2026-02-24
fix(websocket): use OS-assigned port to fix intermittent test failures

### LINKED ISSUE #327: p2p-extractor: in integration test, retry `addconnection` each time
We should call add_connection in each iteration of this loop, with a short wait in between `add_connection` and checking if the connection is established. 

https://github.com/peer-observer/peer-observer/blob/0cda18ca719e8f873e2ae7aaec34c892a992dbaa/extractors/p2p/tests/integration.rs#L157-L183

deadmanoz:
## Summary

Use port 0 (OS-assigned) for the websocket TCP listener in integration tests instead of manually picking random ephemeral ports.

Same approach as the P2P extractor fix in #358. The old code used `AtomicU16` with a random ephemeral start port, a retry loop on `AddrInUse`, and a 1-second sleep to wait for the server to start. This had the same race condition, a window between choosing a port and binding it where another process could grab it.

The `run()` function now accepts an optional `oneshot::Sender<SocketAddr>` that reports the actual bound address back to the caller. Tests bind to port 0, receive the OS-assigned address via the oneshot, and connect clients immediately (no sleep, no race).

This PR also adds a **pipeline readiness barrier** (as was introduced with https://github.com/peer-observer/peer-observer/pull/372) to eliminate the following race: after a client connects and sends its subscription JSON, there's a window where NATS events can arrive at the server before the client is registered in the broadcast map with its subscriptions applied. 

As each client is inserted and subscribed independently, one client receiving events does not prove the others are ready. So the barrier publishes a marker event through NATS for *every* subscribing client and waits for it to arrive on that client's websocket stream, thus proving the full pipeline is live. 

The marker event type matches the client's subscription (just messages or mempool for now, because only those were present in the tests) because `broadcast_to_clients` filters by protobuf variant. A final drain pass flushes any cross-contamination markers between clients before real test events are published.

Fixes #327.

## Testing

All 6 integration tests pass a 50-run stress testing on Linux (NixOS x86_64), zero failures.

### DIFF
--- a/tools/websocket/src/error.rs
+++ b/tools/websocket/src/error.rs
@@ -1,4 +1,5 @@
 use shared::async_nats;
+use shared::async_nats::client::FlushErrorKind;
 use shared::async_nats::ConnectErrorKind;
 use shared::log::SetLoggerError;
 use shared::prost::DecodeError;
@@ -13,6 +14,7 @@ pub enum RuntimeError {
     ProtobufDecode(DecodeError),
     NatsSubscribe(async_nats::client::SubscribeError),
     NatsConnect(shared::async_nats::error::Error<ConnectErrorKind>),
+    NatsFlush(shared::async_nats::error::Error<FlushErrorKind>),
 }
 
 impl fmt::Display for RuntimeError {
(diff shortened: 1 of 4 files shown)


0xB10C:
I re-triggered CI on the job where GitHub seemed like it broke.

PR #354, opened 2026-02-11
fix(tests): move category assertions into guards to prevent log event ordering races

deadmanoz:
## Summary

- Move `LogDebugCategory::Net` assertions from inside `check_event` closures into `if let` guards so tests filter for the expected category rather than asserting on the first `UnknownLogMessage` received
- Document `setup_node`, `setup_two_connected_nodes`, and `check` test harness functions

Fixes https://github.com/peer-observer/peer-observer/issues/338

## Context

On slower CPUs (e.g. my trusty old Xeon E5-2630 v4 or even GitHub's own runners), bitcoind takes longer to initialise. This caused an event ordering race.

Three tests (`unknown_with_category`, `unknown_with_threadname_and_category`, `unknown_with_all_metadata`) matched the first `UnknownLogMessage` and asserted `category == Net`. On slower hardware, non-net messages (e.g. `dnsseed thread start` with category `Unknown`) arrive before any `[net]` line, causing assertion failures. Moving the category check into the `if let` guard lets the event loop skip non-matching messages and continue scanning until a `Net` category event arrives. If the log extractor breaks its category parsing, no event will ever match and the test will time out (preserves the original intent).

Here's a snippet from from logs identifying the failure of the reported issue (https://github.com/peer-observer/peer-observer/issues/338). Note that the test expects category `Net`, but the first `UnknownLogMessage` is from the `dnsseed` thread with category `Unknown`!

```
2026-02-11T02:04:28.780Z INFO  [log_extractor] Started reading lines from bitcoind log pipe at /tmp/nix-shell-319968-3831661800/.tmpocGJcs/bitcoind_pipe
2026-02-11T02:04:28.806Z TRACE [log_extractor] Read log line: 2026-02-11T02:04:27Z [dnsseed] dnsseed thread start
2026-02-11T02:04:28.806Z TRACE [shared::protobuf::event] creating new Event: LogExtractor(Log { log_timestamp: 1770775467000000, category: Unknown, threadname: "dnsseed", log_event: Some(UnknownLogMessage(UnknownLogMessage { raw_message: "dnsseed thread start" })) })
2026-02-11T02:04:28.806Z TRACE [log_extractor] published log into NATS: Event { timestamp: 1770775468806, peer_observer_event: Some(LogExtractor(Log { log_timestamp: 1770775467000000, category: Unknown, threadname: "dnsseed", log_event: Some(UnknownLogMessage(UnknownLogMessage { raw_message: "dnsseed thread start" })) })) }
[...]

Issue #338, opened 2026-01-16
Intermittent integration test failure of test_integration_logextractor_unknown_with_category


0xB10C:
from https://github.com/peer-observer/peer-observer/actions/runs/21064896594/job/60580105023?pr=337
```
---- test_integration_logextractor_unknown_with_category stdout ----
test that we receive unknown log events with category
2026-01-16T11:31:44.747Z INFO  [integration] Running node1 with arg: -debug=net
2026-01-16T11:31:44.747Z INFO  [integration] env BITCOIND_EXE=Ok("/nix/store/rqhjkp8ggnfbrjw9xd3kfxm2qszdv2np-bitcoind-30.0/bin/bitcoind")
2026-01-16T11:31:44.747Z INFO  [integration] exe_path=Ok("/nix/store/rqhjkp8ggnfbrjw9xd3kfxm2qszdv2np-bitcoind-30.0/bin/bitcoind")
2026-01-16T11:31:44.747Z INFO  [integration] Using bitcoind at '/nix/store/rqhjkp8ggnfbrjw9xd3kfxm2qszdv2np-bitcoind-30.0/bin/bitcoind'
2026-01-16T11:31:44.948Z DEBUG [corepc] request: createwallet ["default"]
2026-01-16T11:31:47.056Z TRACE [corepc] response for createwallet: {"name":"default"}
2026-01-16T11:31:47.056Z DEBUG [corepc] request: getblockchaininfo []
2026-01-16T11:31:47.072Z TRACE [corepc] response for getblockchaininfo: {"chain":"regtest","blocks":0,"headers":0,"bestblockhash":"0f9188f13cb7b2c71f2a335e3a4fc328bf5beb436012afca590b1a11466e2206","bits":"207fffff","target":"7fffff0000000000000000000000000000000000000000000000000000000000","difficulty":4.656542373906925e-10,"time":1296688602,"mediantime":1296688602,"verificationprogress":2.119203030044969e-06,"initialblockdownload":true,"chainwork":"0000000000000000000000000000000000000000000000000000000000000002","size_on_disk":293,"pruned":false,"warnings":[]}
2026-01-16T11:31:47.072Z INFO  [integration] env BITCOIND_EXE=Ok("/nix/store/rqhjkp8ggnfbrjw9xd3kfxm2qszdv2np-bitcoind-30.0/bin/bitcoind")
[...]

PR #369, opened 2026-02-25
Fix/log level filter and state guards

### LINKED ISSUE #366: test flakiness uncovered by nix-coverage job
I got the following failure in the `nix-coverage` job (not related to the websocket changes, all tests pass):

https://github.com/peer-observer/peer-observer/actions/runs/22336981723/job/64632027203

```
thread 'test_integration_logextractor_mutated_block_bad_witness_nonce_size' (13212) panicked at extractors/log/tests/integration.rs:172:13:
timed out waiting for check() to complete
```

The log lines arrive with `threadname: "error"` and `category: Unknown`:

```
2026-02-24T04:54:50.618Z TRACE [log_extractor] Read log line: 2026-02-24T04:54:50Z [error] AcceptBlock: bad-witness-nonce-size, CheckWitnessMalleation : invalid witness reserved value size
2026-02-24T04:54:50.618Z TRACE [shared::protobuf::event] creating new Event: LogExtractor(Log { log_timestamp: 1771908890000000, category: Unknown, threadname: "error", log_event: Some(UnknownLogMessage(UnknownLogMessage { raw_message: "AcceptBlock: bad-witness-nonce-size, CheckWitnessMalleation : invalid witness reserved value size" })) })
2026-02-24T04:54:50.624Z TRACE [log_extractor] Read log line: 2026-02-24T04:54:50Z [error] ProcessNewBlock: AcceptBlock FAILED (bad-witness-nonce-size, CheckWitnessMalleation : invalid witness reserved value size)
```

The `check()` never matches these events, so the test times out. Same class of event-ordering issue as #338 I think?

deadmanoz:
## Summary

The log line parser in `shared/src/log_matchers.rs` treats all `[bracketed]` metadata tokens as either thread names or debug categories. This breaks when Bitcoin Core emits log level markers via `LogError()` and `LogWarning()`, which produce lines like:

```
2025-10-02T02:31:14Z [error] AcceptBlock: bad-witness-nonce-size, ...
2025-10-02T02:31:14Z [msghand] [error] ConnectBlock: ...
```

The parser would interpret `[error]` as a thread name (in the first case) or a debug category (in the second). This is incorrect - it's a severity level emitted by `GetLogPrefix()` in Bitcoin Core's [`logging.cpp`](https://github.com/bitcoin/bitcoin/blob/master/src/logging.cpp#L343-L366).

When `-logthreadnames=1` is enabled, the thread name bracket and the level bracket appear as two separate tokens (e.g. `[msghand] [error]` as above), so both cases need handling.

This is the root cause behind the flaky `BlockCheckedLog` test failures reported in #366. The nix-coverage CI job captured logs showing `threadname: "error",` with `[error]` having been misinterpreted as a thread name! I also hit this locally while working on #338.

This PR adds an `is_standalone_log_level()` filter that strips `[error]` and `[warning]` from the parsed metadata items before thread name/category assignment. The function is scoped to only these two levels because they are the only ones that appear as standalone bracket tokens under default Bitcoin Core configuration (see the linked `logging.cpp`): 
- `LogInfo()` produces no bracket at all (`GetLogPrefix` returns empty for uncategorised info-level messages)
- `LogDebug()`/`LogTrace()` both require a category argument so they produce `[category]` or `[category:trace]` as single bracket tokens, never standalone `[debug]` or `[trace]`

In a separate commit, the integration tests for `BlockCheckedLog` events are tightened using the same pattern as #354: the state value check (e.g. `"Valid"`, `"bad-witness-nonce-size"`, `"bad-txnmrklroot"`) is moved from an `assert_eq!` inside the match body into an `if-let` guard condition. This prevents the callback from panicking on a legitimate but unexpected state value arriving first (it skips non-matching events and waits for the correct one).

Fixes #366 🥳 

## Testing

Four new tests in `shared/src/log_matchers.rs`:

- `test_log_matcher_error_level_not_treated_as_threadname` - standalone `[error]` is filtered, not assigned as thread name
- `test_log_matcher_error_level_with_threadname` - `[msghand] [error]` preserves the real thread name and filters the level
- `test_log_matcher_warning_level_filtered` - standalone `[warning]` is filtered
- `test_is_standalone_log_level` - exercises the filter against known levels, non-standalone levels (`info`, `debug`, `trace`), and known non-levels (categories and thread names like `net`, `validation`, `msghand`)

Existing integration tests (`test_integration_logextractor_block_checked`, `test_integration_logextractor_mutated_block_bad_witness_nonce_size`, `test_integration_logextractor_mutated_block_bad_txnmrklroot`) continue to pass with the guard-style predicates. 

Full pre-push quality gate (fmt, clippy, unit tests, NATS integration tests) passes.

See green CI via PR opened against my fork: https://github.com/deadmanoz/peer-observer/actions/runs/22381839058

### DIFF
--- a/extractors/log/tests/integration.rs
+++ b/extractors/log/tests/integration.rs
@@ -349,9 +349,10 @@ async fn test_integration_logextractor_block_checked() {
         |event| {
             match event {
                 PeerObserverEvent::LogExtractor(r) => {
-                    if let Some(log::LogEvent::BlockCheckedLog(block_checked)) = r.log_event {
+                    if let Some(log::LogEvent::BlockCheckedLog(block_checked)) = r.log_event
+                        && block_checked.state == "Valid"
+                    {
                         assert!(!block_checked.block_hash.is_empty());
-                        assert_eq!(block_checked.state, "Valid");
                         info!("BlockCheckedLog event {}", block_checked);
                         return true;
                     }
@@ -400,8 +401,8 @@ async fn test_integration_logextractor_mutated_block_bad_witness_nonce_size() {
                 PeerObserverEvent::LogExtractor(r) => {
                     if let Some(ref e) = r.log_event
                         && let log::LogEvent::BlockCheckedLog(block_checked) = e
+                        && block_checked.state == "bad-witness-nonce-size"
                     {
-                        assert_eq!(block_checked.state, "bad-witness-nonce-size");
                         assert_eq!(
                             block_checked.debug_message,
                             "CheckWitnessMalleation : invalid witness reserved value size"
@@ -455,8 +456,8 @@ async fn test_integration_logextractor_mutated_block_bad_txnmrklroot() {
                 PeerObserverEvent::LogExtractor(l) => {
                     if let Some(ref e) = l.log_event
                         && let log::LogEvent::BlockCheckedLog(block_checked) = e
+                        && block_checked.state == "bad-txnmrklroot"
                     {
-                        assert_eq!(block_checked.state, "bad-txnmrklroot");
                         assert_eq!(block_checked.debug_message, "hashMerkleRoot mismatch");
                         info!("BlockCheckedLog event {}", block_checked);
                         return true;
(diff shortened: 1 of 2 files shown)


m4ycon (approved):
utACK

PR #373, opened 2026-03-01
add(archiver, replayer): archive and replay peer-observer events

### LINKED ISSUE #121: Archiver tool - archive events for future analysis and processing
An archiver tool (something that archives events to disk) for future analysis and processing has been on my wish list for a while.

It should support and consider:
- storing events to disk in a chronological order with event timestamps
- chunking in some why - maybe daily chunks of messages
- possibly filtering by event type (p2p messages, connections, mempool events, ...)
- being somewhat storage space efficient, also with chunking: compressing daily chunks afterwards might be preferred  
- while not a part of the archiver, having a way of playing the events back in real-time would be good
- gracefully allow new event types if we add them
- ...

Two ideas:

1. The messages are exchanged between the `extractor` and tools in protobuf messages. One idea would be to serialize these messages as bytes to a file and to read them by deserializing from the file. Not sure if the individual messages need a separator or size-indicator between them. That'd be open research.
2. ~The preferred~ Another option would be to explore NATS [JetStreams](https://docs.nats.io/nats-concepts/jetstream) for this. This is NATS built-in persistence engine and can be used to replay messages. A few open questions remain: What data format is used? Can these be chunked? Is filtering possible? (edit: 2025-07-15: initial research suggests that the JetStream files aren't as probable as initially hoped Maybe option 1 is better).

octaviolucca:
Closes #121 

Archiver tool that subscribes to NATS and persists events to binary archive files (.bin.zst) with file rotation. 
Supports zstd compression, a TOML manifest, and event type filtering. Includes a **minimal replayer tool** that reads archives and prints decoded events, just enough to enable the roundtrip integration test. Further replayer work (e.g. NATS republishing) will come after this PR is reviewed.

File format: fixed 16-byte header followed by length-delimited protobuf events, with the full stream (header + events) compressed via zstd.
Header: [ 2B magic "PA" ][ 1B version ][ 4B  truncated git commit id ][ 9B reserved ]

Manifest semantics: checksum and size_bytes describe the decompressed archive stream (after zstd decompression). This is intentional: it allows verifying data integrity after decompression. If we later need metadata for the compressed bytes on disk (e.g., compressed size or checksum), we can add separate fields.

### DIFF
--- a/tools/archiver/Cargo.toml
+++ b/tools/archiver/Cargo.toml
@@ -8,6 +8,7 @@ shared = { path = "../../shared" }
 serde = { version = "1", features = ["derive"] }
 toml = "0.8"
 sha2 = "0.10"
+zstd = "0.13"
 
 [features]
 strict = []
--- a/tools/archiver/Cargo.toml
+++ b/tools/archiver/Cargo.toml
@@ -1,0 +1,11 @@
+[package]
+name = "archiver"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+shared = { path = "../../shared" }
+
+[features]
+strict = []
+nats_integration_tests = []
--- a/tools/archiver/Cargo.toml
+++ b/tools/archiver/Cargo.toml
@@ -5,6 +5,9 @@ edition = "2021"
 
 [dependencies]
 shared = { path = "../../shared" }
+serde = { version = "1", features = ["derive"] }
+toml = "0.8"
+sha2 = "0.10"
 
 [features]
 strict = []
--- a/tools/archiver/Cargo.toml
+++ b/tools/archiver/Cargo.toml
@@ -13,6 +13,7 @@ zstd = "0.13"
 [dev-dependencies]
 sha2 = "0.10"
 toml = "0.8"
+replayer = { path = "../replayer" }
 
 [features]
 # Treat warnings as a build error.
--- a/tools/archiver/Cargo.toml
+++ b/tools/archiver/Cargo.toml
@@ -10,6 +10,12 @@ toml = "0.8"
 sha2 = "0.10"
 zstd = "0.13"
 
+[dev-dependencies]
+sha2 = "0.10"
+toml = "0.8"
+
 [features]
+# Treat warnings as a build error.
 strict = []
+# Run integration tests needing a NATS server.
 nats_integration_tests = []
--- a/tools/archiver/Cargo.toml
+++ b/tools/archiver/Cargo.toml
@@ -5,7 +5,6 @@ edition = "2021"
 
 [dependencies]
 shared = { path = "../../shared" }
-serde = { version = "1", features = ["derive"] }
 toml = "0.8"
 sha2 = "0.10"
 zstd = "0.13"
--- a/tools/archiver/Cargo.toml
+++ b/tools/archiver/Cargo.toml
@@ -5,6 +5,7 @@ edition = "2021"
 
 [dependencies]
 shared = { path = "../../shared" }
+time = { version = "0.3", features = ["formatting", "macros"] }
 toml = "0.8"
 sha2 = "0.10"
 
--- a/tools/archiver/Cargo.toml
+++ b/tools/archiver/Cargo.toml
@@ -7,7 +7,6 @@ edition = "2021"
 shared = { path = "../../shared" }
 toml = "0.8"
 sha2 = "0.10"
-zstd = "0.13"
 
 [dev-dependencies]
 sha2 = "0.10"
--- a/tools/archiver/README.md
+++ b/tools/archiver/README.md
@@ -1,0 +1,105 @@
+# `archiver` tool
+
+> archives peer-observer events to disk
+
+A peer-observer tool that subscribes to a NATS server and persists events to binary files on disk.
+By default, all event types are archived. Events can be filtered by type using flags, allowing
+multiple archiver instances to run simultaneously for different recording jobs.
+
+## File format
+
+Events are stored as sequential length-delimited protobuf messages (using `encode_length_delimited`
+from `prost`), preceded by a 16-byte header:
+
+```
+[ 2B magic "PA" ][ 1B version ][ 4B git commit hash ][ 9B reserved ]
+[ varint length ][ protobuf Event bytes ]
+[ varint length ][ protobuf Event bytes ]
+...
+```
+
+The git commit hash in the header identifies the exact version of the protobuf definitions used
+to record the file, allowing future readers to check out that commit if needed.
+
+## Manifest
+
+Each recording session produces a `<base-name>.manifest.toml` file alongside the archive files.
+The manifest is updated on every file rotation and on shutdown. It contains session metadata
+(timestamps, total events, NATS address) and per-file metadata (event count, raw size, SHA-256
+checksum, event types, first/last timestamps).
+
+## Compression
+
+Completed files are compressed with zstd after rotation, in a background task that does not block
+event ingestion. The default compression level is 22 (ultra). Use `--compression-level 3` for
+faster compression (~5x ratio) or `--compression-level 0` to skip compression.
+
+## Example
+
+Archive all events from a NATS server, rotating files at 100 MB, with zstd compression:
+
+```
+$ cargo run -p archiver -- \
+    --nats-address 127.0.0.1:4222 \
+    --output-dir ./archive \
+    --base-name mainnet \
+    --max-file-size 104857600 \
+    --compression-level 22
+```
+
+Archive only P2P messages and mempool events:
+
+```
+$ ./target/release/archiver \
+    --nats-address 127.0.0.1:4222 \
+    --output-dir ./archive \
+    --messages --mempool
+```
+
+## Usage
+
+```
+Archive peer-observer events to disk
+
+Usage: archiver [OPTIONS] --output-dir <OUTPUT_DIR>
+
+Options:
+  -a, --nats-address <ADDRESS>
+          The NATS server address the extractor/tool should connect and subscribe to [default: 127.0.0.1:4222]
+  -u, --nats-username <USERNAME>
+          The NATS username the extractor/tool should try to authentificate to the NATS server with
+  -p, --nats-password <PASSWORD>
+          The NATS password the extractor/tool should try to authentificate to the NATS server with
+  -f, --nats-password-file <PASSWORD_FILE>
+          A path to a file containing a password the extractor/tool should try to authentificate to the NATS server with
+  -o, --output-dir <OUTPUT_DIR>
+          Output directory for archive files
+  -b, --base-name <BASE_NAME>
+          Base name for archive files (e.g., "mainnet" -> "mainnet.0.bin") [default: archive]
+      --max-file-size <MAX_FILE_SIZE>
+          Maximum file size in bytes before rotation (default: 1GB) [default: 1073741824]
+  -l, --log-level <LOG_LEVEL>
+          The log level the tool should run on [default: INFO]
+      --messages
+          If passed, archive P2P message events
+      --connections
+          If passed, archive P2P connection events
+      --addrman
+          If passed, archive addrman events
+      --mempool
+          If passed, archive mempool events
+      --validation
+          If passed, archive validation events
+      --rpc
+          If passed, archive RPC events
+      --p2p-extractor
+          If passed, archive p2p-extractor events
+      --log-extractor
+          If passed, archive log-extractor events
+      --compression-level <COMPRESSION_LEVEL>
+          Zstd compression level (0 = no compression, 3 = default, 22 = ultra) [default: 22]
+  -h, --help
+          Print help
+  -V, --version
+          Print version
+```
--- a/tools/archiver/README.md
+++ b/tools/archiver/README.md
@@ -30,9 +30,10 @@ checksum, event types, first/last timestamps).
 
 ## Compression
 
-Completed files are compressed with zstd after rotation, in a background task that does not block
-event ingestion. The default compression level is 22 (ultra). Use `--compression-level 3` for
-faster compression (~5x ratio) or `--compression-level 0` to skip compression.
+Archive files are compressed with zstd using streaming compression — the writer is wrapped in a
+`zstd::Encoder`, so files are written as `.bin.zst` directly. The default compression level is
+22 (ultra). Use `--compression-level 3` for faster compression (~5x ratio)
+or `--compression-level 0` to skip compression.
 
 ## Example
 
@@ -97,7 +98,7 @@ Options:
       --log-extractor
           If passed, archive log-extractor events
       --compression-level <COMPRESSION_LEVEL>
-          Zstd compression level (0 = no compression, 3 = default, 22 = ultra) [default: 22]
+          Zstd compression level (1-22, 0 = zstd default). Default: 22 (ultra) [default: 22]
   -h, --help
           Print help
   -V, --version
--- a/tools/archiver/README.md
+++ b/tools/archiver/README.md
@@ -25,15 +25,17 @@ to record the file, allowing future readers to check out that commit if needed.
 
 Each recording session produces a `<base-name>.manifest.toml` file alongside the archive files.
 The manifest is updated on every file rotation and on shutdown. It contains session metadata
-(timestamps, total events, NATS address) and per-file metadata (event count, raw size, SHA-256
-checksum, event types, first/last timestamps).
+(timestamps, total events, NATS address) and per-file metadata (event count, uncompressed size,
+SHA-256 checksums, event types, first/last timestamps). When compression is enabled, each file
+entry also includes the compressed file size and checksum.
 
 ## Compression
 
 Archive files are compressed with zstd using streaming compression — the writer is wrapped in a
 `zstd::Encoder`, so files are written as `.bin.zst` directly. The default compression level is
 22 (ultra). Use `--compression-level 3` for faster compression (~5x ratio)
-or `--compression-level 0` to skip compression.
+or `--compression-level 0` to skip compression. Rotation (`--max-file-size`) is checked against
+the compressed output stream. May overshoot slightly due to zstd internal buffering.
 
 ## Example
 
@@ -78,7 +80,7 @@ Options:
   -b, --base-name <BASE_NAME>
           Base name for archive files (e.g., "mainnet" -> "mainnet.0.bin") [default: archive]
       --max-file-size <MAX_FILE_SIZE>
-          Maximum file size in bytes before rotation (default: 1GB) [default: 1073741824]
+          Maximum compressed output size in bytes before rotation (default: 1GB) [default: 1073741824]
   -l, --log-level <LOG_LEVEL>
           The log level the tool should run on [default: INFO]
       --messages
--- a/tools/archiver/README.md
+++ b/tools/archiver/README.md
@@ -23,9 +23,8 @@ to record the file, allowing future readers to check out that commit if needed.
 
 ## Manifest
 
-Each recording session produces a `<base-name>.<timestamp>.manifest.toml` file alongside the
-archive files. The manifest is updated on every file rotation and on shutdown. It contains per-file
-metadata (version, NATS address, event count, uncompressed size, SHA-256 checksum, event types,
+Each archive file has a corresponding `<base-name>.<timestamp>.manifest.toml` with its metadata
+(version, NATS address, event count, uncompressed size, SHA-256 checksum, event types,
 first/last timestamps).
 
 ## Compression
@@ -86,8 +85,6 @@ Options:
           If passed, archive P2P message events
       --connections
           If passed, archive P2P connection events
-      --addrman
-          If passed, archive addrman events
       --mempool
           If passed, archive mempool events
       --validation
--- a/tools/archiver/README.md
+++ b/tools/archiver/README.md
@@ -23,11 +23,10 @@ to record the file, allowing future readers to check out that commit if needed.
 
 ## Manifest
 
-Each recording session produces a `<base-name>.manifest.toml` file alongside the archive files.
-The manifest is updated on every file rotation and on shutdown. It contains session metadata
-(timestamps, total events, NATS address) and per-file metadata (event count, uncompressed size,
-SHA-256 checksums, event types, first/last timestamps). When compression is enabled, each file
-entry also includes the compressed file size and checksum.
+Each recording session produces a `<base-name>.<timestamp>.manifest.toml` file alongside the
+archive files. The manifest is updated on every file rotation and on shutdown. It contains per-file
+metadata (version, NATS address, event count, uncompressed size, SHA-256 checksum, event types,
+first/last timestamps).
 
 ## Compression
 
@@ -78,7 +77,7 @@ Options:
   -o, --output-dir <OUTPUT_DIR>
           Output directory for archive files
   -b, --base-name <BASE_NAME>
-          Base name for archive files (e.g., "mainnet" -> "mainnet.0.bin") [default: archive]
+          Base name for archive files (e.g., "mainnet" -> "mainnet.<timestamp>.bin.zst") [default: archive]
       --max-file-size <MAX_FILE_SIZE>
           Maximum compressed output size in bytes before rotation (default: 1GB) [default: 1073741824]
   -l, --log-level <LOG_LEVEL>
--- a/tools/archiver/README.md
+++ b/tools/archiver/README.md
@@ -100,7 +100,7 @@ Options:
       --log-extractor
           If passed, archive log-extractor events
       --compression-level <COMPRESSION_LEVEL>
-          Zstd compression level (1-22, 0 = zstd default). Default: 22 (ultra) [default: 22]
+          Zstd compression level (0 = no compression, 1-22). Default: 22 (ultra) [default: 22]
   -h, --help
           Print help
   -V, --version
--- a/tools/archiver/build.rs
+++ b/tools/archiver/build.rs
@@ -12,14 +12,15 @@ fn main() {
     let b0 = u8::from_str_radix(&hash[0..2], 16).unwrap();
     let b1 = u8::from_str_radix(&hash[2..4], 16).unwrap();
     let b2 = u8::from_str_radix(&hash[4..6], 16).unwrap();
+    let b3 = u8::from_str_radix(&hash[6..8], 16).unwrap();
 
     let out_dir = std::env::var("OUT_DIR").unwrap();
     let path = std::path::Path::new(&out_dir).join("git_hash.rs");
     std::fs::write(
         path,
         format!(
-            "const GIT_HASH: [u8; 3] = [0x{:02x}, 0x{:02x}, 0x{:02x}];\n",
-            b0, b1, b2
+            "const GIT_HASH: [u8; 4] = [0x{:02x}, 0x{:02x}, 0x{:02x}, 0x{:02x}];\n",
+            b0, b1, b2, b3
         ),
     )
     .unwrap();
--- a/tools/archiver/build.rs
+++ b/tools/archiver/build.rs
@@ -1,0 +1,28 @@
+use std::process::Command;
+
+fn main() {
+    let output = Command::new("git")
+        .args(["rev-parse", "HEAD"])
+        .output()
+        .expect("failed to run git rev-parse HEAD");
+
+    let hash = String::from_utf8(output.stdout).unwrap();
+    let hash = hash.trim();
+
+    let b0 = u8::from_str_radix(&hash[0..2], 16).unwrap();
+    let b1 = u8::from_str_radix(&hash[2..4], 16).unwrap();
+    let b2 = u8::from_str_radix(&hash[4..6], 16).unwrap();
+
+    let out_dir = std::env::var("OUT_DIR").unwrap();
+    let path = std::path::Path::new(&out_dir).join("git_hash.rs");
+    std::fs::write(
+        path,
+        format!(
+            "const GIT_HASH: [u8; 3] = [0x{:02x}, 0x{:02x}, 0x{:02x}];\n",
+            b0, b1, b2
+        ),
+    )
+    .unwrap();
+
+    println!("cargo:rerun-if-changed=../../.git/HEAD");
+}
--- a/tools/archiver/src/lib.rs
+++ b/tools/archiver/src/lib.rs
@@ -8,6 +8,9 @@ use std::io::{BufWriter, Write};
 use std::path::{Path, PathBuf};
 use std::time::{SystemTime, UNIX_EPOCH};
 
+use time::macros::format_description;
+use time::OffsetDateTime;
+
 use sha2::{Digest, Sha256};
 use shared::serde::Serialize;
 
@@ -206,11 +209,13 @@ impl ArchiveFile {
         };
         // Retry on rare timestamp collisions (e.g. fast rotations in tests)
         let (path, file) = loop {
-            let ts = SystemTime::now()
-                .duration_since(UNIX_EPOCH)
-                .unwrap()
-                .as_millis() as u64;
-            let path = output_dir.join(format!("{}.{}.{}", base_name, ts, ext));
+            let timestamp = format_utc_timestamp(
+                SystemTime::now()
+                    .duration_since(UNIX_EPOCH)
+                    .unwrap()
+                    .as_millis() as u64,
+            );
+            let path = output_dir.join(format!("{}.{}.{}", base_name, timestamp, ext));
             match fs::OpenOptions::new()
                 .write(true)
                 .create_new(true)
@@ -224,7 +229,7 @@ impl ArchiveFile {
                     return Err(std::io::Error::new(
                         e.kind(),
                         format!("failed to create archive {}: {}", path.display(), e),
-                    ))
+                    ));
                 }
             }
         };
@@ -516,10 +521,19 @@ fn write_manifest(session: &SessionState, args: &Args) -> std::io::Result<()> {
     };
     let manifest_path = args.output_dir.join(format!(
         "{}.{}.manifest.toml",
-        args.base_name, session.started_at_ms
+        args.base_name,
+        format_utc_timestamp(session.started_at_ms)
     ));
     let toml_str = toml::to_string_pretty(&manifest).map_err(std::io::Error::other)?;
     fs::write(&manifest_path, &toml_str)?;
     log::info!("wrote manifest: {}", manifest_path.display());
     Ok(())
+}
+
+fn format_utc_timestamp(epoch_ms: u64) -> String {
+    let fmt = format_description!("[year][month][day]-[hour][minute][second]-[subsecond digits:3]");
+    OffsetDateTime::from_unix_timestamp_nanos(epoch_ms as i128 * 1_000_000)
+        .expect("timestamp out of range")
+        .format(&fmt)
+        .expect("timestamp formatting failed")
 }
--- a/tools/archiver/src/lib.rs
+++ b/tools/archiver/src/lib.rs
@@ -329,10 +329,6 @@ pub struct Args {
     #[arg(long)]
     pub connections: bool,
 
-    /// If passed, archive addrman events.
-    #[arg(long)]
-    pub addrman: bool,
-
     /// If passed, archive mempool events.
     #[arg(long)]
     pub mempool: bool,
@@ -363,7 +359,6 @@ impl Args {
     pub fn compress_all(&self) -> bool {
         !(self.messages
             || self.connections
-            || self.addrman
             || self.mempool
             || self.validation
             || self.rpc
@@ -379,7 +374,6 @@ pub async fn run(args: Args, mut shutdown_rx: watch::Receiver<bool>) -> Result<(
         log::info!("archiving all events:           {}", args.compress_all());
         log::info!("archiving P2P messages:         {}", args.messages);
         log::info!("archiving P2P connections:      {}", args.connections);
-        log::info!("archiving addrman events:       {}", args.addrman);
         log::info!("archiving mempool events:       {}", args.mempool);
         log::info!("archiving validation events:    {}", args.validation);
         log::info!("archiving rpc events:           {}", args.rpc);
@@ -489,7 +483,6 @@ fn should_archive(event: &Event, args: &Args) -> bool {
     match event_type_name(event) {
         Some("messages") => args.messages,
         Some("connections") => args.connections,
-        Some("addrman") => args.addrman,
         Some("mempool") => args.mempool,
         Some("validation") => args.validation,
         Some("rpc") => args.rpc,
@@ -504,7 +497,6 @@ fn event_type_name(event: &Event) -> Option<&'static str> {
         PeerObserverEvent::EbpfExtractor(e) => match e.ebpf_event.as_ref()? {
             ebpf::EbpfEvent::Message(_) => "messages",
             ebpf::EbpfEvent::Connection(_) => "connections",
-            ebpf::EbpfEvent::Addrman(_) => "addrman",
             ebpf::EbpfEvent::Mempool(_) => "mempool",
             ebpf::EbpfEvent::Validation(_) => "validation",
         },
--- a/tools/archiver/src/lib.rs
+++ b/tools/archiver/src/lib.rs
@@ -2,6 +2,15 @@ mod error;
 
 pub use error::RuntimeError;
 
+use std::collections::HashSet;
+use std::fs::{self, File};
+use std::io::{BufWriter, Write};
+use std::path::{Path, PathBuf};
+use std::time::{SystemTime, UNIX_EPOCH};
+
+use serde::Serialize;
+use sha2::{Digest, Sha256};
+
 use shared::clap;
 use shared::clap::Parser;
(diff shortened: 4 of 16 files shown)


Baldessar (commented):
Hi @octaviolucca, I reviewed and tested your code and overall it looks good! I'd like to highlight two points though:

- The replay tool doesn't seem to read files written by the archive tool when the --compression-level flag is set to 0 (No Compression). Is this the intended behavior?
- I noticed the replay tool is missing a README file. Every other tool and extractor has one explaining how to use it — I think it would be good to have one here as well.

Let me know your thoughts!

tools/archiver/src/lib.rs
```diff
@@ -0,0 +1,501 @@
...
+            created_at: SystemTime::now().duration_since(UNIX_EPOCH).unwrap().as_millis() as u64,
+            file_index: 0,
+            manifest_files: Vec::new(),
+        }
+    }
+}
+
+struct TrackingWriter {
+    inner: BufWriter<File>,
+    hasher: Sha256,
```
0xB10C: I was wondering if we could use the sha256 hashing from the bitcoin crate here and drop the dependency on `sha2`.

e.g.

```
use bitcoin_hashes::sha256;
use bitcoin_hashes::Hash;

let bytes = [0u8; 5];
let hash_of_bytes = sha256::Hash::hash(&bytes);
let hash_of_string = sha256::Hash::hash("some string".as_bytes());
```

tools/archiver/src/lib.rs
```diff
@@ -0,0 +1,501 @@
...
+#[derive(Serialize)]
+#[serde(crate = "shared::serde")]
+struct Session {
+    version: u8,
+    nats_address: String,
+    created_at: u64,
+    finished_at: u64,
+    total_events: u64,
+    total_files: u32,
+}
```
0xB10C: I'm not sure the concept of a Session is needed here. It might even make things a bit more complicated for us.

Currently, each session starts to write a file e.g. `archive.0.bin.zst`, and once that's full, it writes `archive.1.bin.zst`, and so on. Once the archiver tool is restarted (tool is updated, server is restarted, ...), we start to write a `archive.0.bin.zst` again. What happens to the initial file? Is it overwritten?

Also, when uploading files from multiple days into a folder with many archives from multiple sessions, this will cause conflicts.

My ideas would be to use the current UNIX epoch timestamp in seconds as number e.g. `archive.1773267404.bin.zst` or even a human readable date (`archive.2026-03-11-17-42-22.bin.zst`) and move the `version`, and `nats_address` into `FileEntry`. What do you think?

This would allow us to sort the files by date and we'd mostly avoid collisions, I think.

0xB10C:
I re-added `git` in shell.nix in e07289e (see https://github.com/peer-observer/peer-observer/pull/373#issuecomment-4024754017) - I think this commit got lost during a force-push. Feel free to `git pull` before making the next changes, otherwise it might get lost again :)

This also seems to have fixed the `rusttls` issue, which was compeltly unrelated?! Weird.

The `ubunut-check-and-fmt` and `nix-format` job fail as the code isn't formatted. You can run `cargo fmt` and then use e.g. `git absorb` (you'll need to install it first, but it's a handy tool!) and then rebase (it's going to tell you what to run) to fixup your commits. Alterantivly, stepping through the commits one-by-one and formatting them should also work.

I played around with this the other day, and it seems to work well. I think the one major thing here is possibly getting rid of the `Session` as metioned in https://github.com/peer-observer/peer-observer/pull/373#discussion_r2921297606.

tools/archiver/src/lib.rs
octaviolucca: Agreed on all points I have already removed the Session section from the manifest and moved version and nats_address into each FileEntry.

For the filenames, I'm using the epoch timestamp in seconds. When a file is full and rotates, should each new file get its own timestamp (archive.1773267404.bin.zst, archive.1773267512.bin.zst), or do we still need an index within the same run (archive.1773267404.0.bin.zst, archive.1773267404.1.bin.zst)?

The per-file timestamp is cleaner, but if two rotations happen within the same second (e.g. with a small --max-file-size), they'd produce the same filename and the second file would overwrite the first. Using milliseconds instead of seconds (archive.1773267404123.bin.zst) would make collisions extremely unlikely even with small --max-file-size, removing the need for an index entirely. What do you think? Would that work?

tools/archiver/README.md
```diff
@@ -0,0 +1,108 @@
...
+The manifest is updated on every file rotation and on shutdown. It contains session metadata
+(timestamps, total events, NATS address) and per-file metadata (event count, uncompressed size,
+SHA-256 checksums, event types, first/last timestamps). When compression is enabled, each file
+entry also includes the compressed file size and checksum.
+
+## Compression
+
+Archive files are compressed with zstd using streaming compression — the writer is wrapped in a
+`zstd::Encoder`, so files are written as `.bin.zst` directly. The default compression level is
+22 (ultra). Use `--compression-level 3` for faster compression (~5x ratio)
```
deadmanoz, line 34: Just a comment.

I assume that ultra (22) tries to minimise the size as much as possible, so thus maximises CPU? 

What's between 22 and 3?

tools/archiver/tests/integration.rs
```diff
@@ -0,0 +1,558 @@
...
+            "validation" => args.validation = true,
+            "rpc" => args.rpc = true,
+            "p2p_extractor" => args.p2p_extractor = true,
+            "log_extractor" => args.log_extractor = true,
+            _ => {} // show_all
+        }
+        archiver::run(args, shutdown_rx).await.unwrap();
+    });
+
+    sleep(Duration::from_secs(1)).await;
```
deadmanoz, line 228: Using sleep(1s) to wait for NATS readiness is a known source of flakiness under CI load. We've established a marker-based readiness barrier in the log extractor (#372) and websocket (#365) tests that eliminates this problem.

For the archiver, the cleanest approach is the same as the websocket tool: accept an optional `oneshot::Sender<()>` in `run()` and fire it after `nc.subscribe("*").await` succeeds. Tests await the `oneshot` instead of sleeping. Same applies to the standalone tests (`test_file_rotation_with_compression`, `test_replayer_roundtrip`, etc.) which using the same sleep pattern.

Not blocking (I am happy to resolve this after the fact), but worth adopting for consistency and reliability.

tools/archiver/src/lib.rs
```diff
@@ -0,0 +1,501 @@
+mod error;
```
deadmanoz, line 3: Missing `#![cfg_attr(feature = "strict", deny(warnings))]`. All other tools have this. Without it clippy warnings won't fail CI for this crate.

tools/archiver/src/lib.rs
```diff
@@ -0,0 +1,501 @@
...
+use shared::log;
+use shared::nats_util;
+use shared::prost::Message;
+use shared::protobuf::ebpf_extractor::ebpf;
+use shared::protobuf::event::event::PeerObserverEvent;
+use shared::protobuf::event::Event;
+use shared::tokio::sync::watch;
+
+const MAGIC: [u8; 2] = *b"PA";
+const VERSION: u8 = 1;
```
deadmanoz, line 32: This is currently unused on the replayer side?

Edit: see comment

deadmanoz:
I have some comments and questions around the versioning approach.

There are effectively 2 layers in use:
- VERSION: is effectively for the file format (container) 
- git hash: the `Event` protobuf messages within the container.

For the second layer, protobuf's wire format handles schema evolution at the field level: old readers skip unknown fields, new readers tolerate missing ones with defaults. 

The main gap is the `oneof` variants which are used throughout the schema (`PeerObserverEvent`, `EbpfEvent`, `MempoolEvent`, etc.) - an old replayer seeing a new variant at any level gets `None`, not the raw bytes. But the git hash doesn't solve that either; you'd need an updated replayer regardless.

If a breaking schema change does happen, the container/format version byte is presumably the standard mechanism for gating decoding (maybe we can enforce this in CI one day). So do we actually need the git hash, or is VERSION sufficient for both layers (given how protobuf handles things)?

deadmanoz:
So I did a bit of testing of this PR and #377 together to validate cross-schema archive compatibility (AI-assisted). Specifically can a replayer built from one schema version correctly read archives created with a different one). Both directions failed with `decode error: buffer underflow` due to a bug in `replayer::read_archive()`.

The cursor advancement logic uses `encoded_len()` on the decoded struct to compute how far to advance through the length-delimited stream:

```rust
let event = Event::decode_length_delimited(&data[cursor..])?;
let size = event.encoded_len();
let varint_len = prost::length_delimiter_len(size);
cursor += varint_len + size;
```

`encoded_len()` returns the size of the *in-memory* decoded message, not the actual bytes that were on the wire. When the encoding and decoding schemas match exactly (as in all the existing roundtrip tests), these are equal. But when they differ, which is the whole point of archives being long-lived, they diverge:

- **Old archive → New replayer**: the decoded struct includes new `required` fields at their default values, so `encoded_len()` is *larger* than the wire size. The cursor overshoots into the middle of the next event.
- **New archive → Old replayer**: unknown wire fields are dropped during decode, so `encoded_len()` is *smaller* than the wire size. The cursor undershoots.

The fix is to read the varint length prefix directly from the wire data instead of recomputing it from the decoded struct:

```rust
let mut wire = &data[cursor..];
let payload_len = prost::decode_length_delimiter(&mut wire)?;
let varint_len = (data.len() - cursor) - wire.len();
let event = Event::decode(&data[cursor + varint_len..cursor + varint_len + payload_len])?;
cursor += varint_len + payload_len;
```

I verified this fix against a 4-way compatibility matrix (old/new archive × old/new replayer) - all scenarios pass, and existing roundtrip tests are unaffected.

octaviolucca:
> The main gap is the `oneof` variants which are used throughout the schema (`PeerObserverEvent`, `EbpfEvent`, `MempoolEvent`, etc.) - an old replayer seeing a new variant at any level gets `None`, not the raw bytes. But the git hash doesn't solve that either; you'd need an updated replayer regardless.
> 
> If a breaking schema change does happen, the container/format version byte is presumably the standard mechanism for gating decoding (maybe we can enforce this in CI one day). So do we actually need the git hash, or is VERSION sufficient for both layers (given how protobuf handles things)?

Yes, `VERSION` tells us how to decode the container, but not which exact protobuf schema/build produced the payloads. Without the git hash, we would only know the file came from a writer compatible with that container version, not the specific commit that produced it. So I see the hash as provenance/debug metadata, while `VERSION` remains the actual compatibility boundary.

tools/archiver/README.md
octaviolucca: Yes, 22 maximizes compression ratio at the cost of CPU. Levels 1-22 trade off ratio vs speed, 3 is zstd's default. I chose 22 since these are archive files meant for storage, but it is configurable if CPU becomes a concern.

tools/archiver/tests/integration.rs
octaviolucca: Good point, I will adopt the oneshot readiness barrier in a follow-up. Thanks for the pointer to #372 and #365.

octaviolucca:
> `encoded_len()` returns the size of the _in-memory_ decoded message, not the actual bytes that were on the wire. When the encoding and decoding schemas match exactly (as in all the existing roundtrip tests), these are equal. But when they differ, which is the whole point of archives being long-lived, they diverge:
> 
You are right,`encoded_len()` returns the size of the decoded in-memory struct, not the actual wire bytes. Will fix the cursor logic to read the varint length prefix directly from the wire data instead.

0xB10C:
I'm marking this as draft for now, feel free to mark ready for review once it is! No hurry!

0xB10C:
Feel free to mark this as ready-for-review when you think I should have another look! I think we're close to done here. :slightly_smiling_face:

tools/archiver/src/lib.rs
```diff
@@ -0,0 +1,524 @@
...
+    pub log_extractor: bool,
+
+    /// Zstd compression level (0 = no compression, 1-22). Default: 22 (ultra).
+    #[arg(long, default_value_t = 22)]
+    pub compression_level: u32,
+}
+
+impl Args {
+    /// Returns true if all event types should be archived (no filters specified).
+    pub fn show_all(&self) -> bool {
```
0xB10C: could also be (but not important)

```suggestion
    pub fn compress_all(&self) -> bool {
```

tools/archiver/tests/integration.rs
```diff
@@ -0,0 +1,647 @@
...
+    }
+
+    sleep(Duration::from_millis(500)).await;
+    shutdown_tx.send(true).unwrap();
+    archiver_handle.await.unwrap();
+
+    let archive_path = find_files(&tmp_dir, ".bin.zst")
+        .into_iter()
+        .next()
+        .expect("expected a .0.bin.zst file");
```
0xB10C:
```suggestion
        .expect("expected a .<timestamp>.bin.zst file");
```

tools/replayer/Cargo.toml
```diff
@@ -0,0 +1,8 @@
+[package]
+name = "replayer"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+shared = { path = "../../shared" }
+zstd = "0.13"
```
0xB10C: We are using this in the archiver and replayer, and it probably makes sense to update these at the same time. So this would be a good candidate to have in `shared` and re-exported there.

0xB10C:
I've started running this on the `len` demo node with 

```$ cargo run -p archiver --release -- --output-dir ./archive --base-name len-mainnet
2026-04-07T09:07:13.129Z INFO  [archiver] archiving all events: true
2026-04-07T09:07:13.133Z INFO  [archiver] Connected to NATS-server at 127.0.0.1:4222
2026-04-07T09:07:13.447Z INFO  [archiver] Created archive file: ./archive/len-mainnet.1775552833133.bin.zst
```

It's currently sitting at around 5% CPU usage with the ultra compression level and about 1 GB in memory usage (which I wasn't expecting!)

edit: It's been running for 20 minutes and the `len-mainnet.1775552833133.bin.zst` file has 20MB in size. So 3.6 GB in size with everything turned on? We'll see!

edit2: after 11h the file is now 616MB in size. So growth slowed down, which is great! It seems to be using somewhere between 10%-30% CPU at the moment, but memory stayed constant.

0xB10C:
I've also had a look at the most recent coverage report created by the CI: https://github.com/peer-observer/peer-observer/actions/runs/23907662835/artifacts/6246969613

It shows >97% code coverage in `/archiver/src/lib.rs` which is great! I'll need to go through the test once more to review. Sofar only skimmed them. But they seem to cover most of the archiver code.

0xB10C:
I've also been spending time monitoring BIP-54 slow blocks on signet (https://bnoc.xyz/t/slowish-blocks-on-signet/100/19) - there is going to be another event like this tomorrow (https://delvingbitcoin.org/t/consensus-cleanup-demo-of-slow-blocks-on-signet/2367). I was thinking of running the archivers for the duration of the events to have
1. exact messages during the slow blocks
2. can test the archiver and replayer on this

0xB10C:
I think you need to also check in the updated `Cargo.lock` in 4b149e8 - running `cargo build` produces a different file for me locally.

0xB10C:
Leaving a comment for myself to investigate a some later point:

For some reason, during the build of peer-observer in https://github.com/0xB10C/nix/blob/master/pkgs/peer-observer/default.nix it fails with: ` byte index 2 is out of bounds of `. Either, git can't be found (I've added `pkgs.git`), or we are not in a git repository during the build.

```
       Last 25 log lines:
       >    Compiling icu_properties v1.5.1
       >    Compiling prost-build v0.14.1
       >    Compiling bitreq v0.3.1 (https://github.com/0xb10c/corepc?rev=b777fa7160c7b97e1e845e771953a158233660cd#b777fa71)
       >    Compiling http v1.1.0
       >    Compiling futures-executor v0.3.31
       >    Compiling base58ck v0.1.0
       >    Compiling secp256k1 v0.29.0
       >    Compiling jsonrpc v0.19.0 (https://github.com/0xb10c/corepc?rev=b777fa7160c7b97e1e845e771953a158233660cd#b777fa71)
       >    Compiling shared v0.1.0 (/build/source/shared)
       >    Compiling futures v0.3.31
       >    Compiling cargo_metadata v0.19.2
       >    Compiling serde_spanned v0.6.9
       >    Compiling toml_datetime v0.6.11
       >    Compiling tungstenite v0.29.0
       >    Compiling toml_edit v0.22.27
       > error: failed to run custom build command for `archiver v0.1.0 (/build/source/tools/archiver)`
       >
       > Caused by:
       >   process didn't exit successfully: `/build/source/target/release/build/archiver-80a7a012c2729a4d/build-script-build` (exit status: 101)
       >   --- stderr
       >
       >   thread 'main' (5092) panicked at tools/archiver/build.rs:12:38:
       >   byte index 2 is out of bounds of ``
       >   note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
  ```

**Comment by octaviolucca** [comment]
> I think you need to also check in the updated `Cargo.lock` in [4b149e8](https://github.com/peer-observer/peer-observer/commit/4b149e8bf83e2a0ae4024982ac64b6296074c07c) - running `cargo build` produces a different file for me locally.

Fixed Cargo.lock is now included in the commit.

**Comment by octaviolucca** [comment]
> I've also been spending time monitoring BIP-54 slow blocks on signet (https://bnoc.xyz/t/slowish-blocks-on-signet/100/19) - there is going to be another event like this tomorrow (https://delvingbitcoin.org/t/consensus-cleanup-demo-of-slow-blocks-on-signet/2367). I was thinking of running the archivers for the duration of the events to have
> 
> 1. exact messages during the slow blocks
> 2. can test the archiver and replayer on this

Nice, I am also running the archiver during the event so we can cross-check the data.

**Comment by 0xB10C** [comment]
I've posted an idea about comparing our format (and approach with the extra manifest file) to the Wireshark pcap file format in https://github.com/peer-observer/peer-observer/issues/121#issuecomment-4213949531. I think we should merge this PR once it's ready code wise (still planning to do another review a round), and then, if we need, iterate on the format a bit. Essentially treating all recordings as experimental for now.

**Comment by 0xB10C** [comment]
Here are a few archives and manifests (put in a tar.gz for transport with e.g. `tar -czvf inq-signet.1775653665937.tar.gz inq-signet.1775653665937*`) from the slow blocks on signet:

- [core-signet.1775684395280.tar.gz](https://github.com/user-attachments/files/26602191/core-signet.1775684395280.tar.gz)
- [core-signet.1775653608363.tar.gz](https://github.com/user-attachments/files/26602192/core-signet.1775653608363.tar.gz)
- [inq-signet.1775653665937.tar.gz](https://github.com/user-attachments/files/26602292/inq-signet.1775653665937.tar.gz)
- [inq-signet.1775684391104.tar.gz](https://github.com/user-attachments/files/26602293/inq-signet.1775684391104.tar.gz)

**Comment by 0xB10C** [comment]
I've been working with a different dataset that has files with a Unix-epoch timestamp in the filename recently. I just realized that this is a big mental hurdle to reason about the files, especially if you have a lot of them. It always requires humans to convert from 1775684391104 to Wednesday, April 8, 2026 at 9:39:51.104 PM UTC. I think we should use `20260408-093951-104` (YYYYMMDD-HHMMSS-MS) in UTC as the timestamp. This still allows for sorting the files by date and is much more human friendly!

**Comment by 0xB10C** [comment]
I had this running for a while on one of the demo servers.

Log output:

```
2026-04-07T09:07:13.129Z INFO  [archiver] archiving all events: true
2026-04-07T09:07:13.133Z INFO  [archiver] Connected to NATS-server at 127.0.0.1:4222
2026-04-07T09:07:13.447Z INFO  [archiver] Created archive file: ./archive/len-mainnet.1775552833133.bin.zst
2026-04-08T01:46:04.422Z INFO  [archiver] wrote manifest: ./archive/len-mainnet.1775552833133.manifest.toml
2026-04-08T20:27:02.298Z INFO  [archiver] wrote manifest: ./archive/len-mainnet.1775552833133.manifest.toml
2026-04-09T16:26:04.070Z INFO  [archiver] wrote manifest: ./archive/len-mainnet.1775552833133.manifest.toml
2026-04-10T13:21:18.266Z INFO  [archiver] wrote manifest: ./archive/len-mainnet.1775552833133.manifest.toml
2026-04-11T09:37:30.569Z INFO  [archiver] wrote manifest: ./archive/len-mainnet.1775552833133.manifest.toml
2026-04-12T05:32:04.253Z INFO  [archiver] wrote manifest: ./archive/len-mainnet.1775552833133.manifest.toml
2026-04-12T23:19:11.376Z INFO  [archiver] wrote manifest: ./archive/len-mainnet.1775552833133.manifest.toml
2026-04-13T16:28:18.025Z INFO  [archiver] wrote manifest: ./archive/len-mainnet.1775552833133.manifest.toml
2026-04-14T09:57:52.116Z INFO  [archiver] wrote manifest: ./archive/len-mainnet.1775552833133.manifest.toml
2026-04-15T04:37:32.573Z INFO  [archiver] wrote manifest: ./archive/len-mainnet.1775552833133.manifest.toml
2026-04-16T01:17:34.652Z INFO  [archiver] wrote manifest: ./archive/len-mainnet.1775552833133.manifest.toml
2026-04-16T19:44:05.388Z INFO  [archiver] wrote manifest: ./archive/len-mainnet.1775552833133.manifest.toml
```

Is this expected behavior that the manifest file is appended? I was under the impression we'd have a separate manifest file per file, but maybe I'm wrong here.:

```$head len-mainnet.1775552833133.manifest.toml -n 40
[[files]]
name = "len-mainnet.1775552833133.bin.zst"
version = 1
nats_address = "127.0.0.1:4222"
size_bytes = 12817088721
events = 36092523
first_timestamp = 1775552833451
last_timestamp = 1775612758848
event_types = [
    "connections",
    "log_extractor",
    "mempool",
    "messages",
    "p2p_extractor",
    "rpc",
    "validation",
]
checksum = "8d96f3d3c076ec18ec6e07ca7d6703bd77b3e2d004830cfb6a7c1a4f6e694a41"
compressed_size_bytes = 1073851820
compressed_checksum = "b6a36de54f41be7f46cfeb8861e1fe8c97f23c715a4ed509be73c846111e44b8"

[[files]]
name = "len-mainnet.1775612764422.bin.zst"
version = 1
nats_address = "127.0.0.1:4222"
size_bytes = 14945878162
events = 43196011
first_timestamp = 1775612758941
last_timestamp = 1775680022141
event_types = [
    "connections",
    "log_extractor",
    "mempool",
    "messages",
    "p2p_extractor",
    "rpc",
    "validation",
]
checksum = "edf65787b10eaa79b92c8ad7cedfcd11e6714211136733a824fadb4e8a6f3e0a"
compressed_size_bytes = 1073747533
```

Compression rate seems to be good with 12817088721 / 1073851820 = ~12x and 14945878162 / 1073747533 = ~14x!

```
$ ll -h

1.1G Apr  8 01:46 len-mainnet.1775552833133.bin.zst
6.4K Apr 16 19:44 len-mainnet.1775552833133.manifest.toml
1.1G Apr  8 20:27 len-mainnet.1775612764422.bin.zst
1.1G Apr  9 16:26 len-mainnet.1775680022298.bin.zst
1.1G Apr 10 13:21 len-mainnet.1775751964071.bin.zst
1.1G Apr 11 09:37 len-mainnet.1775827278266.bin.zst
1.1G Apr 12 05:32 len-mainnet.1775900250570.bin.zst
1.1G Apr 12 23:19 len-mainnet.1775971924253.bin.zst
1.1G Apr 13 16:28 len-mainnet.1776035951376.bin.zst
1.1G Apr 14 09:57 len-mainnet.1776097698025.bin.zst
1.1G Apr 15 04:37 len-mainnet.1776160672116.bin.zst
1.1G Apr 16 01:17 len-mainnet.1776227852573.bin.zst
1.1G Apr 16 19:44 len-mainnet.1776302254652.bin.zst
878M Apr 17 10:56 len-mainnet.1776368645388.bin.zst
```

0xB10C:
I'm wondering if we really need a manifest file at all.

- If we put all manifests in the same file, we need to do manual work for splitting the individual manifests for sharing.
- We probably end up `tar`-ing the bin.zst up with the `manifest.toml` together again as done in https://github.com/peer-observer/peer-observer/pull/373#issuecomment-4214075970 - why have it separate then?
- Looking at a manifest file, we can re-generate most of it from the the archive. What's the point of creating the file in the archiver then?  

Maybe an alternative would be to have a tool like `capinfos` (https://github.com/peer-observer/peer-observer/issues/121#issuecomment-4213949531), maybe called `archiveinfos` that outputs a similar description of the file as the current manifest contains.

tools/archiver/src/lib.rs
```diff
@@ -0,0 +1,525 @@
...
+    pub log_extractor: bool,
+
+    /// Zstd compression level (0 = no compression, 1-22). Default: 22 (ultra).
+    #[arg(long, default_value_t = 22)]
+    pub compression_level: u32,
+}
+
+impl Args {
+    /// Returns true if all event types should be archived (no filters specified).
+    pub fn compress_all(&self) -> bool {
```
RazorBest: Nit: I'd name this `archive_all`, to keep the distinction between compression and archiving.

tools/archiver/tests/integration.rs
```diff
@@ -0,0 +1,648 @@
...
+        log_level: log::Level::Info,
+        messages: false,
+        connections: false,
+        addrman: false,
+        mempool: false,
+        validation: false,
+        rpc: false,
+        p2p_extractor: false,
+        log_extractor: false,
+        compression_level: 22,
```
RazorBest: The compression level affects the memory usage of the tests. Even if we write one byte to the `ArchiveWriter`, the process can consume a few hundred MBs. I did the following experiment:

I modified `src/lib.rs::ArchiveFile::new` to contain `writer.write_all(&header_bytes[..1])?;`, and panicked immediately after that. I ran `cargo test --no-run --features nats_integration_tests && NATS_SERVER_BINARY=nats-server /usr/bin/time -v cargo test --features nats_integration_tests test_filter_all` with different compression levels, and looked at the `Maximum resident set size (kbytes)`.

Results:
```
compression_level    max_memory_kb
22                         731392
21                         338048
20                         174720
19                         92672
18                         67852
17                         67804
16                         66976
1                          67416
```

RazorBest: This is not necessarily an issue. But on my machine, the tests were sometimes failing because there was not enough RAM. If you run all the tests, the peak memory usage easily goes over 1GB.

When I ran all the tests with compression_level 18, it used at most 128MB (which is better, but I still think it's a lot just for testing).

tools/archiver/tests/integration.rs
```diff
@@ -0,0 +1,648 @@
...
+                    log_extractor::UnknownLogMessage {
+                        raw_message: "test log".to_string(),
+                    },
+                )),
+            }))
+            .unwrap(),
+            "log_extractor",
+        ),
+    ]
+}
```
RazorBest, line 199: I was thinking that this function can become outdated whenever we add a new extractor. I don't think there's an issue if `make_all_event_types` doesn't actually returns all types of events, but I found a way to signal this kind of issue.

You can insert this code in here:
```rust
    let ret = vec![ /* your actual values in here */ ]
    match ret[0].0.peer_observer_event.as_ref().unwrap() {
        PeerObserverEvent::EbpfExtractor(_) => (),
        PeerObserverEvent::RpcExtractor(_) => (),
        PeerObserverEvent::P2pExtractor(_) => (),
        PeerObserverEvent::LogExtractor(_) => (),
        // If this match expression generates a compilation error, it means that a new event type was added.
        // This means that the vector above doesn't contain all the events.
        // Update the vector before fixing this match expression.
    }

    return ret;
```

Which is dead code. The only purpose of the match expression is to generate a compile error whenever a new event is added.

RazorBest (approved):
Successfully tested with:
- `NATS_SERVER_BINARY=nats-server cargo test --release --features nats_integration_tests`
- `NATS_SERVER_BINARY=nats-server cargo test --features nats_integration_tests`

I'm also sharing my thoughts about the manifests, hoping to get closer to the "why?".

I'm thinking if it's better to have one manifest per compressed file. But it depends whether we agree with the following:

From b10c:
>  we need to do manual work for splitting the individual manifests for sharing

Is there really going to be a case where: the recorder is started, we keep it running, at some point we decide we want to share the archives, and we would want to share only a fraction of them? If we want to share a fraction, how are going to decide which ones to choose?

Also, with regards to the need of generating the manifest files at all:

> Looking at a manifest file, we can re-generate most of it from the the archive. What's the point of creating the file in the archiver then?

I think if the only purpose for the archives is to share them to be fed to the replayer, then the manifests are not needed at all. But a different use case would be: you have a directory with a lot of archives, and you don't remember anything about them, or someone else gave them to you; the easiest way to get some information about them is through the human readable summary of the archives. This validates the usefulness of the manifests, but as b10c mentioned, it might still not be justified to create the manifests at the same time with the archives. You could bring additional arguments to why we should do X instead of Y, but at some point this becomes "fairy engineering", if we don't have use cases that have already proven to be needed.

RazorBest:
I also started thinking about compression. And why we're not using just a simple encoding convention that minimizes space. But my guess is that compression tries to do that dynamically for us. So, if we have 100 events of the same type, and all 100 of them have a common substring because of the protobuf encoding, most probably, the compression algorithm will detect that. And I'm also guessing that our generated data doesn't have a lot of "depth". One way to quantify that would be: what is the path of fields that need to be accessed to get from a top level event struct to a primitive type? I think length of these paths and their amount could describe the depth of these structures.

So I did an experiment by generating 10000 random events with `proptest` and `proptest-derive` (very fun experience), and tried archiving with different compression levels. The results were underwhelming:

| compression level | archive size |
| --- | --- |
| 22 | 3.8M |
| 21 | 3.6M |
| 20 | 3.5M |
| 19 | 3.7M |
| 18 | 3.6M |
| 17 | 3.6M |
| 10 | 3.8M |
| 5 | 3.5M |
| 1 | 3.9M |
| 0 | 4.1M |

The variations you see are also because every time, new random events were generated. But, there's no measurable difference between compression 1 and compression 22.

Maybe things are different when files are GBs of size. However, `proptest` has a specific way in which it derives test cases which make it pretty slow. Generating 10000 samples on a slow machine was taking about 5-6 minutes.

Also, this test was pretty extreme. It didn't take into consideration that some of the fields have way less entropy or they have a constant value (like the name of the node) throughout the lifetime of the program.

octaviolucca:
> Is this expected behavior that the manifest file is appended? I was under the impression we'd have a separate manifest file per file, but maybe I'm wrong here.:

Right, the current implementation creates one manifest file per session with the data from all files of that session. I will fix this in the next commit so that each file gets its own manifest.

octaviolucca:
> I'm wondering if we really need a manifest file at all.
> Maybe an alternative would be to have a tool like `capinfos` ([#121 (comment)](https://github.com/peer-observer/peer-observer/issues/121#issuecomment-4213949531)), maybe called `archiveinfos` that outputs a similar description of the file as the current manifest contains.

I agree with your point here. After thinking about it, generating the manifest as part of the archiver probably does not add much value, especially if the same information can be derived from the archive later. For now, though, I don't think keeping the manifest causes any real issues, so I would treat this more as a future improvement rather than something that needs to be changed immediately.

0xB10C:
Yes, agree. IMO priority here is to get this in a mergable state and merge it. It's a big change and has been open for a while. Once merged, we can experiment with it and iterate. For example, removing the mainifest once we have a `capinfos`-like tool.

tools/archiver/src/lib.rs
```diff
@@ -0,0 +1,503 @@
...
+                if let Some(msg) = maybe_msg {
+                    let event = match Event::decode(msg.payload.as_ref()) {
+                        Ok(event) => event,
+                        Err(e) => {
+                            log::warn!("failed to decode event: {}, skipping", e);
+                            continue;
+                        }
+                    };
+                    if should_archive(&event, &args){
+                        current_file.write_event(&event)?;
```
deadmanoz: `write_event` failing here propagates via `?` and returns from `run()` directly - it skips the post-loop `close_file` call, so the `zstd::Encoder` is dropped without `finish()`, leaving the `.bin.zst` truncated and undecodable and no manifest written. `Encoder::finish()` is [documented](https://docs.rs/zstd/0.13/zstd/stream/write/struct.Encoder.html#method.finish) as required. Probably worth turning the `?` here (and on `rotate` below) into `log::error!(...); break;` so the existing `close_file` at the end of `run()` runs on the error path too?

> Required: Finishes the stream. You need to finish the stream when you're done writing, either with this method or with try_finish(self).

tools/replayer/src/lib.rs
```diff
@@ -0,0 +1,65 @@
...
+
+    let data = &buf[HEADER_SIZE..];
+    let mut cursor = 0;
+    let mut events = Vec::new();
+
+    while cursor < data.len() {
+        let event = Event::decode_length_delimited(&data[cursor..]).map_err(|e| {
+            std::io::Error::other(format!("decode error at byte {}: {}", cursor, e))
+        })?;
+        let size = event.encoded_len();
```
0xB10C: I think this still has the `encoded_len()` problem @deadmanoz mentioned in https://github.com/peer-observer/peer-observer/pull/373#issuecomment-4089895180, no?

tools/replayer/src/main.rs
```diff
@@ -0,0 +1,55 @@
...
+                            println!("[{n}] ts={ts} rpc: {}", r.rpc_event.as_ref().unwrap())
+                        }
+                        Some(PeerObserverEvent::P2pExtractor(p)) => {
+                            println!("[{n}] ts={ts} p2p: {}", p.p2p_event.as_ref().unwrap())
+                        }
+                        Some(PeerObserverEvent::LogExtractor(l)) => {
+                            println!("[{n}] ts={ts} log: {}", l.log_event.as_ref().unwrap())
+                        }
+                        None => println!("[{n}] ts={ts} <unknown>"),
+                    }
```
0xB10C, line 48: As a followup: Maybe we should actually log the error here instead of using `.unwrap()`.

tools/replayer/src/main.rs
```diff
@@ -0,0 +1,55 @@
...
+use shared::protobuf::event::event::PeerObserverEvent;
+
+fn main() {
+    let files: Vec<String> = env::args().skip(1).collect();
+    if files.is_empty() || files.iter().any(|f| f == "--help" || f == "-h") {
+        println!("Read and display peer-observer archive files");
+        println!();
+        println!("Usage: replayer <file.bin[.zst]> [file2.bin[.zst] ...]");
+        return;
+    }
```
0xB10C, line 13: As a followup: We should use `clap` for arguments and `log` for logging here.

tools/archiver/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+
+[dependencies]
+shared = { path = "../../shared" }
+time = { version = "0.3", features = ["formatting", "macros"] }
+toml = "0.8"
+sha2 = "0.10"
+
+[dev-dependencies]
+sha2 = "0.10"
+toml = "0.8"
```
0xB10C, line 14:
```suggestion
[dev-dependencies]
```

dev-dependencies for things only needed in tests, examples, etc, but we already have them in the normal dependecies

tools/archiver/README.md
```diff
@@ -0,0 +1,104 @@
...
+(version, NATS address, event count, uncompressed size, SHA-256 checksum, event types,
+first/last timestamps).
+
+## Compression
+
+Archive files are compressed with zstd using streaming compression — the writer is wrapped in a
+`zstd::Encoder`, so files are written as `.bin.zst` directly. The default compression level is
+22 (ultra). Use `--compression-level 3` for faster compression (~5x ratio)
+or `--compression-level 0` to skip compression. Rotation (`--max-file-size`) is checked against
+the compressed output stream. May overshoot slightly due to zstd internal buffering.
```
0xB10C, line 36: in a followup: could also mention that this has effects on memory usage, as @RazorBest mentioned!

tools/archiver/src/lib.rs
```diff
@@ -0,0 +1,515 @@
...
+use std::path::{Path, PathBuf};
+use std::time::{SystemTime, UNIX_EPOCH};
+
+use time::macros::format_description;
+use time::OffsetDateTime;
+
+use sha2::{Digest, Sha256};
+use shared::serde::Serialize;
+
+use shared::clap;
```
0xB10C, line 19: nit

```suggestion
use shared::serde::Serialize;
use shared::clap;
```

0xB10C:
Sorry only getting back to you now @RazorBest (https://github.com/peer-observer/peer-observer/pull/373#pullrequestreview-4177282605):

>>  we need to do manual work for splitting the individual manifests for sharing. 
>
> Is there really going to be a case where: the recorder is started, we keep it running, at some point we decide we want to share the archives, and we would want to share only a fraction of them? If we want to share a fraction, how are going to decide which ones to choose?

My use case would be:
- I run a archiver on the production hosts
- It rotates after e.g. 1GB
- Each night, I upload the completed files to archival storage
- One week/month/year later, someone needs to look something up on day X (e.g. we saw a bug / attack on that day)
- Download the two 1GB files from day X and e.g. replay them
- The oder files are not interesting at this point and I don't want to spend time finding the manifest for it

tools/archiver/tests/integration.rs
```diff
@@ -0,0 +1,653 @@
...
+            "rpc" => args.rpc = true,
+            "p2p_extractor" => args.p2p_extractor = true,
+            "log_extractor" => args.log_extractor = true,
+            _ => {} // archive_all
+        }
+        archiver::run(args, shutdown_rx).await.unwrap();
+    });
+
+    sleep(Duration::from_secs(1)).await;
+
```
0xB10C, line 229: for a followup: This will cause intermitting failures: See https://github.com/peer-observer/peer-observer/pull/373/changes#r2957410341

tools/archiver/tests/integration.rs
```diff
@@ -0,0 +1,653 @@
...
+    let dir = tmp_dir.clone();
+    let archiver_handle = tokio::spawn(async move {
+        let mut args = make_test_args(nats_server.port, &dir);
+        args.max_file_size = 1; // force rotation on every event
+        args.compression_level = 1;
+        archiver::run(args, shutdown_rx).await.unwrap();
+    });
+
+    sleep(Duration::from_secs(1)).await;
```
0xB10C, line 329: followup: https://github.com/peer-observer/peer-observer/pull/373/changes#r2957410341

0xB10C (commented):
I think this is fine to merge as experimental at this point, but we should follow up with a few things to make it ready for actual production usage.

I'm still favoring removing the mainfest writing, especially session based, and having a `capinfos` like tool to show and generate them.

Some other follow up ideas below. Please only mark them as addressed once they are actually done. Feel free to pick them up, if you want. If not, feel free to let others know that there are free to do so!

Sorry for keeping this open for over two months, but this turned out to be a rather involved change.

Normally, I'd ask you to squash the commits into a couple (one for archiver, one for replayer, one for tests), but I think that will create unnecessary friction now.

As a next step, I'll look into adding support for building and running this on the Nix side. Deploying this to one of my hosts eventually.

0xB10C:
Thank you @octaviolucca for working on this!

Thank you @deadmanoz @RazorBest @Baldessar for your comments, review and feedback!

tools/replayer/Cargo.toml
```diff
@@ -0,0 +1,7 @@
+[package]
+name = "replayer"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+shared = { path = "../../shared" }
```
0xB10C, line 7:
```suggestion
shared = { path = "../../shared" }

[features]
# Treat warnings as a build error.
strict = []
```

0xB10C: and `#![cfg_attr(feature = "strict", deny(warnings))]` at the top of the file

tools/archiver/build.rs
```diff
@@ -0,0 +1,29 @@
...
+        .output()
+        .expect("failed to run git rev-parse HEAD");
+
+    let hash = String::from_utf8(output.stdout).unwrap();
+    let hash = hash.trim();
+
+    let b0 = u8::from_str_radix(&hash[0..2], 16).unwrap();
+    let b1 = u8::from_str_radix(&hash[2..4], 16).unwrap();
+    let b2 = u8::from_str_radix(&hash[4..6], 16).unwrap();
+    let b3 = u8::from_str_radix(&hash[6..8], 16).unwrap();
```
0xB10C, line 15: The build.rs failure error of this (when no git is installed) is slightly confusing:

       >    Compiling tracing-subscriber v0.3.20
       > error: failed to run custom build command for `archiver v0.1.0 (/build/source/tools/archiver)`
       >
       > Caused by:
       >   process didn't exit successfully: `/build/source/target/release/build/archiver-80a7a012c2729a4d/build-script-build` (exit status: 101)
       >   --- stderr
       >
       >   thread 'main' (3769) panicked at tools/archiver/build.rs:12:38:
       >   byte index 2 is out of bounds of ``
       >   note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

0xB10C: I think the actual problem is that git is available, but in the nix build env it's not a repository, just a plain folder?

0xB10C: fixing in https://github.com/peer-observer/peer-observer/pull/429

Issue #285, opened 2025-10-29
log-extractor followups


0xB10C:
In #272, @m4ycon implemented a basic log extractor that reads Bitcoin Core debug.logs and produces events for parsed log messages.

A few followups include:

from @0xB10C in https://github.com/0xB10C/peer-observer/pull/272#issuecomment-3457201310
- add more matchers for different log messages
- add support for multiline matching. e.g. the `bench` block connection logs
- look into fuzzing the matcher input to see if it breaks somewhere
- ideally, the matcher should return an error on e.g. invalid timestamps and not 0

from @m4ycon in https://github.com/0xB10C/peer-observer/pull/272#issuecomment-3458843192

- How to organize a LogMatcher to support many implementations of a same log (e.g. diff btc-core versions), don't think that it'll be difficult but deciding on some pattern it's important.

>"ideally, the matcher should return an error on e.g. invalid timestamps and not 0"

if you say this, I assume that an unexpected error on log-extractor parsing problably won't stop its loop, right? If so I'm ok with it.
 - Before I was thinking that in cases where we don't match the log (even common infos, e.g. timestamp) or get some error while parsing, we could still store the raw message (when https://github.com/0xB10C/peer-observer/issues/121 is implemented), maybe to debug later
- Also saying this because I saw you did a test that breaks the timestamp but it stores an empty raw message ([test_log_matcher_with_broken_timestamp](https://github.com/0xB10C/peer-observer/pull/272/files#diff-3177bf62a6f249dea5b7d0551575baae6f82432745ce1b8edc5dbb7ea4073114R242))

m4ycon:
About multiline matching, I've done some sketches. Log extractor could have a state of "pending matchers", `LogMatcher` trait would change to require two methods:

- `consume(line: String) -> bool`: every incoming log must pass through this method on each `LogMatcher` implementation. If it returns `true`, it indicates that this log was succesfully consumed/matched.
- `get_event() -> Option<Event>`: whenever `consume(line)` returns `true`, it should try to get the event. If `None`, it would store in a log extractor's state; otherwise, it would send to NATS.

I suppose multiline logs like `bench` ones don't get mixed up with something else, but even if it does, I think this structure would handle it fine.

Each `LogMatcher` would implement this trait in its own way, I will not enforce specific attributes, handling the inner state will be entirely the responsability of the implementation. This prevents unnecessary complexity in cases where it's just a single line.

[...]

PR #379, opened 2026-03-13
Add IPC extractor

### LINKED ISSUE #370: Implement an experimental ipc-extractor against the Bitcoin Core chain IPC interface (proposed in #29409)
https://github.com/bitcoin/bitcoin/pull/29409 adds a capnp wrapper for the Bitcoin Core chain interface. To test this, and give feedback on the PR, it would be cool to build an experimental IPC extractor.

xyzconstant:
## Description

This PR introduces an experimental IPC extractor, designed to interface with a [Core v31.0](https://github.com/bitcoin/bitcoin/tree/v31.0) `bitcoin-node` process over a Unix Socket (generated by passing `-ipcbind=unix`) and Cap 'n Proto. 

## Context

Bitcoin Core has an on-going [Process Separation](https://github.com/bitcoin-core/bitcoin-devwiki/wiki/Process-Separation) project, which aims to split `bitcoind` into smaller and manageable processes that communicate together through Unix Sockets. Within this architecture, `bitcoin-node` ([introduced in Core 30.0](https://bitcoincore.org/en/releases/30.0/#install-changes)) exposes multiple interfaces that external processes can consume by leveraging Cap 'n Proto RPC system.

As `bitcoin-node` keeps expanding its interfaces (e.g bitcoin/bitcoin#29409 adds a wrapper for the Chain interface), this extractor will facilitate reviewing and testing on IPC-related PRs in Core.

## Key changes

- Add new binary crate `ipc-extractor`.
- Add new Github Action for verifying and installing `bitcoin-node` (for Ubuntu tests).
- Vendor `.capnp` files from [Bitcoin Core v31.0](https://github.com/bitcoin/bitcoin/tree/v31.0).
- Add new dependencies to integrate Cap 'n Proto (`capnpc`, `capnp-rpc` and `capnp`).
- Add Protocol Buffer schema and NATS subject for IPC events.
- Update tools to enable inspecting IPC events.
- Update `README.md` with documentation for the new extractor.

## How to use

1. Download and compile the code at https://github.com/bitcoin/bitcoin/tree/v31.0
2. Launch a `bitcoin-node` process: `./build/bin/bitcoin-node -regtest -ipcbind=unix OR ./build/bin/bitcoin -m node -regtest -ipcbind=unix`
3. Start the NATS server: `nats-server`
4. Run the logger tool: `cargo run --bin logger -- --ipc`
5. Run the extractor: `cargo run --bin ipc-extractor -- --ipc-socket-path="path/to/node.sock"`
6. Observe `BlockTip(height=404, hash=...)` events being published to NATS.
image

Closes #370

### DIFF
--- a/.github/actions/install-bitcoin/action.yml
+++ b/.github/actions/install-bitcoin/action.yml
@@ -1,0 +1,79 @@
+name: 'Install and Verify Bitcoin Core'
+description: 'Downloads, Verify signatures and installs Bitcoin Core'
+
+inputs:
+  version:
+    description: 'Bitcoin Core version to install'
+    required: true
+    default: '31.0'
+  trusted_keys:
+    description: 'List of GPG fingerprints to validate'
+    required: false
+    # Signatures from the following GPG public keys checked during verification of the source code.
+    # The list can be found at https://github.com/bitcoin-core/guix.sigs/tree/main/builder-keys
+    # 15281230078...: achow101.gpg
+    # 9EDAFF80E08...: Emzy.gpg
+    # D1DBF2C4B96...: hebasto.gpg
+    default: >-
+      152812300785C96444D3334D17565732E08E5E41
+      9EDAFF80E080659604F4A76B2EBB056FD847F8A7
+      D1DBF2C4B96F2DEBF4C16654410108112E7EA81F
+
+runs:
+  using: "composite"
+  steps:
+    - name: Install verification dependencies
+      shell: bash
+      run: sudo apt-get update && sudo apt-get install -y gnupg wget
+
+    - name: Download and Verify Bitcoin Core
+      shell: bash
+      run: |
+        VERSION="${{ inputs.version }}"
+        
+        # 1. Download Binaries & Manifests
+        wget -q "https://bitcoincore.org/bin/bitcoin-core-${VERSION}/bitcoin-${VERSION}-x86_64-linux-gnu.tar.gz"
+        wget -q "https://bitcoincore.org/bin/bitcoin-core-${VERSION}/SHA256SUMS"
+        wget -q "https://bitcoincore.org/bin/bitcoin-core-${VERSION}/SHA256SUMS.asc"
+
+        # 2. Fetch builder keys
+        git clone --depth 1 https://github.com/bitcoin-core/guix.sigs
+        
+        # 3. Setup temporary GPG environment
+        export GNUPGHOME=$(mktemp -d)
+        gpg --batch --import guix.sigs/builder-keys/*.gpg
+
+        # 4. Verify the SHA256SUMS signature
+        gpg --batch --verify --status-fd 1 SHA256SUMS.asc SHA256SUMS > verify.log
+        
+        # 5. Check against trusted fingerprints
+        TRUSTED_KEYS=(${{ inputs.trusted_keys }})
+        MATCH=false
+        for KEY in "${TRUSTED_KEYS[@]}"; do
+          if grep -q "VALIDSIG .* $KEY" verify.log; then
+            echo "Verified signature from trusted developer: $KEY"
+            MATCH=true
+            break
+          fi
+        done
+
+        if [ "$MATCH" != true ]; then
+          echo "No signatures from the trusted key list were found."
+          exit 1
+        fi
+
+        # 6. Verify checksum and extract
+        sha256sum --ignore-missing --check SHA256SUMS
+        tar -xzf bitcoin-${VERSION}-x86_64-linux-gnu.tar.gz
+        
+        # 7. Install to path (core v31.0 lists `bitcoin-node` in libexec/)
+        sudo install -m 0755 -o root -g root -t /usr/local/bin bitcoin-${VERSION}/bin/*
+        sudo install -m 0755 -o root -g root -t /usr/local/bin bitcoin-${VERSION}/libexec/*
+        
+        # Cleanup workspace
+        rm -rf guix.sigs bitcoin-${VERSION}* SHA256SUMS* verify.log
+        echo "Bitcoin Core v${VERSION} installed successfully."
+
+    - name: Verify Installation
+      shell: bash
+      run: bitcoind --version
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -106,10 +106,14 @@ jobs:
     steps:
     - uses: actions/checkout@v6
     - uses: ./.github/actions/setup-ubuntu-action
+    - name: Setup Bitcoin
+      uses: ./.github/actions/install-bitcoin
+      with:
+        version: '31.0'
     - name: Install integration test dependecies
       run: sudo apt-get install -y nats-server
     - name: Run tests and integration tests
-      run: NATS_SERVER_BINARY=$(which nats-server) cargo test --all-features
+      run: NATS_SERVER_BINARY=$(which nats-server) BITCOIN_NODE_EXE=$(which bitcoin-node) cargo test --all-features
 
   ubuntu-test-detect-intermittent-failures:
     runs-on: ubuntu-latest
@@ -117,29 +121,33 @@ jobs:
     steps:
     - uses: actions/checkout@v6
     - uses: ./.github/actions/setup-ubuntu-action
+    - name: Setup Bitcoin
+      uses: ./.github/actions/install-bitcoin
+      with:
+        version: '31.0'
     - name: Install integration test dependecies
       run: sudo apt-get install -y nats-server
     # use separate steps for iterations here to easily see on GHA the time it took
     - name: Iteration 0
-      run: NATS_SERVER_BINARY=$(which nats-server) cargo test --all-features
+      run: NATS_SERVER_BINARY=$(which nats-server) BITCOIN_NODE_EXE=$(which bitcoin-node) cargo test --all-features
     - name: Iteration 1
-      run: NATS_SERVER_BINARY=$(which nats-server) cargo test --all-features
+      run: NATS_SERVER_BINARY=$(which nats-server) BITCOIN_NODE_EXE=$(which bitcoin-node) cargo test --all-features
     - name: Iteration 2
-      run: NATS_SERVER_BINARY=$(which nats-server) cargo test --all-features
+      run: NATS_SERVER_BINARY=$(which nats-server) BITCOIN_NODE_EXE=$(which bitcoin-node) cargo test --all-features
     - name: Iteration 3
-      run: NATS_SERVER_BINARY=$(which nats-server) cargo test --all-features
+      run: NATS_SERVER_BINARY=$(which nats-server) BITCOIN_NODE_EXE=$(which bitcoin-node) cargo test --all-features
     - name: Iteration 4
-      run: NATS_SERVER_BINARY=$(which nats-server) cargo test --all-features
+      run: NATS_SERVER_BINARY=$(which nats-server) BITCOIN_NODE_EXE=$(which bitcoin-node) cargo test --all-features
     - name: Iteration 5
-      run: NATS_SERVER_BINARY=$(which nats-server) cargo test --all-features
+      run: NATS_SERVER_BINARY=$(which nats-server) BITCOIN_NODE_EXE=$(which bitcoin-node) cargo test --all-features
     - name: Iteration 6
-      run: NATS_SERVER_BINARY=$(which nats-server) cargo test --all-features
+      run: NATS_SERVER_BINARY=$(which nats-server) BITCOIN_NODE_EXE=$(which bitcoin-node) cargo test --all-features
     - name: Iteration 7
-      run: NATS_SERVER_BINARY=$(which nats-server) cargo test --all-features
+      run: NATS_SERVER_BINARY=$(which nats-server) BITCOIN_NODE_EXE=$(which bitcoin-node) cargo test --all-features
     - name: Iteration 8
-      run: NATS_SERVER_BINARY=$(which nats-server) cargo test --all-features
+      run: NATS_SERVER_BINARY=$(which nats-server) BITCOIN_NODE_EXE=$(which bitcoin-node) cargo test --all-features
     - name: Iteration 9
-      run: NATS_SERVER_BINARY=$(which nats-server) cargo test --all-features
+      run: NATS_SERVER_BINARY=$(which nats-server) BITCOIN_NODE_EXE=$(which bitcoin-node) cargo test --all-features
 
   ubuntu-docs-tools:
     needs: ubuntu-build
--- a/README.md
+++ b/README.md
@@ -110,7 +110,7 @@ I'll do my best to add more documentation.
 
 To run the integration tests, run with the feature `nats_integration_tests` and
 `node_integration_tests`. If you are not using the nix-shell, you need to set the
-`NATS_SERVER_BINARY` to the path to your `nats-server` binary. Addtionally,
+`NATS_SERVER_BINARY` to the path to your `nats-server` binary and `BITCOIN_NODE_EXE` to the path to a `bitcoin-node` binary. Additionally,
 `BITCOIND_EXE` can be set to a custom `bitcoind` binary. By default, a recent
 release will be downloaded and used if `BITCOIND_SKIP_DOWNLOAD` is unset.
 
--- a/README.md
+++ b/README.md
@@ -24,6 +24,8 @@ selected P2P measurements as events into a NATS pub-sub queue.
 
 The `log-extractor` publishes them parsed `debug.log` log messages as events to NATS.
 
+And finally an experimental `ipc-extractor` which periodically fetch data from a `bitcoin-node` binary through a UNIX socket created with the `-ipcbind` option. Publishes to NATS as IPC events.
+
 The tools are written in Rust (or any other language that supports NATS 
 and protobuf). They subscribe to the NATS server. For example, the `logger` tool 
 simply prints out all messages that it receives, the `metrics` tool produces prometheus 
@@ -36,21 +38,27 @@ messages. For other languages, types can be generated directly from the Protobuf
                                            messages
 ┌─────────────┐    ┌───────────────┐     ┌─────────┐      ┌──────────────────────┐
 │             ├────► ebpf-extractor├─────┤         │      │                      │
-│             │    └───────────────┘     │         │      │ Tools                │
-│             │                          │         │      │                      │
-│             │    ┌───────────────┐     │         ├──────┼──►logger             │
+│             │    └───────────────┘     │         │      │                      │
+│             │                          │         │      │ Tools                │
+│             │    ┌───────────────┐     │         │      │                      │
 │             ├────► rpc-extractor │─────┤         │      │                      │
-│   Bitcoin   │    └───────────────┘     │ NATS.io ├──────┼──►metrics            │
+│             │    └───────────────┘     │         ├──────┼──►logger             │
 │             │                          │         │      │                      │
-│     Node    │    ┌───────────────┐     │ PUB-SUB ├──────┼──►websocket          │
+│   Bitcoin   │    ┌───────────────┐     │ NATS.io ├──────┼──►metrics            │
 │             ├────► p2p-extractor │─────┤         │      │                      │
-│             │    └───────────────┘     │         ├──────┼──►connectivity-check │
+│     Node    │    └───────────────┘     │ PUB-SUB ├──────┼──►websocket          │
 │             │                          │         │      │                      │
-│             │    ┌───────────────┐     │         │      │   ...                │
+│             │    ┌───────────────┐     │         ├──────┼──►connectivity-check │
 │             ├────► log-extractor │─────┤         │      │                      │
+│             │    └───────────────┘     │         │      │   ...                │
+│             │                          │         │      │                      │
+│             │    ┌───────────────┐     │         │      │                      │
+│             ├────► ipc-extractor ├─────┤         │      │                      │
 └─────────────┘    └───────────────┘     └─────────┘      └──────────────────────┘
 
  (edit on asciiflow.com)
+
+
 ```
 
 [nats.io]: https://nats.io
@@ -69,6 +77,7 @@ NATS server. Each extractor connects to a different interface:
 | rpc           | periodically fetches RPC for events   | [extractors/rpc/](extractors/rpc)   |
 | p2p           |Bitcoin P2P events from an inbound node| [extractors/p2p/](extractors/p2p)   |
 | log           | parses the debug.log of a node        | [extractors/log/](extractors/log)   |
+| ipc           | Fetch data over via IPC socket (experimental) | [extractors/ipc/](extractors/ipc)    |
 
 ## Tools
 
--- a/extractors/ipc/Cargo.toml
+++ b/extractors/ipc/Cargo.toml
@@ -1,0 +1,17 @@
+[package]
+name = "ipc-extractor"
+version = "0.1.0"
+edition = "2024"
+
+[dependencies]
+shared = { path = "../../shared" }
+
+[features]
+# Treat warnings as a build error.
+strict = []
+
+# Run integration tests needing a NATS server.
+nats_integration_tests = []
+
+# Run integration tests needing a Bitcoin Core node.
+node_integration_tests = []
--- a/extractors/ipc/Cargo.toml
+++ b/extractors/ipc/Cargo.toml
@@ -4,7 +4,10 @@ version = "0.1.0"
 edition = "2024"
 
 [dependencies]
+capnp = "0.25.1"
+capnp-rpc = "0.25.0"
 shared = { path = "../../shared" }
+tokio-util = { version = "0.7.18", features = ["compat"] }
 
 [features]
 # Treat warnings as a build error.
@@ -15,3 +18,6 @@ nats_integration_tests = []
 
 # Run integration tests needing a Bitcoin Core node.
 node_integration_tests = []
+
+[build-dependencies]
+capnpc = "0.25.0"
--- a/extractors/ipc/README.md
+++ b/extractors/ipc/README.md
@@ -2,7 +2,7 @@
 
 > publishes data fetched from IPC
 
-A peer-observer extractor that periodically queries the Bitcoin Core IPC interfaces and publishes the results as events into a NATS pub-sub queue.
+The peer-observer ipc-extractor periodically queries data from the Bitcoin Core IPC interface and publishes the results as events into a NATS pub-sub queue
 
 ## Usage
 
@@ -13,6 +13,8 @@ The peer-observer ipc-extractor periodically queries data from the Bitcoin Core 
 Usage: ipc-extractor [OPTIONS] --ipc-socket-path <IPC_SOCKET_PATH>
 
 Options:
+  -i, --ipc-socket-path <IPC_SOCKET_PATH>
+          A path to an UNIX socket to read IPC data from
   -a, --nats-address <ADDRESS>
           The NATS server address the extractor/tool should connect and subscribe to [default: 127.0.0.1:4222]
   -u, --nats-username <USERNAME>
@@ -23,8 +25,6 @@ Options:
           A path to a file containing a password the extractor/tool should try to authentificate to the NATS server with
   -l, --log-level <LOG_LEVEL>
           The log level the extractor should run with. Valid log levels are "trace", "debug", "info", "warn", "error". See https://docs.rs/log/latest/log/enum.Level.html [default: DEBUG]
-      --ipc-socket-path <IPC_SOCKET_PATH>
-          A UNIX socket path to read IPC data from
       --query-interval <QUERY_INTERVAL>
           Interval (in seconds) in which to query from the Bitcoin Core IPC interface [default: 10]
   -h, --help
--- a/extractors/ipc/README.md
+++ b/extractors/ipc/README.md
@@ -1,0 +1,34 @@
+# `ipc` extractor
+
+> publishes data fetched from IPC
+
+A peer-observer extractor that periodically queries the Bitcoin Core IPC interfaces and publishes the results as events into a NATS pub-sub queue.
+
+## Usage
+
+```
+$ cargo run --bin ipc-extractor -- --help
+The peer-observer ipc-extractor periodically queries data from the Bitcoin Core IPC endpoint and publishes the results as events into a NATS pub-sub queue
+
+Usage: ipc-extractor [OPTIONS] --ipc-socket-path <IPC_SOCKET_PATH>
+
+Options:
+  -a, --nats-address <ADDRESS>
+          The NATS server address the extractor/tool should connect and subscribe to [default: 127.0.0.1:4222]
+  -u, --nats-username <USERNAME>
+          The NATS username the extractor/tool should try to authentificate to the NATS server with
+  -p, --nats-password <PASSWORD>
+          The NATS password the extractor/tool should try to authentificate to the NATS server with
+  -f, --nats-password-file <PASSWORD_FILE>
+          A path to a file containing a password the extractor/tool should try to authentificate to the NATS server with
+  -l, --log-level <LOG_LEVEL>
+          The log level the extractor should run with. Valid log levels are "trace", "debug", "info", "warn", "error". See https://docs.rs/log/latest/log/enum.Level.html [default: DEBUG]
+      --ipc-socket-path <IPC_SOCKET_PATH>
+          A UNIX socket path to read IPC data from
+      --query-interval <QUERY_INTERVAL>
+          Interval (in seconds) in which to query from the Bitcoin Core IPC interface [default: 10]
+  -h, --help
+          Print help
+  -V, --version
+          Print version
+```
--- a/extractors/ipc/README.md
+++ b/extractors/ipc/README.md
@@ -1,0 +1,47 @@
+# `ipc` extractor
+
+> publishes data fetched from IPC
+
+A peer-observer extractor that periodically queries the Bitcoin Core IPC interfaces and publishes the results as events into a NATS pub-sub queue.
+
+## Example
+
+Start `bitcoin-node` with `-ipcbind=unix` to expose the IPC interface over a UNIX socket. With the bare `unix` value the socket is created at `<datadir>/node.sock` relative to the *network* data directory, for instance regtest with a default datadir results in the socket located at `~/.bitcoin/regtest/node.sock`. `~/.bitcoin/node.sock` for mainnet.
+
+For example, run a regtest node and point the ipc-extractor at the generated socket:
+
+```bash
+$ bitcoin-node -regtest -ipcbind=unix &
+$ cargo run --bin ipc-extractor -- --ipc-socket-path ~/.bitcoin/regtest/node.sock
+```
+
+Note that `bitcoin-node` is compiled by setting `-DENABLE_IPC=ON` when building from source. Alternatively, there is also a proxy binary named `bitcoin` that lets you spawn a specific process with the `-m` option. That's it, running `bitcoin -m node -regtest -ipcbind=unix` is equivalent to the start command above.
+
+## Usage
+
+```
+$ cargo run --bin ipc-extractor -- --help
+The peer-observer ipc-extractor periodically queries data from the Bitcoin Core IPC endpoint and publishes the results as events into a NATS pub-sub queue
+
+Usage: ipc-extractor [OPTIONS] --ipc-socket-path <IPC_SOCKET_PATH>
+
+Options:
+  -a, --nats-address <ADDRESS>
+          The NATS server address the extractor/tool should connect and subscribe to [default: 127.0.0.1:4222]
+  -u, --nats-username <USERNAME>
+          The NATS username the extractor/tool should try to authentificate to the NATS server with
+  -p, --nats-password <PASSWORD>
+          The NATS password the extractor/tool should try to authentificate to the NATS server with
+  -f, --nats-password-file <PASSWORD_FILE>
+          A path to a file containing a password the extractor/tool should try to authentificate to the NATS server with
+  -l, --log-level <LOG_LEVEL>
+          The log level the extractor should run with. Valid log levels are "trace", "debug", "info", "warn", "error". See https://docs.rs/log/latest/log/enum.Level.html [default: DEBUG]
+      --ipc-socket-path <IPC_SOCKET_PATH>
+          A UNIX socket path to read IPC data from
+      --query-interval <QUERY_INTERVAL>
+          Interval (in seconds) in which to query from the Bitcoin Core IPC interface [default: 10]
+  -h, --help
+          Print help
+  -V, --version
+          Print version
+```
--- a/extractors/ipc/README.md
+++ b/extractors/ipc/README.md
@@ -2,7 +2,7 @@
 
 > publishes data fetched from IPC
 
-A peer-observer extractor that periodically queries the Bitcoin Core IPC interfaces and publishes the results as events into a NATS pub-sub queue.
+The peer-observer ipc-extractor periodically queries data from the Bitcoin Core IPC interface and publishes the results as events into a NATS pub-sub queue
 
 ## Example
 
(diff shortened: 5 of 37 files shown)


extractors/ipc/src/lib.rs
```diff
@@ -0,0 +1,94 @@
...
+use shared::nats_util::{self, NatsArgs};
+use shared::tokio::sync::watch;
+use shared::tokio::time::{self, Duration};
+
+mod error;
+
+use error::RuntimeError;
+
+/// The peer-observer ipc-extractor periodically queries data from the
+/// Bitcoin Core IPC endpoint and publishes the results as events into
```
0xB10C: in f777a2978a3ba6447a38d4c836c572a8b660b220: `endpoint` -> `interface`

extractors/ipc/capnp/chain.capnp
```diff
@@ -0,0 +1,172 @@
...
+
+interface ChainNotifications $Proxy.wrap("interfaces::Chain::Notifications") {
+    destroy @0 (context :Proxy.Context) -> ();
+    transactionAddedToMempool @1 (context :Proxy.Context, tx :Data) -> ();
+    transactionRemovedFromMempool @2 (context :Proxy.Context, tx :Data, reason :Int32) -> ();
+    blockConnected @3 (context :Proxy.Context, role: ChainstateRole, block :BlockInfo) -> ();
+    blockDisconnected @4 (context :Proxy.Context, block :BlockInfo) -> ();
+    updatedBlockTip @5 (context :Proxy.Context) -> ();
+    chainStateFlushed @6 (context :Proxy.Context, role: ChainstateRole, locator :Data) -> ();
+}
```
0xB10C: being able to subscribe to some of these would be very cool!

0xB10C (commented):
Thank you, that's awesome!

I haven't run this yet, but I'm planning to play around a bit (soon ™️ )! A few thoughts:

- if we want to merge the `ipc-extractor` in the short-term, I think having it support the mining interface (which is merged and released in v30 and v31) and one or two IPC calls from there would be good
- then, we can maintain a patch for the chain interface and use it for futher testing
- and even have a patch for a WIP IPC tracing interface, if we decide to work on this
- Since I expect the captn' proto files to change from time to time (i think it's still far from final), it would be good to think about how to best support multiple Bitcoin Core releases / versions. Maybe we can supply them during build step. However, breaking API changes in captn' proto will probably cause a few headaches for us, so best to treat the ipc-extractor as experimental for a while

xyzconstant:
> * if we want to merge the `ipc-extractor` in the short-term, I think having it support the mining interface (which is merged and released in v30 and v31) and one or two IPC calls from there would be good

Yeah sure, for me should be replacing current `Chain::getHeight` with mining's `Mining::getTip` and obtain the block height from there. Regarding the `.capnp` schemas these should mirror v30's ones instead of bitcoin/bitcoin#29409

> * then, we can maintain a patch for the chain interface and use it for futher testing
> * and even have a patch for a WIP IPC tracing interface, if we decide to work on this

Agree, it would be amazing to test the tracing interface.

> I think mentioning it's experiment in the READMEs is fine for now. No need to rename it to `experimental-ipc-extractor`.

Great

> This also means implementing IPC support in the infrastructure https://github.com/peer-observer/infra-library at some point.

I'm not familiar with that repository but you mean to add changes like node setup? i.e passing `ipcbind=unix` or is it more about build stuff for the extractor? anyways I'll gave it a look later, but I might say I'm not familiar with Nix neither, I tested the extractor locally without a container.

> * then, maintain a patchset (possibly in a WIP PR) that implements some functionallity for the chain interface as proposed in [multiprocess: Add capnp wrapper for Chain interface bitcoin/bitcoin#29409](https://github.com/bitcoin/bitcoin/pull/29409) and use it to test that manually. We can't really test this in CI yet, I think, as we don't have a good way of running custom Bitcoin Core binaries in CI (yet). Also, the demo and production peer-observer nodes won't all be running [multiprocess: Add capnp wrapper for Chain interface bitcoin/bitcoin#29409](https://github.com/bitcoin/bitcoin/pull/29409) until it's merged
> * use the Chain interface one to play around an implement a few cool demos or PoCs, and give feedback on the Bitcoin Core PR
> * possibly in paralell, experiment with a custom IPC Tracing interface in Bitcoin Core, see how much performance we can get out of it, if we can improve the performance, and evaluate if it would be an option

That's awesome, thanks

extractors/ipc/Cargo.toml
```diff
@@ -4,7 +4,10 @@ version = "0.1.0"
 edition = "2024"
 
 [dependencies]
+capnp = "0.25.1"
+capnp-rpc = "0.25.0"
```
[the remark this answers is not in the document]
xyzconstant: Done at 2b3cdc1b8136b59f5f4d92e6f2836da58374205b

extractors/ipc/src/lib.rs
xyzconstant: Corrected at ecab94f9ee9496eab0ead36f8dd7bc8e1ffd1d57

extractors/ipc/capnp/mp/proxy.capnp
```diff
@@ -0,0 +1,65 @@
...
+
+annotation wrap(interface, struct): Text;
+# Wrap capnp interface generating ProxyClient / ProxyServer C++ classes that
+# forward calls to a C++ interface with same methods and parameters. Text
+# string should be the name of the C++ interface.
+# If applied to struct rather than an interface, this will generate a ProxyType
+# struct with get methods for introspection and copying fields between C++ and
+# capnp structs.
+
+annotation count(param, struct, interface): Int32;
```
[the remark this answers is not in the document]
xyzconstant: Not having this file in here would cause the other `.capnp` files to break at build time because all of them (i.e [mining.capnp](https://github.com/xyzconstant/peer-observer/blob/2b3cdc1b8136b59f5f4d92e6f2836da58374205b/extractors/ipc/capnp/mining.capnp#L11)) import the proxy. I'm not sure if there's another way to fix this (since importing missing `c++.capnp` or `.h` files won't break) but I'd prefer to have these schemas as untouched as possible, it's a mirror to me after all.

0xB10C:
> I'm not familiar with that repository but you mean to add changes like node setup? i.e passing ipcbind=unix or is it more about build stuff for the extractor? anyways I'll gave it a look later, but I might say I'm not familiar with Nix neither, I tested the extractor locally without a container.

No worries! I can take care of that!

extractors/ipc/capnp/chain.capnp
xyzconstant: Maybe in the patch version? Just removed it because it's not currently available at v30

xyzconstant:
Update:

I was working on adding integration tests and noticed that there's no trivial way to mock/setup a `bitcoin-node` process with `corepc-node`, haven't gone deep but at first glance it looks like https://github.com/0xb10c/corepc needs some update for this.

xyzconstant:
Ok I think I found a workaround for spinning up a `bitcoin-node` instance with `corepc-node`: we just need to disable the wallet via `conf.wallet = None;`.

But the new challenge is to get to build `bitcoin-node` binary in bitcoind's nix repository.
image

extractors/ipc/src/lib.rs
```diff
@@ -93,8 +117,103 @@ pub async fn run(args: Args, mut shutdown_rx: watch::Receiver<bool>) -> Result<(
...
+    let stream = UnixStream::connect(ipc_socket_path).await.map_err(|e| {
+        io::Error::new(
+            e.kind(),
+            format!(
+                "could not connect to IPC socket at --ipc-socket-path '{}': {}",
+                ipc_socket_path, e
+            ),
+        )
+    })?;
+    log::info!("Connected to IPC socket path at {}", ipc_socket_path);
```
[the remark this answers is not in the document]
xyzconstant: Done (squashed) at 523c37b

deadmanoz:
> Ok I think I found a workaround for spinning up a `bitcoin-node` instance with `corepc-node`: we just need to disable the wallet via `conf.wallet = None;`.

Yeah, I think that's correct. I also discovered something similar in making some small changes to corepec: https://github.com/rust-bitcoin/corepc/issues/429

0xB10C:
> But the new challenge is to get to build bitcoin-node binary in bitcoind's nix repository.

I think we can do that! Should not be a problem. I can have a look.

0xB10C:
I came across this: https://github.com/2140-dev/bitcoin-capnp-types/ which might be interesting, but I think we don't strictly need it as a depdency for now.

0xB10C:
> > But the new challenge is to get to build bitcoin-node binary in bitcoind's nix repository.
> 
> I think we can do that! Should not be a problem. I can have a look.

I checked and it's seems the binary is compiled in nixpkgs master:

```
$ nix build .#bitcoind --print-build-logs --rebuild
bitcoind> Configure summary
bitcoind> =================
bitcoind> Executables:
bitcoind>   bitcoin ............................. ON
bitcoind>   bitcoind ............................ ON
bitcoind>   bitcoin-node (multiprocess) ......... ON
bitcoind>   bitcoin-qt (GUI) .................... OFF
bitcoind>   bitcoin-gui (GUI, multiprocess) ..... OFF
bitcoind>   bitcoin-cli ......................... ON
bitcoind>   bitcoin-tx .......................... ON
bitcoind>   bitcoin-util ........................ ON
bitcoind>   bitcoin-wallet ...................... ON
bitcoind>   bitcoin-chainstate (experimental) ... OFF
bitcoind>   libbitcoinkernel (experimental) ..... OFF
bitcoind> Optional features:
bitcoind>   wallet support ...................... ON
bitcoind>   external signer ..................... ON
bitcoind>   ZeroMQ .............................. TRUE
bitcoind>   IPC ................................. ON
bitcoind>   USDT tracing ........................ TRUE
bitcoind>   QR code (GUI) ....................... OFF
bitcoind>   DBus (GUI) .......................... OFF
bitcoind> Tests:
bitcoind>   test_bitcoin ........................ ON
bitcoind>   test_bitcoin-qt ..................... OFF
bitcoind>   bench_bitcoin ....................... FALSE
bitcoind>   fuzz binary ......................... OFF
...
```

However, the `bitcoin-node` binary isn't listed in the nixpkgs docs, as it's not in `bin`. It's in `libexec` (see https://bitcoincore.org/en/releases/30.0/#install-changes)

So we should add

```bash
export BITCOIN_NODE_EXE=${pkgs.bitcoind}/libexec/bitcoin-node
```
 
to
 
https://github.com/peer-observer/peer-observer/blob/de427d5846a5b16c20923e28318b9f2b134372ed/shell.nix#L44

Does this help?

xyzconstant:
> I came across this: https://github.com/2140-dev/bitcoin-capnp-types/ which might be interesting, but I think we don't strictly need it as a depdency for now.

By coincidence, lately I've been thinking about doing something very similar: a repo which exports the the generated capnproto bindings (exactly the same as what you shared) with the plus of maintaining the bindings for IPC-related PRs (i.e bitcoin/bitcoin#32898, bitcoin/bitcoin#29409 and maybe others). Thanks for sharing!

xyzconstant:
> However, the `bitcoin-node` binary isn't listed in the nixpkgs docs, as it's not in `bin`. It's in `libexec` (see https://bitcoincore.org/en/releases/30.0/#install-changes)

Wow I missed this, thanks for noticing it!

> So we should add
> 
> ```shell
> export BITCOIN_NODE_EXE=${pkgs.bitcoind}/libexec/bitcoin-node
> ```
> 
> to
> 
> https://github.com/peer-observer/peer-observer/blob/de427d5846a5b16c20923e28318b9f2b134372ed/shell.nix#L44
> 
> Does this help?

Yeah it worked! Thanks to this I noticed I wasn't vendoring actual `capnp` schemas at v30.2 so I had to revert those and now the whole is working properly for v30.2. Thanks!

.github/workflows/ci.yml
0xB10C, line 131: I think you have to give it BITCOIN_NODE_EXE here too

xyzconstant: Yeah just pushed it

0xB10C:
I was wondering what happens when we run a different (newer) Bitoin Core version than 30.2 and the schemas don't match anymore. I usually run a recent `master` and currently the release candidates. If it breaks and we can't use the ipc-extractor with a newer node version, we could think about a way of supplying the schemas during compilation as build inputs, based on the Bitcoin Core node we're building against. Not needed for this PR though.

Also, we can test running a Bitcoin Core v31.0rc1 against this version of the IPC-extractor to see if it works, and if not, what changes are required.

xyzconstant:
> I was wondering what happens when we run a different (newer) Bitoin Core version than 30.2 and the schemas don't match anymore. I usually run a recent `master` and currently the release candidates. If it breaks and we can't use the ipc-extractor with a newer node version, we could think about a way of supplying the schemas during compilation as build inputs, based on the Bitcoin Core node we're building against. Not needed for this PR though.

We also should think about preventing these schemas from breaking rust code, I'll check this out but my first guess is that this should be handled manually.

> Also, we can test running a Bitcoin Core v31.0rc1 against this version of the IPC-extractor to see if it works, and if not, what changes are required.

It raises this error:
```
thread 'test_integration_ipc' (75797074) panicked at extractors/ipc/tests/integration.rs:122:21:
ipc_extractor stopped unexpectedly: Err(IpcCall(Error { kind: Failed, extra: "remote exception: std::exception: Old mining interface (@2) not supported. Please update your client!" }))
```

As mentioned before, I was pointing to `.capnp` proto files at master so I think the fix is just to update them.

extractors/ipc/src/lib.rs
```diff
@@ -0,0 +1,139 @@
...
+                e.kind(),
+                format!(
+                    "could not connect to IPC socket at --ipc-socket-path '{}': {}",
+                    &args.ipc_socket_path, e
+                ),
+            )
+        })?;
+    log::info!("Connected to IPC socket at {}", &args.ipc_socket_path);
+
+    let ipc_session = IpcClient::init(stream).await?;
```
GuiSchet, line 87: If `bitcoin-node` restarts, `get_tip()` will fail on every tick and log errors indefinitely without attempting to reconnect.

GuiSchet (commented):
Image

awesome! it works!!

extractors/ipc/src/lib.rs
xyzconstant: Awesome, I haven't personally tested it to break like that. Thanks for pointing it out @GuiSchet!

xyzconstant: After thinking about this for a few minutes, I noticed that there's no reconnection logic on fetching in any of the extractors neither

0xB10C: When the socket comes back online (i.e. Bitcoin Core comes back up) will it work again? 

If not, then I think shutting down the extractor is fine. May log that it's shutting down to Bitcoin Core not being reachable anymore.

For both the rpc-extractor and the log-extractor, the node can restart and the extractors with start working again. I'm not sure about the ebpf one on top of my head. For the p2p-extractor, the reconnection logic is actually in Bitcoin Core, since the node makes a connection to the p2p-extractor. So when it comes back up, the node will connect to the p2p-extractor again.

GuiSchet: I tried shutting down Core and up again, get_tip() failed on ticks, but I think is not important right now.

IPC brings a lot of new things to work on! nice work @xyzconstant!

xyzconstant: > When the socket comes back online (i.e. Bitcoin Core comes back up) will it work again?

No, it won't.

> If not, then I think shutting down the extractor is fine. May log that it's shutting down to Bitcoin Core not being reachable anymore.

Does this work? ba08e803f32af04c0136cb76b6e87ebbea31610d

xyzconstant:
Rebased to solve conflicts with master + bumped support for v31.0.

ba322337b8702ff275ace43b2f13a49e25323e37 -> 0bb507e

Nix tests will fail until NixOS/nixpkgs#511775 arrives

0xB10C:
This is still on my todo list for testing! Will probably be after https://github.com/NixOS/nixpkgs/pull/511775 is merged.

README.md
```diff
@@ -101,7 +110,7 @@ I'll do my best to add more documentation.
 
 To run the integration tests, run with the feature `nats_integration_tests` and
 `node_integration_tests`. If you are not using the nix-shell, you need to set the
-`NATS_SERVER_BINARY` to the path to your `nats-server` binary. Addtionally,
+`NATS_SERVER_BINARY` to the path to your `nats-server` binary and `BITCOIN_NODE_EXE` to the path to a `bitcoin-node` binary. Addtionally,
```
ViniciusCestarii: nit (pre-existing): Addtionally → Additionally. Since this line was already touched, it may be worth fixing while we're here.

.github/workflows/ci.yml
```diff
@@ -106,40 +106,46 @@ jobs:
...
+      run: NATS_SERVER_BINARY=$(which nats-server) BITCOIN_NODE_EXE=$(which bitcoin-node) cargo test --all-features
 
   ubuntu-test-detect-intermittent-failures:
     runs-on: ubuntu-latest
     needs: ubuntu-test
     steps:
     - uses: actions/checkout@v6
     - uses: ./.github/actions/setup-ubuntu-action
+    - name: Setup Bitcoin
+      uses: ./.github/actions/install-bitcoin
```
ViniciusCestarii, line 125: nit: ubuntu-test pins version: '31.0' explicitly. This job relies on the action's default. I believe it's better to either pin both or unpin both so the two jobs can't silently diverge.
```
      with:
        version: '31.0'
```

ViniciusCestarii (commented):
Nice PR! Everything worked here. Not a maintainer, just thinking out loud: could be worth considering as a future follow-up to wrap each call in a measure_ipc_call helper to expose per-method `ipc_fetch_duration` / `ipc_fetch_errors` on a local Prometheus endpoint, like the rpc extractor does (7dff361)?. Curious what you think @0xB10C, especially given the "experimental for a while" framing.

.github/workflows/ci.yml
xyzconstant: I just pinned both, thanks!

cde69e8 -> 6f23cdd ([compare](https://github.com/peer-observer/peer-observer/compare/cde69e8078c3a1778db8a924673d8807d3db3da5..6f23cdd03116acde0db59af29d46951f6d9f2449))

xyzconstant:
re: https://github.com/peer-observer/peer-observer/pull/379#pullrequestreview-4154925468

> Nice PR! Everything worked here. Not a maintainer, just thinking out loud: could be worth considering as a future follow-up to wrap each call in a measure_ipc_call helper to expose per-method `ipc_fetch_duration` / `ipc_fetch_errors` on a local Prometheus endpoint, like the rpc extractor does ([7dff361](https://github.com/peer-observer/peer-observer/commit/7dff361a55f521bce07738f362e0e9aae497d209))?. Curious what you think @0xB10C, especially given the "experimental for a while" framing.

Thanks! I think this makes sense, curious what @0xB10C thinks. Happy to follow up in a separate PR.

README.md
xyzconstant: Fixed!

6f23cdd03116acde0db59af29d46951f6d9f2449 -> dcd2716448ea317640f8ffb6e7e594d251757844 ([compare](https://github.com/peer-observer/peer-observer/compare/6f23cdd03116acde0db59af29d46951f6d9f2449..dcd2716448ea317640f8ffb6e7e594d251757844))

0xB10C:
@ViniciusCestarii thanks for testing! 


> ipc_fetch_duration / ipc_fetch_errors on a local Prometheus endpoint

yes, I think that's a good idea. Let's keep it as a follow up PR once this is merged.

xyzconstant:
Rebased and fixed conflicts with master

dcd2716448ea317640f8ffb6e7e594d251757844 -> b32d7f2587df9e3f33a3f70db7c63efdd7d5b810 ([compare](https://github.com/peer-observer/peer-observer/compare/dcd2716448ea317640f8ffb6e7e594d251757844..b32d7f2587df9e3f33a3f70db7c63efdd7d5b810))

0xB10C:
We are waiting for the backport in https://github.com/NixOS/nixpkgs/pull/512820 now until the Nix CI will work.

0xB10C:
I tried re-running the CI now that the backport has been merged, but seems like we need to wait a few more hours/days until it's in nixos-25.11. See https://nixpk.gs/pr-tracker.html?pr=512820

0xB10C:
I was wondering if this makes things easier for us @xyzconstant: https://github.com/rust-bitcoin/corepc/pull/572

xyzconstant:
re: https://github.com/peer-observer/peer-observer/pull/379#issuecomment-4350530623

Yeah it would definitely solve the `bitcoin-node` setup we have

xyzconstant:
I'm currently extending this branch to test against bitcoin/bitcoin#29409 `ChainNotifications`

https://github.com/xyzconstant/peer-observer/tree/ipc-extractor-over-%2329409

extractors/ipc/src/ipc.rs
```diff
@@ -0,0 +1,81 @@
...
+    pub async fn get_tip(&self) -> Result<BlockTip, RuntimeError> {
+        let mut req = self.mining.get_tip_request();
+        set_context(req.get().get_context()?, &self.thread);
+
+        let response = req.send().promise.await?;
+        let tip = response.get()?.get_result()?;
+
+        let height = tip.get_height();
+        let mut hash = tip.get_hash()?.to_vec();
+        hash.reverse();
```
deadmanoz: Other `bytes`-typed block hashes in the project store the natural-LE bytes on the wire as-is, that is, no `.reverse()` at the call site, and then format via `bitcoin::BlockHash::from_slice(&self.hash)`, which produces display-order hex on its own (the reversal lives in `BlockHash`'s `Display` impl, gated by `DISPLAY_BACKWARD = true` inherited from `sha256d::Hash`). `shared/src/protobuf/ebpf_extractor/validation.rs:11,26` is the reference shape: `connected.hash.to_vec()` for storage, `BlockHash::from_slice(&self.hash)` inside `Display`.

`tip.get_hash()` already arrives in natural LE, so the explicit `.reverse()` here makes `BlockTip.hash` the only `bytes`-typed block-hash field in the project carrying display-order bytes. A consumer treating `*.hash` uniformly via `BlockHash::from_slice` will mis-decode IPC events. Probably worth dropping the `.reverse()` and using `BlockHash::from_slice(&self.hash)` in `BlockTip`'s `Display` impl, mirroring `validation.rs:26`?

extractors/ipc/src/ipc.rs
```diff
@@ -0,0 +1,81 @@
...
+            mining,
+        })
+    }
+
+    pub async fn get_tip(&self) -> Result<BlockTip, RuntimeError> {
+        let mut req = self.mining.get_tip_request();
+        set_context(req.get().get_context()?, &self.thread);
+
+        let response = req.send().promise.await?;
+        let tip = response.get()?.get_result()?;
```
deadmanoz: `Mining.getTip` is optional in Core (`std::optional<BlockRef>`), and the IPC schema exposes that as `result` plus `hasResult` ([Core interface](https://github.com/bitcoin/bitcoin/blob/v31.0/src/interfaces/mining.h#L107-L108), [PR schema](https://github.com/peer-observer/peer-observer/blob/1e1546fedbe3a534365dc069f6372b5a75e8469a/extractors/ipc/capnp/mining.capnp#L20-L24)). This reads `get_result()` without checking `get_has_result()`, so it is relying on Core always having an active tip by the time this method is called. If that optional is empty, capnp-rust reads the absent result as defaults and this builds `BlockTip { height: 0, hash: [] }`.

I mean this could be fine? Alternatively, we could check `response.get()?.get_has_result()` before reading the result, and then skip publishing or return a clear error when Core reports no tip?

deadmanoz (changes requested):
I think that the `.reverse()` issue needs to be addressed!

extractors/ipc/src/ipc.rs
xyzconstant: Thanks for the review!

Just addressed this: 1e1546f -> 0e8a32a ([compare](https://github.com/peer-observer/peer-observer/compare/1e1546fedbe3a534365dc069f6372b5a75e8469a..0e8a32a337559a6eff8beac07710159b975e4558))

extractors/ipc/src/ipc.rs
xyzconstant: >If that optional is empty, capnp-rust reads the absent result as defaults and this builds BlockTip { height: 0, hash: [] }.

Good catch! In practice this would only happen in scenarios where `ActivateBestChain` hasn't run yet (i.e. node startup), under normal circumstances the tip is non-null. But yes, checking `hasResult` first for optional return types is the correct usage of `libmultiprocess` api regardless.

xyzconstant: Opted for skipping NATS publishing when there's no tip available. 

663c70f -> d45ec46 ([compare](https://github.com/peer-observer/peer-observer/compare/663c70f94c2faa308c87172ba79bc293e4f5f10f..d45ec4635aa617e974a88d7d692ee600039c7345))

Thanks for the feedback!

xyzconstant:
Rebased against master and addressed feedback

1e1546f-> 08a75bd ([compare](https://github.com/peer-observer/peer-observer/compare/1e1546fedbe3a534365dc069f6372b5a75e8469a..08a75bd0a9995cb47105f1f2bca3d01327da5a19))

0xB10C:
https://github.com/rust-bitcoin/corepc/pull/572 was merged! I can update the custom corepc branch we use to include the changes from there. This should allow us to drop the bitcoin-node binary extraction from our CI

0xB10C:
> [rust-bitcoin/corepc#572](https://github.com/rust-bitcoin/corepc/pull/572) was merged! I can update the custom corepc branch we use to include the changes from there. This should allow us to drop the bitcoin-node binary extraction from our CI

Done in https://github.com/0xB10C/corepc/tree/2026-05-corepc-0.12%2Bcustom-peer-observer-patches

feel free to upgrade this

https://github.com/peer-observer/peer-observer/blob/a8ab4559f0fc956f0d99515e0996a6fc12bfba16/shared/Cargo.toml#L25-L31

to (2026-04 -> 2026-05; and update the commit hash)

```toml
 # Use custom commit to support: 
 # - cpu_load and inv_to_send in getpeerinfo 
 # (branch 2026-05-corepc-0.12+custom-peer-observer-patches) 
 # When updating the Bitcoin Core version used, grep for CORE_VERSION_GREP 
 # and update the corepc_client::types::v* to the new version. 
 corepc-node = { git = "https://github.com/0xb10c/corepc", rev = "8798e7f9c240c626dc3aa61adadeaf42f8221e40", features = ["download", "30_2"] } 
 corepc-client = { git = "https://github.com/0xb10c/corepc", rev = "8798e7f9c240c626dc3aa61adadeaf42f8221e40", features = ["client-sync"]} 
```

and drop the CI scripts downloading and extracting the `bitcoin-node` binary. I think you might need to change the path too. Thanks!

xyzconstant:
re: https://github.com/peer-observer/peer-observer/pull/379#issuecomment-4387047963

Thanks you @0xB10C for the notice. I’d like to highlight something:

There's no support for Bitcoin Core v31.0 on corepc yet. I [forked and branched off](https://github.com/xyzconstant/corepc/tree/add-support-for-core-31.0) from master to work on adding such support (though [I already made some progress](https://github.com/xyzconstant/corepc/commit/b720056c6cdae7b9945e40adaa96c1275bb0af25) I think could be integrated into your patch right away)

Also they renamed `corepc-node` to `bitcoind`, this is not a big issue but just wanted to note it too.

EDIT: one of the issues was not that big, already resolved no need to open another PR.

0xB10C:
> There's no support for Bitcoin Core v31.0 on corepc yet. I [forked and branched off](https://github.com/xyzconstant/corepc/tree/add-support-for-core-31.0) from master to work on adding such support (though [I already made some progress](https://github.com/xyzconstant/corepc/commit/b720056c6cdae7b9945e40adaa96c1275bb0af25) I think could be integrated into your patch right away)

Right! I forgot about this. I think you can open this upstream at corepc - v31.0 support would surely be welcome there.

xyzconstant:
re: https://github.com/peer-observer/peer-observer/pull/379#issuecomment-4387047963
>to (2026-04 -> 2026-05; and update the commit hash)

Just opened #427 to handle this, so we keep this PR's scope thin and focused on the IPC extractor.

re: https://github.com/peer-observer/peer-observer/pull/379#issuecomment-4392104929
> Right! I forgot about this. I think you can open this upstream at corepc - v31.0 support would surely be welcome there.

Done at rust-bitcoin/corepc#586. I'd appreciate any feedback in there too :)

.github/actions/install-bitcoin/action.yml
```diff
@@ -0,0 +1,79 @@
...
+    required: false
+    # Signatures from the following GPG public keys checked during verification of the source code.
+    # The list can be found at https://github.com/bitcoin-core/guix.sigs/tree/main/builder-keys
+    # 15281230078...: achow101.gpg
+    # 9EDAFF80E08...: Emzy.gpg
+    # D1DBF2C4B96...: hebasto.gpg
+    default: >-
+      152812300785C96444D3334D17565732E08E5E41
+      9EDAFF80E080659604F4A76B2EBB056FD847F8A7
+      D1DBF2C4B96F2DEBF4C16654410108112E7EA81F
```
0xB10C, line 20: I didn't review this and this script too closely. My understanding is that we can drop it once https://github.com/rust-bitcoin/corepc/pull/586 is merged. Also, it's only used in CI.

.github/workflows/ci.yml
```diff
@@ -106,40 +106,48 @@ jobs:
...
+      with:
+        version: '31.0'
     - name: Install integration test dependecies
       run: sudo apt-get install -y nats-server
     - name: Run tests and integration tests
-      run: NATS_SERVER_BINARY=$(which nats-server) cargo test --all-features
+      run: NATS_SERVER_BINARY=$(which nats-server) BITCOIN_NODE_EXE=$(which bitcoin-node) cargo test --all-features
```
0xB10C, line 116: My understanding is that we can drop this once https://github.com/rust-bitcoin/corepc/pull/586 is merged. There are multiple occurences of this.

```suggestion
      run: NATS_SERVER_BINARY=$(which nats-server) cargo test --all-features
```

extractors/ipc/capnp/mining.capnp
```diff
@@ -0,0 +1,67 @@
...
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+@0xc77d03df6a41b505;
+
+using Cxx = import "/capnp/c++.capnp";
+$Cxx.namespace("ipc::capnp::messages");
+
+using Common = import "common.capnp";
+using Proxy = import "./mp/proxy.capnp";
```
0xB10C, line 11: There's a slight diff to https://raw.githubusercontent.com/bitcoin/bitcoin/refs/heads/31.x/src/ipc/capnp/mining.capnp here. Is this on purpose?

```diff
< using Proxy = import "./mp/proxy.capnp";
---
> using Proxy = import "/mp/proxy.capnp";
67c67
< }
\ No newline at end of file
---
> }
```

extractors/ipc/capnp/echo.capnp
```diff
@@ -0,0 +1,17 @@
+# Copyright (c) 2021-present The Bitcoin Core developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+@0x888b4f7f51e691f7;
+
+using Cxx = import "/capnp/c++.capnp";
```
0xB10C, line 7: similar diff here:

```diff
< using Proxy = import "./mp/proxy.capnp";
---
> using Proxy = import "/mp/proxy.capnp";
```

extractors/ipc/capnp/mining.capnp
0xB10C: Ah yeah I think I understand where this is comming from now!

extractors/ipc/src/lib.rs
```diff
@@ -0,0 +1,147 @@
...
+        net::UnixStream,
+        sync::watch,
+        time::{self, Duration},
+    },
+};
+use std::io;
+
+mod error;
+use error::RuntimeError;
+
```
0xB10C, line 23: nit
```suggestion
```

extractors/ipc/src/main.rs
```diff
@@ -0,0 +1,40 @@
+use ipc_extractor::Args;
+use shared::{
+    clap::Parser,
+    log, simple_logger,
+    tokio::{self, signal, sync::watch, task::LocalSet},
+};
+
+#[tokio::main(flavor = "current_thread")]
```
0xB10C: Any specific reason for `current_thread`?

extractors/ipc/build.rs
```diff
@@ -0,0 +1,18 @@
+fn main() -> Result<(), Box<dyn std::error::Error>> {
```
0xB10C, line 1: note to myself: this will require adding the `cptnproto` dependecies to the nix flake building peer-observer

extractors/ipc/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+name = "ipc-extractor"
+version = "0.1.0"
+edition = "2024"
+
+[dependencies]
+capnp = "0.25.1"
+capnp-rpc = "0.25.0"
+shared = { path = "../../shared" }
+tokio-util = { version = "0.7.18", features = ["compat"] }
```
0xB10C, line 10: `tokio-util` is probably Ok here, but if possible without adding a new dependecy, I'd favor dropping this.

extractors/ipc/README.md
```diff
@@ -0,0 +1,34 @@
+# `ipc` extractor
+
+> publishes data fetched from IPC
+
+The peer-observer ipc-extractor periodically queries data from the Bitcoin Core IPC interface and publishes the results as events into a NATS pub-sub queue
+
```
0xB10C, line 6: I think it would be nice to have instructions on how to run a Bitcoin Core node to in IPC mode here.

tools/logger/tests/integration.rs
```diff
@@ -120,7 +123,17 @@ async fn publish_and_check(events: &[Event], subject: Subject, expected: &str) {
...
+            nats_server.port,
+            true,
+            true,
+            true,
+            true,
+            true,
+            true,
+            true,
+            true,
+        );
```
0xB10C, line 136: not for this PR, but somewhere else we had a struct with default args for this. We should use this in `tools/logger/tests/integration.rs` too. The `true, true, true, true, true, true, true, true,` is ugly, requires updating on unrelated changes, and is very unreadable.

tools/websocket/src/lib.rs
```diff
@@ -254,6 +255,7 @@ async fn broadcast_to_clients(event: &PeerObserverEvent, clients: &Clients) {
             PeerObserverEvent::RpcExtractor(_) => client.subscriptions.rpc,
             PeerObserverEvent::P2pExtractor(_) => client.subscriptions.p2p,
             PeerObserverEvent::LogExtractor(_) => client.subscriptions.log,
+            PeerObserverEvent::IpcExtractor(_) => client.subscriptions.ipc,
```
0xB10C, line 258: We should check if this requires an update to the websocket frontend JS tools. I think it might work fine, but hopefully doesn't break it.

0xB10C (approved):
This looks good! I left a few comments that we might want to address (e.g. adding docs on how to run the bitcoin-node binary in the README for the ipc-extractor), but nothing that's really blocking this from being merged. All things we can follow up in small, targeted PRs.

I've also been contemplating to merge this before https://github.com/rust-bitcoin/corepc/pull/586 is merged. Once it's ready, we can upgrade corepc and drop the downloading and verification of Bitcoin Core from our CI. What do you think?

I've tested this with the instructions provided in the PR description! 

Thank you, high-quality work!

.github/workflows/ci.yml
0xB10C: Ah I don't think we can drop it.

0xB10C:
Follow-up ideas as GitHub isn't showing the full comment history:

- @ViniciusCestarii's comment on https://github.com/peer-observer/peer-observer/pull/379#pullrequestreview-4154925468: ipc_fetch_duration / ipc_fetch_errors on a local Prometheus endpoint is a good followup too.
- https://github.com/peer-observer/peer-observer/pull/379#discussion_r3265583406 docs on how to run the bitcoin-node binary and test the IPC extractor
- https://github.com/peer-observer/peer-observer/pull/379#discussion_r3265443316 cleanup when corepc supports v31.0
- https://github.com/peer-observer/peer-observer/pull/379#discussion_r3265604669 cleanup of true, true, true, true, ...
- https://github.com/peer-observer/peer-observer/pull/379#discussion_r3266253521: move `tokio-util` to shared. It's already a dependency for p2p-extractor
Probably good to go through this once merged and compile a list of ipc-extractor followups in an issue. Similar to #285.

tools/websocket/src/lib.rs
0xB10C: Tested this manually. Marking as resolved.

extractors/ipc/Cargo.toml
0xB10C: I just noticed that we already have `tokio-util` as a dependency for `extractors/p2p/Cargo.toml`. We should move it to shared and export it there!

Can be done in a follow-up.

.github/actions/install-bitcoin/action.yml
xyzconstant: This is based on https://github.com/NixOS/nixpkgs/blob/nixos-25.11/pkgs/applications/blockchains/bitcoin/default.nix

And yes, once v31 support on `corepc` lands we should definitely drop this action. I've already tested the new `bitcoin-node` binary locally and found no issue. Reverting the `capnp` schemas back to Core v30.2 so we can drop this action in this PR is an option too.

.github/workflows/ci.yml
xyzconstant: Yeah, we need it a separate variable for `bitcoin-node` path

extractors/ipc/capnp/mining.capnp
xyzconstant: Correct, this diff is needed to resolve locally `proxy.capnp` and not mess with prefix paths

extractors/ipc/src/main.rs
xyzconstant: Not really, this was inspired by https://github.com/bitcoin-dev-tools/ipc-exporter-rust/blob/master/src/main.rs#L189

xyzconstant:
Thanks for the review @0xB10C, I'll address it in a moment. I think we can drop the CI action as `bitcoin-node` is already available in `corepc`, the tradeoff is downgrading the `capnp` schemas to Core v30.2 which is the latest version available in `corepc`.

Then a smaller and properly scoped PR bumping to v31 can await for rust-bitcoin/corepc#586

xyzconstant:
Added instructions on how to run a `bitcoin-node` instance to `extractors/ipc/README.md`.

Working on rebasing and fixing the CI failures.

extractors/ipc/README.md
xyzconstant: Done!

extractors/ipc/src/main.rs
xyzconstant: Just removed `flavor = current_thread `

xyzconstant:
Rebased with master and fixed archiver tool errors in d71b6ba636051302f15c565153f282ab7efc25c9

xyzconstant:
Update:

nix-shell will resolve `bitcoind` to `v31.0` which breaks our `v30.2` capnp schemas, added an `IN_NIX_SHELL` guard to skip nix-shell IPC test until we bump support for v31.0.

b9c1410121a16180f4b076bea70fa25e0a0f641c ->  047b300d47373302f32d592d41974b8801b1c3d1 ([compare](https://github.com/peer-observer/peer-observer/compare/b9c1410121a16180f4b076bea70fa25e0a0f641c..047b300d47373302f32d592d41974b8801b1c3d1))

0xB10C:
> downgraded capnp schemas to v30.2

my nodes are all v31.0 by now so I won't be albe to test and deploy this until https://github.com/rust-bitcoin/corepc/pull/586 is merged 😄 I'll put that PR on my list to review. I don't think it's far from being merged.

Actually, merging this without v31.0 support will break my NixOS build, so I'll hold off for now.

One thing that we could thing about in a follow-up is "is it possible to support multiple versions of Bitcoin Core"?

0xB10C:
@xyzconstant told me out-of-band that he reverted back to v31.0 types for now. I just had another look, and this seems fine at 8028a5dc5e0105b15ee4279ac30d6aec7dbd3ec4

Thanks @xyzconstant - and sorry for the back and forth! I could have been clearer in https://github.com/peer-observer/peer-observer/pull/379#pullrequestreview-4317942170 that an approval and listing followups means "no-more-touching-unless-really-needed" and "I'm-going-to-merge-this-as-is"!

PR #380, opened 2026-03-13
refactor(ebpf-extractor): split binary into library crate with graceful shutdown

deadmanoz:
## Summary

The ebpf-extractor was the only extractor without a `lib.rs` / `main.rs` split: all logic lived in `main.rs`, which blocked integration test coverage for issue #26. This PR restructures it to match the rpc, log, and p2p extractors: `lib.rs` now exposes a public `Args` struct and `pub async fn run(args, shutdown_rx)`, while `main.rs` is reduced to argument parsing, logger initialization, shutdown setup, and calling `run()`.

Two implementation details intentionally diverge from the other extractors:

1. **`run()` is not spawned with `tokio::spawn`**  
The ebpf extractor's `run()` future is `!Send` because it holds libbpf state built on raw-pointer-backed FFI types. `tokio::spawn` requires `Send`, since spawned tasks may move between runtime worker threads. As a result, `run()` executes directly on the main task and only the Ctrl+C handler is spawned.

2. **Shutdown uses synchronous `has_changed()` instead of `tokio::select!`**  
The main extraction loop calls `ring_buffers.poll_raw(Duration::from_secs(1))`, which is a blocking FFI call rather than an async `.await` point. A non-blocking shutdown check before each poll iteration is therefore the natural fit, with up to ~1 second shutdown latency. Dropped senders are treated as shutdown, matching the `Err(_) => break` behavior used by the other extractors.

The work is split into three commits:
1. pure code movement into `lib.rs`
2. shutdown channel support
3. logger-init alignment with the other extractors

## Testing

Verified locally (NixOS x86_64):

- **clippy**: zero warnings
- **Release build**: clean
- **`--help`**: CLI interface unchanged
- **Bogus PID invocation without required privileges**: graceful error through `run()`, for example:

```
$ target/release/ebpf-extractor --bitcoind-path /usr/bin/bitcoind --bitcoind-pid 99999
INFO  [ebpf_extractor] Using bitcoind PID=99999 specified via command line option
INFO  [ebpf_extractor] Opening BPF skeleton with debug=false..
INFO  [ebpf_extractor] Loading BPF functions and maps into kernel..
ERROR [ebpf_extractor] Fatal error during extractor runtime: libbpf error: Operation not permitted (os error 1)
```

No integration tests added yet.. though this PR provides the structural prerequisite for integration tests to follow!


0xB10C:
Reviewed and tested!

Thank you @deadmanoz!

Issue #185, opened 2025-07-11
Implement `alerts` tool that logs when a heuristic is triggered


0xB10C:
While the `metrics` tool (and to some extend the other tools) help searching for unknown unknowns, the `alerts` tool should use predefined heuristics to alert once that heuristic is triggered.

A simple example might be: Once a peer is sending more than 10 pings per minute, consider it a ping-spammer and alert.

An alert could work in different and multiple ways:
- just log a human readable message with a timestamp to a file and stdout; use some other tool to visualize this. Maybe even show these on the peer-observer infra website somewhere
- increment some Prometheus metrics counter that then might trigger an alert in e.g. Grafana
- push some data or text to a websocket. For example, send a message into a Discord channel
- ...

Some alerts that come to mind:
- [ ] ping-spammer: alerts when a peer sends more than N pings per minute (N=10?)
- [ ] addr-spammer: alerts when a peer sends more than N addr / addrv2 entries per minute (N = ?; match with Bitcoin Core rate limiting?)
- [ ] inbound-connection-drop: alerts when over a time frame of N minutes more than X inbound connections drop (N = 5?; X=10?). Maybe ignore evicted connections as these can happen frequently when inbound slots are full. Need decided if the metric to look at is absolute inbound connection count or if just looking at "X connections dropped in N minutes" is more helpful.
- [ ] old-connections dropping: alerts when N old in- or outbound connections drop. A connection is considered old when it's older than X (N = 5? X = 24h?)
- [ ] detect spy peers: Similar to #43: when a peer is only listening to our invs, but never requesting transactions or INVing something to us, they are probably a spy node. Could alert about these here too
- [ ] block (height, hash) processed by a node slower than usual. Record on stddev and alert on high Z-scores (e.g., greater than 2 or 3 standard deviations).
- [ ] block processing takes longer than N  seconds (with N being 60 seconds?)
- [ ] alert on stale tip: i.e. no blocks processed in N minutes (120 or 180 minutes)
- [ ] many peers M with a ping_await higher than Y seconds (see https://github.com/0xB10C/peer-observer/issues/198#issuecomment-3151179883)
- [ ] A peer advertises that it relays transactions to us, is connected for a while X (>1 min), but never advertised a transaction to us
- [ ] track our ping response timings to peers (they send a ping and we respond with a pong) and alert if it is over X ms on average (see also #212)
- [ ] if we have more than X peers, alert if single peer makes up for more than Y% of the traffic
[...]

PR #383, opened 2026-03-16
Alerts tool

GuiSchet:
This PR implements the initial `tools/alerts` crate, the first step towards resolving [#185 — Implement `alerts` tool that logs when a heuristic is triggered](https://github.com/peer-observer/peer-observer/issues/185).

The `alerts` tool subscribes to NATS events published by `ebpf-extractor` and applies sliding-window counters to detect peers that send anomalously high rates of specific P2P message types. Alerts are emitted at `WARN` level with a structured `[ALERT]` prefix so they can be grepped, piped, or forwarded to any log-aggregation system.

---

### PingSpammer
Fires when a peer sends more than `--ping-threshold` inbound `ping` messages within a `--ping-window-secs` sliding window.

**Default:** `> 3 pings / 120 s`

### AddrSpammer
Fires when a peer sends more than `--addr-threshold` inbound `addr` or `addrv2` messages within a `--addr-window-secs` sliding window. Both message types share one counter.

**Default:** `> 6 addr/addrv2 messages / 60 s`

## Other features

| Feature | Detail |
|---|---|
| One-shot flag | Each peer/type pair fires at most one alert per session |
| Outbound ignored | Only inbound messages are counted |
| SpammerGone alert | Emitted on disconnect/stale-eviction of a flagged peer; logs spamming duration |
| Connection cleanup | Per-peer state is freed when a `ConnectionEvent::Close` is received |
| Inactivity cleanup | --peer-stale-secs (default 300 s) evicts idle peer state to prevent memory leaks |
| All thresholds are CLI flags | No recompilation required when tuning sensitivity |

### DIFF
--- a/tools/alerts/README.md
+++ b/tools/alerts/README.md
@@ -1,0 +1,100 @@
+# alerts
+
+A peer-observer tool that subscribes to NATS events and emits alerts when anomalous
+peer behavior is detected.
+
+## Heuristics
+
+### PingSpammer
+
+Detects peers that send more than `--ping-threshold` inbound `ping` messages within a
+`--ping-window-secs` second sliding window.
+
+Alert format (as rendered by `LoggingAlerter`):
+```
+PingSpammer | peer_id=<id> addr=<addr> | <count> pings in last <secs>s (threshold: <threshold>)
+```
+
+### AddrSpammer
+
+Detects peers that send more than `--addr-threshold` inbound `addr` or `addrv2` messages
+within an `--addr-window-secs` second sliding window. Both message types are counted in
+the same window.
+
+Alert format (as rendered by `LoggingAlerter`):
+```
+AddrSpammer | peer_id=<id> addr=<addr> | <count> addr/addrv2 messages in last <secs>s (threshold: <threshold>)
+```
+
+Each peer/kind combination fires at most one alert per session (one-shot).
+
+## Peer lifecycle
+
+When a previously flagged peer disconnects (either via a `ConnectionEvent::Close`
+or by being evicted after `--peer-stale-secs` of inactivity), the same `Alerter`
+emits a `PeerDisconnected` alert with the duration of the spamming episode:
+
+```
+PeerDisconnected | peer_id=<id> addr=<addr> | active=<duration>s
+```
+
+Non-flagged peers disconnect silently (no alert is emitted).
+
+## Usage
+
+```
+cargo run -p alerts -- [OPTIONS]
+```
+
+## Options
+
+```
+Arguments for the connection the the NATS server that each extractor and tool needs
+
+Usage: alerts [OPTIONS]
+
+Options:
+  -a, --nats-address <ADDRESS>
+          The NATS server address the extractor/tool should connect and subscribe to [default: 127.0.0.1:4222]
+  -u, --nats-username <USERNAME>
+          The NATS username the extractor/tool should try to authentificate to the NATS server with
+  -p, --nats-password <PASSWORD>
+          The NATS password the extractor/tool should try to authentificate to the NATS server with
+  -f, --nats-password-file <PASSWORD_FILE>
+          A path to a file containing a password the extractor/tool should try to authentificate to the NATS server with
+      --ping-threshold <PING_THRESHOLD>
+          Number of pings in the window before alerting [default: 3]
+      --ping-window-secs <PING_WINDOW_SECS>
+          Sliding window size for ping detection (seconds) [default: 120]
+      --addr-threshold <ADDR_THRESHOLD>
+          Number of addr/addrv2 messages in the window before alerting [default: 6]
+      --addr-window-secs <ADDR_WINDOW_SECS>
+          Sliding window size for addr detection (seconds) [default: 60]
+  -l, --log-level <LOG_LEVEL>
+          [default: DEBUG]
+      --peer-stale-secs <PEER_STALE_SECS>
+          Remove peers not seen for this many seconds (prevents OOM on missed Close events) [default: 300]
+  -h, --help
+          Print help
+  -V, --version
+          Print version
+```
+
+## Example
+
+```bash
+# Use low thresholds for testing
+cargo run -p alerts -- \
+  --nats-address nats://localhost:4222 \
+  --ping-threshold 3 \
+  --ping-window-secs 30 \
+  --addr-threshold 2 \
+  --addr-window-secs 60
+```
+
+Example output:
+```
+INFO  [alerts::alerter] PingSpammer | peer_id=42 addr=1.2.3.4:8333 | 4 pings in last 30s (threshold: 3)
+INFO  [alerts::alerter] AddrSpammer | peer_id=7 addr=5.6.7.8:8333 | 3 addr/addrv2 messages in last 60s (threshold: 2)
+INFO  [alerts::alerter] PeerDisconnected | peer_id=42 addr=1.2.3.4:8333 | active=47s
+```
--- a/tools/alerts/src/error.rs
+++ b/tools/alerts/src/error.rs
@@ -1,0 +1,50 @@
+use shared::async_nats;
+use shared::async_nats::ConnectError;
+use std::error;
+use std::fmt;
+use std::io;
+
+#[derive(Debug)]
+pub enum RuntimeError {
+    NatsSubscribe(async_nats::client::SubscribeError),
+    NatsConnect(ConnectError),
+    Io(io::Error),
+}
+
+impl fmt::Display for RuntimeError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            RuntimeError::NatsSubscribe(e) => write!(f, "NATS subscribe error {}", e),
+            RuntimeError::NatsConnect(e) => write!(f, "NATS connection error {}", e),
+            RuntimeError::Io(e) => write!(f, "IO error {}", e),
+        }
+    }
+}
+
+impl error::Error for RuntimeError {
+    fn source(&self) -> Option<&(dyn error::Error + 'static)> {
+        match *self {
+            RuntimeError::NatsSubscribe(ref e) => Some(e),
+            RuntimeError::NatsConnect(ref e) => Some(e),
+            RuntimeError::Io(ref e) => Some(e),
+        }
+    }
+}
+
+impl From<async_nats::client::SubscribeError> for RuntimeError {
+    fn from(e: async_nats::client::SubscribeError) -> Self {
+        RuntimeError::NatsSubscribe(e)
+    }
+}
+
+impl From<ConnectError> for RuntimeError {
+    fn from(e: ConnectError) -> Self {
+        RuntimeError::NatsConnect(e)
+    }
+}
+
+impl From<io::Error> for RuntimeError {
+    fn from(e: io::Error) -> Self {
+        RuntimeError::Io(e)
+    }
+}
--- a/tools/alerts/src/lib.rs
+++ b/tools/alerts/src/lib.rs
@@ -1,0 +1,470 @@
+#![cfg_attr(feature = "strict", deny(warnings))]
+
+use shared::clap;
+use shared::clap::Parser;
+use shared::futures::stream::StreamExt;
+use shared::log;
+use shared::nats_subjects::Subject;
+use shared::nats_util;
+use shared::nats_util::NatsArgs;
+use shared::prost::Message;
+use shared::protobuf::ebpf_extractor::connection::connection_event;
+use shared::protobuf::ebpf_extractor::ebpf;
+use shared::protobuf::ebpf_extractor::message::message_event::Msg;
+use shared::protobuf::event::Event;
+use shared::tokio::sync::watch;
+use std::collections::{HashMap, VecDeque};
+use std::time::{Duration, Instant};
+
+use crate::error::RuntimeError;
+
+pub mod alerter;
+pub mod error;
+
+pub use crate::alerter::{Alert, Alerter, LoggingAlerter, SpammerKind};
+
+#[derive(Parser, Debug, Clone)]
+#[command(version, about, long_about = None)]
+pub struct Args {
+    /// Arguments for the connection to the NATS server
+    #[command(flatten)]
+    pub nats: NatsArgs,
+
+    /// Number of pings in the window before alerting
+    #[arg(long, default_value_t = 3)]
+    pub ping_threshold: usize,
+
+    /// Sliding window size for ping detection (seconds)
+    #[arg(long, default_value_t = 120)]
+    pub ping_window_secs: u64,
+
+    /// Number of addr/addrv2 messages in the window before alerting
+    #[arg(long, default_value_t = 6)]
+    pub addr_threshold: usize,
+
+    /// Sliding window size for addr detection (seconds)
+    #[arg(long, default_value_t = 60)]
+    pub addr_window_secs: u64,
+
+    #[arg(short, long, default_value_t = shared::log::Level::Debug)]
+    pub log_level: shared::log::Level,
+
+    /// Remove peers not seen for this many seconds (prevents OOM on missed Close events)
+    #[arg(long, default_value_t = 300)]
+    pub peer_stale_secs: u64,
+}
+
+#[derive(Default)]
+struct AlertTypeState {
+    timestamps: VecDeque<Instant>,
+    alerted_at: Option<Instant>,
+}
+
+struct PeerState {
+    ping: AlertTypeState,
+    addr: AlertTypeState,
+    last_seen: Instant,
+    peer_addr: String,
+}
+
+impl PeerState {
+    fn new(peer_addr: String, now: Instant) -> Self {
+        PeerState {
+            ping: AlertTypeState::default(),
+            addr: AlertTypeState::default(),
+            last_seen: now,
+            peer_addr,
+        }
+    }
+
+    /// Records a hit for `kind` and returns an `Alert` the first time the
+    /// threshold is exceeded. After alerting once, subsequent hits are dropped
+    /// and the function returns `None` without growing the sliding window
+    fn check(
+        &mut self,
+        kind: SpammerKind,
+        peer_id: u64,
+        now: Instant,
+        args: &Args,
+    ) -> Option<Alert> {
+        let (state, window_secs, threshold) = match kind {
+            SpammerKind::Ping => (&mut self.ping, args.ping_window_secs, args.ping_threshold),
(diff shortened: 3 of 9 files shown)


0xB10C:
Cool! A few thoughts, but I haven't looked at the code much:

> --alert-cooldown-secs (default 120 s) prevents alert storms for the same peer/type

We could also mark a peer as ping/addr spammer once it crosses a threshold with e.g. a boolean. Not sure if it's worth it to retrigger an alert. I think you can probably remove `--alert-cooldown-secs` then.

> Per-peer state is freed when a ConnectionEvent::Close is received

Just to be sure, we should also remove peers we didn't hear anything from in a while (e.g. 5 minutes) and remove peers that aren't in an RPC event getpeerinfo anymore. Otherwise, if we miss the `ConnectionEvent::Close` event, we might never clear it and OOM at some point.

> --ignore-local (default true) silences loopback

This might be useful for silencing the peer-observer `p2p-extractor`, but inbound Tor peers also show up as connecting via loopback (as that's where the torproxy sits). I think it's fine to mark the p2p-extractor as a ping-spammer (as that's what it does). If we don't re-alert without `--alert-cooldown-secs`, I think `--ignore-local` might not be needed.

GuiSchet:
Hi b10, I'll draft the PR to review it, thanks!
About the Cooldown: If we use a boolean and remove the flag, we won't know how long the peer keeps attacking us. I thought that info could be useful ('ll think how to maintain the time without spamming messages).
Remove peer: You are right. I will add an inactivity cleanup to avoid OOM.
Ignore local: I understand the Tor issue now. I will remove the flag.

GuiSchet:
I've changed:                                                                                                                   
                                                                                                                                                                       
  - Cooldown → one-shot flag: Replaced `--alert-cooldown-secs` with a per-peer/per-type boolean. Once a peer crosses the threshold, a single alert fires and the flag is set — no repeated alerts, no extra CLI param.                                                                                          
  - `--ignore-local` remove: Dropped the flag entirely.                                                                   
  - Inactivity cleanup: Peers not seen for `--peer-stale-secs` (default 300s) are evicted from state, so a missed `ConnectionEvent::Close` won't leak memory indefinitely.                                                                                                                                          
  - `SpammerGone` alert: when a flagged peer disconnects, a `SpammerGone` log line is emitted with the total duration of the episode.      
  
SpammerGone is logged at WARN like the other alerts, but it could make sense to use INFO (or a separate [RESOLVED] prefix) to make it easier to filter active vs. resolved alerts, leaving that as a future consideration.

0xB10C:
Nice, feel free to mark this as ready for review again once I should have a look.

Played around with this briefly and it seems to work: `[alerts] [ALERT] PingSpammer | peer_id=1 addr=127.0.0.1:28213 | 4 pings in last 120s (threshold: 3)`!

tools/alerts/src/lib.rs
```diff
@@ -0,0 +1,340 @@
...
+        state.alerted_at = Some(now);
+        Some(state.timestamps.len())
+    } else {
+        None
+    }
+}
+
+/// Removes a peer from state. If the peer was flagged as a spammer, emits a
+/// SpammerGone alert with the total duration of the spamming episode.
+fn remove_peer(peer_id: u64, peer: PeerState, now: Instant) {
```
RazorBest: nit: Only second part of the documentation is true. This function doesn't remove a peer.

tools/alerts/src/lib.rs
```diff
@@ -0,0 +1,340 @@
...
+                        break;
+                    }
+                }
+            }
+        }
+    }
+    Ok(())
+}
+
+/// Routes a NATS event to the appropriate handler, then prunes stale peers
```
RazorBest: nit: I'd remove the "then prunes stale peers" part, since this is function doesn't do this.

tools/alerts/src/lib.rs
```diff
@@ -0,0 +1,340 @@
...
+    state: &mut AlertState,
+    args: &Args,
+    now: Instant,
+) {
+    if !msg.meta.inbound {
+        return;
+    }
+
+    let peer_id = msg.meta.peer_id;
+    let addr = msg.meta.addr.clone();
```
RazorBest:
```suggestion
    let addr = &msg.meta.addr;
```

nit

RazorBest (approved):
Looks good! Only nits.

The tool cannot not be easily DOS-ed (as you periodically remove from the HashMap the entries that are outside of the lookup window), which I think is good.

Ran the integration tests successfully,

Manually tested with p2p_extractor, with:
- 1 ping every second
- 1 ping + 1 addr every second
- 1 ping + 1 addrv2 every second

Output from `cargo run --release` (as expected):
```
2026-04-09T02:12:03.953Z INFO  [alerts] ping_threshold:       3
2026-04-09T02:12:03.953Z INFO  [alerts] ping_window_secs:     120
2026-04-09T02:12:03.953Z INFO  [alerts] addr_threshold:       6
2026-04-09T02:12:03.953Z INFO  [alerts] addr_window_secs:     60
2026-04-09T02:12:03.953Z INFO  [alerts] peer_stale_secs:      300
2026-04-09T02:12:03.953Z DEBUG [shared::nats_util] Connecting to NATS-server at 127.0.0.1:4222 without authentification
2026-04-09T02:12:03.956Z INFO  [alerts] Connected to NATS-server at 127.0.0.1:4222
2026-04-09T02:17:29.553Z WARN  [alerts] [ALERT] PingSpammer | peer_id=0 addr=127.0.0.1:9555 | 4 pings in last 120s (threshold: 3)
2026-04-09T02:18:16.828Z WARN  [alerts] [ALERT] SpammerGone | peer_id=0 addr=127.0.0.1:9555 | active for 47s
2026-04-09T02:24:22.160Z WARN  [alerts] [ALERT] PingSpammer | peer_id=0 addr=127.0.0.1:9555 | 4 pings in last 120s (threshold: 3)
2026-04-09T02:24:24.752Z WARN  [alerts] [ALERT] AddrSpammer | peer_id=0 addr=127.0.0.1:9555 | 7 addr/addrv2 messages in last 60s (threshold: 6)
2026-04-09T02:24:32.327Z WARN  [alerts] [ALERT] SpammerGone | peer_id=0 addr=127.0.0.1:9555 | active for 10s
2026-04-09T02:26:49.860Z WARN  [alerts] [ALERT] PingSpammer | peer_id=0 addr=127.0.0.1:9555 | 4 pings in last 120s (threshold: 3)
2026-04-09T02:26:51.541Z WARN  [alerts] [ALERT] AddrSpammer | peer_id=0 addr=127.0.0.1:9555 | 7 addr/addrv2 messages in last 60s (threshold: 6)
2026-04-09T02:27:00.243Z WARN  [alerts] [ALERT] SpammerGone | peer_id=0 addr=127.0.0.1:9555 | active for 10s
```

GuiSchet:
ey @RazorBest, thanks for the feedback! I've applied the suggested changes

tools/alerts/src/lib.rs
RazorBest: I can't resolve these comments, so I'm just gonna reply "Done" to them.

Done

tools/alerts/src/lib.rs
RazorBest: Done

tools/alerts/src/lib.rs
RazorBest: Done

tools/alerts/src/error.rs
```diff
@@ -0,0 +1,70 @@
+use shared::async_nats;
+use shared::async_nats::ConnectErrorKind;
```
RazorBest: nit: I've recently learned that the two lines above can be written like this, in one line:
```rust
use shared::async_nats::{self, ConnectErrorKind};
```

tools/alerts/src/error.rs
```diff
@@ -0,0 +1,70 @@
...
+use std::error;
+use std::fmt;
+use std::io;
+
+#[derive(Debug)]
+pub enum RuntimeError {
+    SetLogger(SetLoggerError),
+    ProtobufDecode(DecodeError),
+    NatsSubscribe(async_nats::client::SubscribeError),
+    NatsConnect(shared::async_nats::error::Error<ConnectErrorKind>),
```
RazorBest: nit: you can use `shared::async_nats::ConnectError`

https://docs.rs/async-nats/latest/async_nats/type.ConnectError.html

RazorBest: The fix to this makes the previous comment stale

tools/alerts/tests/integration.rs
```diff
@@ -0,0 +1,555 @@
...
+
+    assert!(
+        logs_contain("[ALERT] PingSpammer"),
+        "expected PingSpammer alert in logs, got: {:?}",
+        get_logged_messages()
+    );
+
+    shutdown_tx.send(true).unwrap();
+    alerts_handle.await.unwrap().unwrap();
+}
```
RazorBest, line 229: nit: one principle I apply to testing is to also test things that are not happening. In this case, I would publish 3 pings, check that the logs do not contain an alert. Then, send the fourth ping, and do the final assert.

tools/alerts/tests/integration.rs
```diff
@@ -0,0 +1,555 @@
...
+                RuntimeError::NatsConnect(_) => (),
+                _ => panic!("Expected NatsConnect error since the NATS server isn't reachable."),
+            },
+        };
+    });
+
+    alerts_handle.await.unwrap();
+}
+
+/// Publishes a `ClosedConnection` event and verifies the alerts tool shuts down
```
RazorBest: nit: "and verifies the alerts tool shuts down" - I don't think the test verifies this.

RazorBest (approved):

RazorBest:
Btw, the above comments are nits, because I don't want them to be a blocker for merging. You can choose to ignore them.

tools/alerts/src/lib.rs
```diff
@@ -0,0 +1,340 @@
...
+        (None, Some(a)) => Some(a),
+        (None, None) => return,
+    };
+    if let Some(t) = first_alerted {
+        log::warn!(
+            "[ALERT] SpammerGone | peer_id={} addr={} | active for {}s",
+            peer_id,
+            peer.peer_addr,
+            now.duration_since(t).as_secs(),
+        );
```
0xB10C: - I think this can be a normal info! logging. It's expected behavior that we see someone disconnecting after a while.
- I don't think we need to label it `[ALERT]` if someone disconnects
- I'm not sure if I'd pick `SpammerGone` to describe a peer disconnected :)

```suggestion
        log::info!(
            "Peer {} disconnected (addr={}, active={}s)",
            peer_id,
            peer.peer_addr,
            now.duration_since(t).as_secs(),
        );
```

tools/alerts/src/lib.rs
```diff
@@ -0,0 +1,340 @@
...
+
+/// Sliding-window counter with one-shot alert
+/// Returns Some(count) the first time the threshold is exceeded
+fn check_spammer(
+    state: &mut AlertTypeState,
+    now: Instant,
+    window: Duration,
+    threshold: usize,
+) -> Option<usize> {
+    state.timestamps.push_back(now);
```
0xB10C: I think we can return here early if we've alerted before. No need to keep track of this.

```suggestion
) -> Option<usize> {
    if state.alerted {
       return None
    }

    state.timestamps.push_back(now);
```

tools/alerts/src/lib.rs
```diff
@@ -0,0 +1,340 @@
...
+                    Alert {
+                        kind: AlertKind::AddrSpammer,
+                        peer_id,
+                        addr,
+                        count,
+                        window_secs: args.addr_window_secs,
+                        threshold: args.addr_threshold
+                    }
+                );
+            }
```
0xB10C:
```suggestion
                log::info!(
                    "[ALERT] {}",
                    Alert {
                        kind: AlertKind::AddrSpammer,
                        peer_id,
                        addr,
                        count,
                        window_secs: args.addr_window_secs,
                        threshold: args.addr_threshold
                    }
                );
            }
```

tools/alerts/src/lib.rs
```diff
@@ -0,0 +1,340 @@
...
+                        threshold: args.ping_threshold
+                    }
+                );
+            }
+        }
+        Some(Msg::Addr(_)) | Some(Msg::Addrv2(_)) => {
+            let window = Duration::from_secs(args.addr_window_secs);
+            if let Some(count) =
+                check_spammer(&mut peer.addr_spam, now, window, args.addr_threshold)
+            {
```
0xB10C: Two things to note here about the design:

1. I think it would be better to have a `check()` function on the `PeerState`. Pass `AlertKind::AddrSpammer` to it for it to known which timings to use.
2. `check` could return `Option<Alert>` instead of a count.

I think this would make it a lot cleaner

tools/alerts/tests/integration.rs
0xB10C: I hope the tool doesn't shut down when a connection is closed! 😄

tools/alerts/README.md
```diff
@@ -0,0 +1,98 @@
+# alerts
+
+A peer-observer tool that subscribes to NATS events and emits alerts to stdout when
+anomalous peer behavior is detected. Alerts are logged at the `WARN` level with the
+prefix `[ALERT]`.
```
0xB10C: As mentioned above: I think info logging is enough. Also not sure about the `[ALERT]`

0xB10C:
Left a few code comments above. Some more general comments here:

- Your PR contains a commit `Merge branch 'peer-observer:master' into alerts-tool` - I don't think that's needed. Generally, I think you can squash a bunch of the commits. Maybe one clean commit adding the alerts tool and one adding tests.
- I'm not sure about the internal design. I mentioned some changes I'd make in https://github.com/peer-observer/peer-observer/pull/383#discussion_r3091581386 and I think we should generalize this more and allow easy addition of other alert channels in the future. While a textual log is a good start, it's probably not the main way the tool is going to be used in the future. I recommended `"You can build it in a way that allows us to hook up more ways to publish these alerts in the future."` in https://github.com/peer-observer/peer-observer/pull/383#discussion_r3091581386 - one way of archiving this be to have a general "Alerter" trait and you'd implement e.g. a `LoggingAlerter` for it in this PR. A future PR could than add `TextfileAlerter` that e.g. writes to a file on disk. And futher in the future a `WebhookAlerter`, `NATSAlerter`, `PrometheusAlerter`, ...
- I think you could also simplify the tests by not having to read and match the logs. I think a `IntegrationTestAlerter` could be helpful here. It might just give you a channel where it sends alerts to the test.
- Additionally, I think we can cover some of the AlertState functions with unit tests. The integration test can verify that receiving via NATS works, but we can test the `check()` (or `check_spammer()`) functions with a unit test.
- I just saw that I recommened `log::warn!` in https://github.com/peer-observer/peer-observer/issues/185#issuecomment-3853094523 and now recommend `log:info!` - sorry! I think `info!` is a better fit here.


At this point, I'm not sure how to best proceed here. I have a pretty clear idea of how it could look like, and I'd be happy to use your basis and implement it. On the other hand, if you want to do it, feel free to do it.

GuiSchet:
Thanks for the detailed review! I agree that the Alerter trait approach makes the tool much more extensible.
 I'll give it a shot following your suggestions.

GuiSchet:
Refactored on feedback from @0xB10C and @RazorBest (thanks for the help with the tests!):

- Introduced an `Alerter` trait. This PR ships `LoggingAlerter`, future channels can be added without touching detection logic.
- Tests now use a channel-based `IntegrationTestAlerter` instead of log-string matching.
- Switched alert level from `warn!` to `info!`.
- Squashed history into two commits (`feat:` + `test:`) and removed the merge commit.

Ready for another look when you guys have time 🙏

tools/alerts/src/error.rs
```diff
@@ -0,0 +1,50 @@
+use shared::async_nats;
```
RazorBest, line 1: Just a heads-up that this PR exists: https://github.com/peer-observer/peer-observer/pull/418

Which gets rid of the immediate necessity of creating custom error structs whenever you add a new tool, by replacing all the errors with `anyhow::Error`.

RazorBest (commented):
Hi! Just a comment for now. I plan to come back again for this.

tools/alerts/src/error.rs
GuiSchet: yep I'm following the PR! 'll update that when that PR is closed.

0xB10C:
A few things we should keep in mind when adding more Alerters in the future:

-  `emit` must not block for e.g. Websockets or HTTP requests. Added a comment for this.
-  When we want to fan-out to multiple sinks, which is not implemented yet as we only have one Alerter, we could use a small `CompositeAlerter(Vec<Box<dyn Alerter>>)` that implements `Alerter` and dispatches to each. We don't want to have multiple NATS subscriptions at a time

tools/alerts/tests/integration.rs
```diff
@@ -0,0 +1,498 @@
...
+    /// `mk_args` receives the NATS port and returns the `Args` to pass to `run`
+    /// so each test can pin its own thresholds while reusing the harness setup
+    async fn new(mk_args: impl FnOnce(u16) -> Args) -> Self {
+        let server = NatsServerForTesting::new(&[]).await;
+        let publisher = NatsPublisherForTesting::new(server.port).await;
+        let (shutdown_tx, shutdown_rx) = watch::channel(false);
+        let (alerter, rx) = IntegrationTestAlerter::new();
+        let args = mk_args(server.port);
+        let handle = tokio::spawn(alerts::run(args, alerter, shutdown_rx));
+        sleep(Duration::from_secs(1)).await;
```
0xB10C, line 187: I feel like this might cause a #405 going forward. I'm tending to leave as is and merge with this in (and fix it later)

tools/alerts/tests/integration.rs
```diff
@@ -0,0 +1,498 @@
...
+                peer_id,
+                addr: addr.to_string(),
+                conn_type: 1,
+                command: "addr".to_string(),
+                inbound: true,
+                size: 30,
+            },
+            msg: Some(Msg::Addr(Addr { addresses: vec![] })),
+        })),
+    }))
```
0xB10C, line 124: I noticed that we need these kind of test events in a bunch of places by now. Probably good to create helpers for this in `shared`/`testing` at some point.

0xB10C (commented):
I think this is probably fine to merge, thank you @GuiSchet! We can run this a bit on the demo nodes and see what good next steps are. We probably want to implement a few other alerts nex from #185. Not sure yet what the next alerter would be.

tools/alerts/tests/integration.rs
GuiSchet: good idea!

Issue #356, opened 2026-02-12
rpc-extractor: incorrect min feerate calculation


0xB10C:
After picking up the changes from https://github.com/rust-bitcoin/corepc/pull/483 in https://github.com/peer-observer/peer-observer/pull/351, the min feerate seems to be reported incorrectly.

<img width="996" height="940" alt="Image" src="https://github.com/user-attachments/assets/4f56b4ee-5b1f-4637-a1e4-4bbb4e8abe65" />

Likely, our implementation of `to_sat_per_vb_f64` is now wrong.

https://github.com/peer-observer/peer-observer/blob/d9b0ff3174d9c2936b3c938daee7e94d367d0281/shared/src/protobuf/rpc_extractor.rs#L23-L31

0xB10C:
Fixed by @m4ycon as part of https://github.com/peer-observer/peer-observer/pull/386

PR #386, opened 2026-03-17
add(rpc-extractor, metrics): estimatesmartfee metric

### LINKED ISSUE #362: metrics-tool: Observe estimatesmartfee results overtime
I think that would be very interesting to observe how estimatesmartfee perform comparing to the results from confirmed transactions. I had a look at Ismael Sadeeq studies at this matter and as I know peer-observer, thought that it would fit very well here.

I took a look if it wasn't a data already being extracted but it doesn't seem so. Problably this would involve `rpc_extractor` and `metrics`, collecting results from `estimatesmartfee` (both modes) and last block fees (don't know about this one, but I suppose it's possible via rpc).

What do you think @0xB10C ?

Refs:
- https://delvingbitcoin.org/t/mempool-based-fee-estimation-on-bitcoin-core/703
- https://delvingbitcoin.org/t/empirical-data-analysis-of-fee-rate-forecasters-for-asap-next-block-fee-estimation/1022
    - https://github.com/bitcoin/bitcoin/pull/34075
- https://delvingbitcoin.org/t/bitcoind-policy-estimator-modes-analysis/964

m4ycon:
Add `estimatesmartfee` calls that watch how the feerate values move for the combinations of target (1, 6, 144) and mode (economical, conservative).

This PR touches rpc-extractor and metrics tool.

Image

closes #362

### DIFF
--- a/extractors/rpc/src/lib.rs
+++ b/extractors/rpc/src/lib.rs
@@ -1,6 +1,7 @@
 use shared::clap::{ArgGroup, Parser};
 use shared::corepc_client::client_sync::Auth;
 use shared::corepc_client::client_sync::v30::Client;
+use shared::corepc_client::types::v30::EstimateSmartFee as RPCEstimateSmartFee;
 use shared::corepc_node::mtype::{
     GetBlockchainInfo, GetChainTxStats, GetNetworkInfo, GetOrphanTxsVerboseTwo,
 };
@@ -9,7 +10,7 @@ use shared::nats_subjects::Subject;
 use shared::nats_util::{self, NatsArgs};
 use shared::prost::Message;
 use shared::protobuf::event::{Event, event::PeerObserverEvent};
-use shared::protobuf::rpc_extractor;
+use shared::protobuf::rpc_extractor::{self, EstimateSmartFee, FeeEstimateMode};
 use shared::tokio::sync::watch;
 use shared::tokio::time::{self, Duration};
 use shared::{async_nats, clap};
@@ -116,6 +117,10 @@ pub struct Args {
     /// Disable querying and publishing of `getrawaddrman` data.
     #[arg(long, default_value_t = false)]
     pub disable_getrawaddrman: bool,
+
+    /// Disable querying and publishing of `estimatesmartfee` data.
+    #[arg(long, default_value_t = false)]
+    pub disable_estimatesmartfee: bool,
 }
 
 impl Args {
@@ -139,6 +144,7 @@ impl Args {
         disable_getblockchaininfo: bool,
         disable_getorphantxs: bool,
         disable_getrawaddrman: bool,
+        disable_estimatesmartfee: bool,
     ) -> Args {
         Self {
             nats,
@@ -161,6 +167,7 @@ impl Args {
             disable_getblockchaininfo,
             disable_getorphantxs,
             disable_getrawaddrman,
+            disable_estimatesmartfee,
             // when adding more disable_* args, make sure to update the disable_all below
         }
     }
@@ -243,6 +250,10 @@ pub async fn run(args: Args, mut shutdown_rx: watch::Receiver<bool>) -> Result<(
         "Querying getrawaddrman enabled: {}",
         !args.disable_getrawaddrman
     );
+    log::info!(
+        "Querying estimatesmartfee enabled: {}",
+        !args.disable_estimatesmartfee
+    );
     // check if we have at least one RPC to query
     let disable_all = args.disable_getpeerinfo
         && args.disable_getmempoolinfo
@@ -254,7 +265,8 @@ pub async fn run(args: Args, mut shutdown_rx: watch::Receiver<bool>) -> Result<(
         && args.disable_getnetworkinfo
         && args.disable_getblockchaininfo
         && args.disable_getorphantxs
-        && args.disable_getrawaddrman;
+        && args.disable_getrawaddrman
+        && args.disable_estimatesmartfee;
     if disable_all {
         log::warn!("No RPC configured to be queried!");
     }
@@ -293,6 +305,10 @@ pub async fn run(args: Args, mut shutdown_rx: watch::Receiver<bool>) -> Result<(
                 if !args.disable_getorphantxs
                     && let Err(e) = getorphantxs(&rpc_client, &nats_client, &metrics).await {
                         log::error!("Could not fetch and publish 'getorphantxs': {}", e)
+                }
+                if !args.disable_estimatesmartfee
+                    && let Err(e) = estimatesmartfee(&rpc_client, &nats_client, &metrics).await {
+                        log::error!("Could not fetch and publish 'estimatesmartfee': {}", e)
                 }
             }
             _ = less_frequent_interval.tick() => {
@@ -629,5 +645,52 @@ async fn getrawaddrman(
                 .with_label_values(&["getrawaddrman"])
                 .inc();
         })?;
+    Ok(())
+}
+
+async fn estimatesmartfee(
+    rpc_client: &Client,
+    nats_client: &async_nats::Client,
+    metrics: &Metrics,
+) -> Result<(), FetchOrPublishError> {
+    const BLOCKS_10MIN: u32 = 1;
+    const BLOCKS_1HOUR: u32 = 6;
+    const BLOCKS_1DAY: u32 = 144;
+    const BLOCK_TARGETS: [u32; 3] = [BLOCKS_10MIN, BLOCKS_1HOUR, BLOCKS_1DAY];
+
+    const MODES: [FeeEstimateMode; 2] =
+        [FeeEstimateMode::Economical, FeeEstimateMode::Conservative];
+
+    for target in BLOCK_TARGETS {
+        for mode in MODES {
+            let estimate: RPCEstimateSmartFee =
+                measure_rpc_call("estimatesmartfee", metrics, || {
+                    // TODO: when https://github.com/rust-bitcoin/corepc/pull/531 gets released
+                    // use estimate_smart_fee_with_mode instead of the generic call
+                    rpc_client.call::<RPCEstimateSmartFee>(
+                        "estimatesmartfee",
+                        &[target.into(), mode.to_string().into()],
+                    )
+                })?;
+            let estimate = estimate.into_model().unwrap();
+
+            let proto = Event::new(PeerObserverEvent::RpcExtractor(rpc_extractor::Rpc {
+                rpc_event: Some(rpc_extractor::rpc::RpcEvent::EstimateSmartFee(
+                    EstimateSmartFee::from_rpc(estimate, target, mode),
+                )),
+            }))?;
+
+            nats_client
+                .publish(Subject::Rpc.to_string(), proto.encode_to_vec().into())
+                .await
+                .inspect_err(|_| {
+                    metrics
+                        .nats_publish_errors
+                        .with_label_values(&["estimatesmartfee"])
+                        .inc();
+                })?;
+        }
+    }
+
     Ok(())
 }
--- a/extractors/rpc/tests/integration.rs
+++ b/extractors/rpc/tests/integration.rs
@@ -18,9 +18,12 @@ use shared::{
     prost::Message,
     protobuf::{
         event::{Event, event::PeerObserverEvent},
-        rpc_extractor::rpc::RpcEvent::{
-            Addrman, AddrmanInfo, BlockchainInfo, ChainTxStats, MemoryInfo, MempoolInfo, NetTotals,
-            NetworkInfo, OrphanTxs, PeerInfos, Uptime,
+        rpc_extractor::{
+            FeeEstimateMode,
+            rpc::RpcEvent::{
+                Addrman, AddrmanInfo, BlockchainInfo, ChainTxStats, EstimateSmartFee, MemoryInfo,
+                MempoolInfo, NetTotals, NetworkInfo, OrphanTxs, PeerInfos, Uptime,
+            },
         },
     },
     testing::nats_server::NatsServerForTesting,
@@ -585,6 +588,75 @@ async fn test_integration_rpc_testsshouldtimeout() {
         |event| match event {
             PeerObserverEvent::RpcExtractor(_) => {
                 panic!("We should never receive an event here.")
+            }
+            _ => panic!("unexpected event {:?}", event),
+        },
+    )
+    .await;
+}
+
+#[tokio::test]
+async fn test_integration_rpc_estimatesmartfee() {
+    println!("test that we receive estimatesmartfee RPC events");
+
+    check(
+        EnabledRPCsInTest {
+            estimatesmartfee: true,
+            ..Default::default()
+        },
+        |node1, node2| {
+            let miner_address = node1
+                .client
+                .new_address()
+                .expect("failed to get new address");
+            node1
+                .client
+                .generate_to_address(101, &miner_address)
+                .expect("failed to generate to address");
+
+            // generate some transactions to create enough data for fee estimation
+            for _ in 0..20 {
+                let address = node2
+                    .client
+                    .new_address()
+                    .expect("failed to get new address");
+                node1
+                    .client
+                    .call::<String>("sendtoaddress", &[address.to_string().into(), "1".into()])
+                    .expect("failed to send to address");
+
+                node1
+                    .client
+                    .generate_to_address(1, &miner_address)
+                    .expect("failed to generate to address");
+            }
+        },
+        |event| match event {
+            PeerObserverEvent::RpcExtractor(r) => {
+                if let Some(ref e) = r.rpc_event {
+                    match e {
+                        EstimateSmartFee(result) => {
+                            // Fee rate most of the time is 0.0001_0000 on this scenario, but sometimes is greater
+                            assert!(result.fee_rate >= Some(10.0));
+                            assert!(result.fee_rate < Some(10.1));
+
+                            let expected_combinations = [
+                                (1, FeeEstimateMode::Economical as i32),
+                                (1, FeeEstimateMode::Conservative as i32),
+                                (6, FeeEstimateMode::Economical as i32),
+                                (6, FeeEstimateMode::Conservative as i32),
+                                (144, FeeEstimateMode::Economical as i32),
+                                (144, FeeEstimateMode::Conservative as i32),
+                            ];
+                            let found = expected_combinations.iter().any(|(blocks, mode)| {
+                                *blocks == result.requested_blocks && *mode == result.mode
+                            });
+
+                            assert!(found);
+                        }
+                        _ => panic!("unexpected RPC data {:?}", r.rpc_event),
+                    }
+                }
             }
             _ => panic!("unexpected event {:?}", event),
         },
--- a/extractors/rpc/tests/integration.rs
+++ b/extractors/rpc/tests/integration.rs
@@ -148,10 +148,9 @@ async fn test_integration_rpc_getmempoolinfo() {
                             assert_eq!(info.bytes, 0);
                             assert_eq!(info.total_fee, 0.0);
                             assert_eq!(info.max_mempool, 300000000);
-                            // These will change between v29 and v30, so don't hardcode something here.
-                            assert!(info.mempoolminfee > 0.0);
-                            assert!(info.minrelaytxfee > 0.0);
-                            assert!(info.incrementalrelayfee > 0.0);
+                            assert_eq!(info.mempoolminfee, 0.1);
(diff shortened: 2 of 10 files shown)


extractors/rpc/src/lib.rs
```diff
@@ -631,3 +647,61 @@ async fn getrawaddrman(
...
+                    target,
+                    mode,
+                    estimative.errors
+                );
+                continue;
+            }
+
+            let mut estimative: EstimateSmartFee = estimative.into();
+            estimative.blocks = target; // TODO: core outputs a different value from what we send, should we override?
+            estimative.mode = mode.into();
```
m4ycon: Should we? Maybe this is a problem for a pruned node that doesn't run 24/7 like mine, when fetching for targets like 144, it suggests a "lower" target, which kind of makes sense as probably at some point the feerate would be really low before the target (?). But it should be more of a "fixed" value as this makes it easier to analyze.

extractors/rpc/tests/integration.rs
```diff
@@ -591,3 +594,71 @@ async fn test_integration_rpc_testsshouldtimeout() {
...
+    println!("test that we receive estimatesmartfee RPC events");
+
+    let expected_combinations = [
+        (1, FeeEstimateMode::Economical as i32),
+        (1, FeeEstimateMode::Conservative as i32),
+        (6, FeeEstimateMode::Economical as i32),
+        (6, FeeEstimateMode::Conservative as i32),
+        (144, FeeEstimateMode::Economical as i32),
+        (144, FeeEstimateMode::Conservative as i32),
+    ];
```
m4ycon: In the current state of this test, I just check if any of these options are delivered, just one of them as the `check` method allows. I thought of refactoring it and make something like I did in metrics tests (0090e580d25b3a3be451b1cd1d496f89f56b5724), having something like a expected list of events (?), in this case it would match every expected event (not less not more). Maybe some tests doesn't fit into it, but what do you think?

tools/metrics/dashboards/fees-estimatesmartfee.json
m4ycon, line 1: 1. Any tips here? I know that id's here get messed up when importing and using a "new" datasource.
2. About playlist, how can I fit this in playlists? I mean, this is similar to `logs.json`.
3. Also, I did dashboards like the image below, what do you think? In the current state, we could see just one node data, how can we see many? Maybe one dash (3 in total) for each X target blocks with each "line" being a node?

Image

m4ycon (commented):
Those are questions that I would appreciate some opinion, so I can work (or not) on them, so then I can set this PR as ready to review.

shared/src/protobuf/rpc_extractor.rs
```diff
@@ -537,3 +538,35 @@ impl From<GetRawAddrMan> for Addrman {
...
+impl From<RPCEstimateSmartFee> for EstimateSmartFee {
+    fn from(estimate: RPCEstimateSmartFee) -> Self {
+        let fee_rate_btc_per_kvb = estimate.fee_rate.unwrap_or(0.);
+        let fee_rate_sat_per_vb = fee_rate_btc_per_kvb * 1e8 / 1e3;
+
+        EstimateSmartFee {
+            fee_rate: fee_rate_sat_per_vb,
+            blocks: estimate.blocks,
+            mode: FeeEstimateMode::default().into(),
+        }
```
GuiSchet, line 582: Any future caller that forgets to do so will silently get wrong data — `mode = UNSET` and a potentially wrong `blocks` value. 
We should replace `From` with a dedicated constructor `EstimateSmartFee::from_rpc(estimate, target, mode)` that takes the missing context as arguments.

shared/src/protobuf/rpc_extractor.rs
```diff
@@ -537,3 +538,35 @@ impl From<GetRawAddrMan> for Addrman {
...
+
+impl fmt::Display for FeeEstimateMode {
+    fn fmt(&self, estimate: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(estimate, "{}", self.as_str_name())
+    }
+}
+
+impl From<RPCEstimateSmartFee> for EstimateSmartFee {
+    fn from(estimate: RPCEstimateSmartFee) -> Self {
+        let fee_rate_btc_per_kvb = estimate.fee_rate.unwrap_or(0.);
```
GuiSchet: If `fee_rate` is `None` but `errors` is also `None`, a `0.0` gauge gets published, which looks like a valid low fee rate, should add a guard maybe?

extractors/rpc/tests/integration.rs
```diff
@@ -591,3 +594,71 @@ async fn test_integration_rpc_testsshouldtimeout() {
...
+                    .expect("failed to generate to address");
+            }
+        },
+        |event| match event {
+            PeerObserverEvent::RpcExtractor(r) => {
+                if let Some(ref e) = r.rpc_event {
+                    match e {
+                        EstimateSmartFee(result) => {
+                            // Fee rate most of the time is 0.0001_0000 on this scenario, but sometimes is greater
+                            assert!(result.fee_rate >= 10.0);
```
GuiSchet: The comment and the assertion contradict each other. `0.0001 BTC/kVB` converts to `0.01 sat/vB`, well below the asserted `10.0` threshold.

GuiSchet (commented):
ey M4ycon, good work! I made some minor comments...

extractors/rpc/tests/integration.rs
m4ycon: 🤔 Are you sure about this? My math was:

1. 0.0001_0000 BTC/kVB
2. 0.0001_0000 * 10^8 = 10000 sats/kVB
3. 10000 / 10^3 = 10 sats/vB

If this is wrong, let me know where exactly. I don't understand "vByte" yet but since I don't change this unit, I suppose it doesn't make a difference here.

GuiSchet: oh I misread the comment, thought 0.0001_0000 was the typical value returned, not the minimum
the assertion is fine!

m4ycon:
Reorganized how I handle `EstimateSmartFee` possible errors and also avoid to construct this object with wrong data as suggested by @GuiSchet 👍 .

extractors/rpc/src/lib.rs
```diff
@@ -631,3 +647,54 @@ async fn getrawaddrman(
...
+    const BLOCK_TARGETS: [u32; 3] = [BLOCKS_10MIN, BLOCKS_1HOUR, BLOCKS_1DAY];
+
+    const MODES: [FeeEstimateMode; 2] =
+        [FeeEstimateMode::Economical, FeeEstimateMode::Conservative];
+
+    for target in BLOCK_TARGETS {
+        for mode in MODES {
+            let estimative: RPCEstimateSmartFee =
+                measure_rpc_call("estimatesmartfee", metrics, || {
+                    // TODO: when https://github.com/rust-bitcoin/corepc/pull/531 gets released
```
0xB10C, line 668: Once we merge this peer-observer PR #386, it would be good to open an issue for this so we don't forget about this.

extractors/rpc/src/lib.rs
0xB10C: I think we should make it clear in the event which `target` we requested and for which the node returned an estimate `target`. Hiding it here makes no sense. We should also document that in the event docs.

In e.g. the metrics tool, we might need to handle these differently, but that's not the job of the RPC extractor.

extractors/rpc/tests/integration.rs
```diff
@@ -591,3 +594,71 @@ async fn test_integration_rpc_testsshouldtimeout() {
...
+                node1
+                    .client
+                    .call::<String>("sendtoaddress", &[address.to_string().into(), "1".into()])
+                    .expect("failed to send to address");
+
+                node1
+                    .client
+                    .generate_to_address(1, &miner_address)
+                    .expect("failed to generate to address");
+            }
```
0xB10C, line 631: What is this loop doing? Generating UTXOs for us?

shared/src/protobuf/rpc_extractor.rs
```diff
@@ -537,3 +538,48 @@ impl From<GetRawAddrMan> for Addrman {
...
+    }
+}
+
+impl EstimateSmartFee {
+    pub fn from_rpc(
+        estimate: RPCEstimateSmartFee,
+        target: u32,
+        mode: FeeEstimateMode,
+    ) -> Result<Self, String> {
+        if estimate.errors.is_some() {
```
0xB10C: I think we should pass estimation errors on in the peer-observer event. These are important for tools to known about, because they could indicate a problem

tools/metrics/dashboards/fees-estimatesmartfee.json
0xB10C: 1. I'm not sure how to fix the datasource ID error yet. We have to experiment a bit I guess. I think it should work somehow. Maybe we can agree on a fixed datasource id for all dashboards, but I'm not sure you can set the ID in grafana?

0xB10C: 2. Feel free to just do a overview page with `fees.json` or similar. We can always add a playlist dashboard later.

0xB10C: yes, supporting multiple-node dasboards would be good. Maybe having six panels (1-eco, 1-cons, 6-eco, ...) would be fine. Each showing the data for all nodes.

extractors/rpc/tests/integration.rs
m4ycon: It's generating some transactions with a fee, as coinbase doesn't count to this, so when we call `estimatesmartfee` it has enough data and we don't get a `{"errors":["Insufficient data or no feerate found"],"blocks":0}`.

tools/metrics/dashboards/fees-estimatesmartfee.json
m4ycon: Made changes to the dash, now it displays by nodes, and I created a grafana variable for datasource that can be changed via dash. It seems that will be easier to fix the id error that way, instead of going on each panel to fix it, it would be just one place (top left).

> Maybe we can agree on a fixed datasource id for all dashboards, but I'm not sure you can set the ID in grafana?

Yeah, seems to be the easier solution, we can edit the `json` exports with it.

I will still work on the other comments tomorrow, then I commit this new dash.

image

extractors/rpc/tests/integration.rs
0xB10C: Makes sense! Please add a comment about this!

m4ycon:
- Added `requested_blocks` and `errors` to `EstimateSmartFee` protobuf message;
- Added a comment in rpc `estimatesmartfee` test to explain about minimum feerate data required by it;
- Refactored fees dashboard to a model that supports multiple sources (nodes) and displays each combination of target and mode into separate panels.

I think this PR is ready for review, I don't have more questions.

image

tools/metrics/src/lib.rs
```diff
@@ -981,6 +981,18 @@ fn handle_rpc_event(e: &rpc::RpcEvent, state_arc: Arc<Mutex<State>>, metrics: me
             }
             state.addrman = addrman.clone();
         }
+        rpc::RpcEvent::EstimateSmartFee(estimate) => {
+            if let Some(fee_rate) = estimate.fee_rate {
+                let mode = FeeEstimateMode::try_from(estimate.mode)
+                    .expect("Invalid fee estimate mode")
+                    .to_string();
```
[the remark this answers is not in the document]
m4ycon: I don't know about this one, I saw some other examples that seems to not care about having an `expect` call which could lead to the same problem. Any opinions here? I'm comfortable to let it that way.

0xB10C:
Nice, I'll do another code review pass soon and will then run it on the demo nodes for us to collect some sample data. Thanks @m4ycon!

0xB10C:
With merging #377 there seems to be a conflict on the metrics integration test. Can you rebase? Also, you probably want to add documenation for the newly supported RPC (see #390), otherwise CI is going to complain.

m4ycon:
Rebased with `master` and added `estimatesmartfee` row to rpc-extractor readme table.

0xB10C:
I've deployed your branch to the demo nodes for testing. You can find the dashboard here: https://demo.peer.observer/monitoring/d/ad72ms8/fees-fee-rate-estimates?orgId=1&from=now-24h&to=now&timezone=browser&var-datasource=P1809F7CD0C75ACF3&refresh=1m

`logger --rpc` currently prints:
```
rpc: EstimateSmartFee(estimation failed, blocks=0, mode=ECONOMICAL, errors=["Insufficient data or no feerate found"])
rpc: EstimateSmartFee(estimation failed, blocks=0, mode=CONSERVATIVE, errors=["Insufficient data or no feerate found"])
rpc: EstimateSmartFee(estimation failed, blocks=0, mode=ECONOMICAL, errors=["Insufficient data or no feerate found"])
rpc: EstimateSmartFee(estimation failed, blocks=0, mode=CONSERVATIVE, errors=["Insufficient data or no feerate found"])
rpc: EstimateSmartFee(estimation failed, blocks=0, mode=ECONOMICAL, errors=["Insufficient data or no feerate found"])
rpc: EstimateSmartFee(estimation failed, blocks=0, mode=CONSERVATIVE, errors=["Insufficient data or no feerate found"])
```

I assume this just takes a few more minutes? Will check back soon.

m4ycon:
Nice, it seems to be working right now! I think it takes some time to see different values between each panel. And just to share here, I've been running this on mine and it's looking like this.

image

extractors/rpc/src/lib.rs
```diff
@@ -139,6 +144,7 @@ impl Args {
         disable_getblockchaininfo: bool,
         disable_getorphantxs: bool,
         disable_getrawaddrman: bool,
+        disable_getestimatesmartfee: bool,
```
0xB10C: No need to add the `get` here 😄 

```suggestion
        disable_estimatesmartfee: bool,
```

extractors/rpc/src/lib.rs
```diff
@@ -1,6 +1,7 @@
 use shared::clap::{ArgGroup, Parser};
 use shared::corepc_client::client_sync::Auth;
 use shared::corepc_client::client_sync::v30::Client;
+use shared::corepc_client::types::v30::EstimateSmartFee as RPCEstimateSmartFee;
 use shared::corepc_node::mtype::{
     GetBlockchainInfo, GetChainTxStats, GetNetworkInfo, GetOrphanTxsVerboseTwo,
 };
```
0xB10C, line 7: Hm, I think we might want to use the generic `mtype` for `EstimateSmartFee` here too, if we can. This avoids us having to update this (hopefully) once we move to something newer than v30.

https://docs.rs/corepc-node/latest/corepc_node/mtype/struct.EstimateSmartFee.html

shared/src/protobuf/rpc_extractor.rs
0xB10C, line 18: Same here. I think we should use the generic mtypes where we can.

Is avaliable, so I think we should be able to use it here: https://docs.rs/corepc-node/latest/corepc_node/mtype/struct.EstimateSmartFee.html

tools/metrics/src/metrics.rs
```diff
@@ -110,6 +112,16 @@ macro_rules! g {
...
+macro_rules! gv {
+    ($name:ident, $desc:expr, $labels:expr, $registry:expr) => {
+        let $name: GaugeVec = register_gauge_vec_with_registry!(
+            Opts::new(stringify!($name), $desc),
+            &$labels,
+            $registry
+        )
+        .expect(concat!("Could not create metric '", stringify!($name), "'"));
+    };
+}
```
0xB10C, line 124: nice! yes I think adding a GaugeVec for this is good.

extractors/rpc/tests/integration.rs
```diff
@@ -39,7 +42,7 @@ const TEST_TIMEOUT_SECONDS: u64 = 5;
 async fn check(
     rpcs: EnabledRPCsInTest,
     test_setup: fn(&corepc_node::Node, &corepc_node::Node),
-    check_expected: fn(PeerObserverEvent) -> (),
+    mut check_expected: impl FnMut(PeerObserverEvent),
```
0xB10C: I was wondering why this has to be `mut` now.

extractors/rpc/tests/integration.rs
```diff
@@ -591,3 +594,72 @@ async fn test_integration_rpc_testsshouldtimeout() {
...
+        |node1, node2| {
+            let miner_address = node1
+                .client
+                .new_address()
+                .expect("failed to get new address");
+            node1
+                .client
+                .generate_to_address(101, &miner_address)
+                .expect("failed to generate to address");
+
```
0xB10C, line 615: I recently learned from another project that using an address from the wallet of the node makes tests a lot slower since the wallet spends time processing each transaction to it's addresses. Since we are not spending the UTXOs here, we could use a static regtest address here and in other places. I've used the one on https://bitcoin.stackexchange.com/questions/91222/bech32-addresses-beginning-with-bcrt-on-regtest (`bcrt1qs758ursh4q9z627kt3pp5yysm78ddny6txaqgw`) in the past.

I don't think you need to fix it up here, but maybe we do it in another PR. There, we can then fix up all places were we mine to a wallet address.

tools/metrics/dashboards/fees-estimatesmartfee.json
```diff
@@ -0,0 +1,618 @@
...
+        "options": [],
+        "query": "prometheus",
+        "refresh": 1,
+        "regex": "",
+        "type": "datasource"
+      }
+    ]
+  },
+  "time": {
+    "from": "now-24h",
```
0xB10C: We could even do new-7d or similar. 24h is usually a short timeframe for monitoring and we might miss something if we don't look at it every day.

0xB10C:
Dashboard is now filled up nicely!

Code looks good. Left a few minor comments that I think would be good to address. Mostly using `mtype` if possible.

Afterwards, we can merge this.

extractors/rpc/tests/integration.rs
m4ycon: Hmm makes sense. I can create an issue for it, as I'll do for the other one related to CorePC when this PR is closed.

extractors/rpc/src/lib.rs
m4ycon: omg 🤭, I remember fixing this, but it seems I've missed this one .-.

extractors/rpc/src/lib.rs
m4ycon: I did it for `rpc_extractor.rs`, but for this case I can't. The "raw" call that I do for `estimatesmartfee` returns a `fee_rate: Option<f64>` which `v30::EstimateSmartFee` type matches correctly, but `mtype::EstimateSmartFee` doesn't, it has a `fee_rate: Option<FeeRate>` struct instead.

I saw that `shared` module is pulling corepc from your fork, I think if you update it with master, I can do it here too.

extractors/rpc/tests/integration.rs
m4ycon: Previously, I wanted to make my test match every expected rpc event, convering all combinations of targets and modes, since this one sends 6 calls "at once". But the `check` method only checked the first event, so I couldn't keep my idea - I forgot to undo this hehe.

m4ycon:
Updates:
- Fixed `disable_getestimatesmartfee` outdated naming;
- Used `mtype::EstimateSmartFee` on `rpc_extractor.rs`;
- Undid change to `check` on rpc integration tests;
- Updated default grafana dash time range to "now-7d";
- Updated grafana panels to show 3 decimal places, as sometimes they rounded to 2 decimal places.

And a last one, fixes #356 with 00289fc, since I needed to use it on `EstimateSmartFee::from_rpc`.

RazorBest (commented):
Ran the tests with `NATS_SERVER_BINARY=nats-server cargo test --release --features nats_integration_tests --features node_integration_tests` successfully.

Tested the metrics with a grafana dashboard
Image

m4ycon:
Updated `fees-estimatesmartfee.json` with "export externally format" from grafana, following #403.

0xB10C:
> Updated `fees-estimatesmartfee.json` with "export externally format" from grafana, following #403.

I think this doesn't work for provisioning: https://demo.peer.observer/monitoring/d/ad72ms8/fees-fee-rate-estimates

`Datasource ${DS_PROMETHEUS} was not found.`

m4ycon:
Undid change to `fees-estimatesmartfee.json` and set datasource id to "P1809F7CD0C75ACF3" as it is used already set on other dashboards.

extractors/rpc/src/lib.rs
0xB10C: I think this was released recently! Will upgrade corepc soon!

0xB10C:
Thanks @m4ycon!

Issue #304, opened 2025-12-10
metrics-tool: anomaly-detection with Prometheus recording rules


0xB10C:
As part of #13, I think an interesting approach is to use the metrics from the [metrics-tool](https://github.com/peer-observer/peer-observer/tree/master/tools/metrics) and Prometheus to detect anomalies. I came across this blog post [How to use Prometheus to efficiently detect anomalies at scale](https://grafana.com/blog/2024/10/03/how-to-use-prometheus-to-efficiently-detect-anomalies-at-scale/) (based on this talk https://www.youtube.com/watch?v=BTAba-Vq3xE). This looks interesting and is probably something worth trying out.

They published Prometheus recording rules here: https://github.com/grafana/promql-anomaly-detection/tree/main/rules

One of the metrics to start out with is probably number of inbound connections to the node. It's an anomaly, when one or more of the monitored nodes start dropping connections over a time span X. See also https://b10c.me/observations/15-inv-to-send-queue/. As a next step, it's worthwhile to think about not treating node restarts as anomalies, even if the inbound node count drops to zero briefly. 

<img width="1684" height="455" alt="Image" src="https://github.com/user-attachments/assets/c0a96c61-a0b5-44aa-af79-3688ea8c0a3e" />

The bulk of the work is going to be thinking about how to best apply the Prometheus recording rules to Bitcoin specific P2P metrics and what to monitor for. The goal would be to have alerting based on the Prometheus anomaly detection.

Happy to help brainstorm other metrics relevant for anomaly-detection if someone picks this up and needs input. 

Also, very happy to provide data and metrics for this. See e.g. the [Grafana](https://demo.peer.observer/monitoring) data on https://demo.peer.observer/. I can provide a NATS endpoint that sends you events in real-time: You shouldn't be required to run your own peer-observer instance just for working on this! However, being able to run e.g. the metrics tool locally and being familiar with the rough peer-observer infrastructure is good. You being able to run Prometheus locally is probably a requirement. Also, setting up a Grafana instance with a dashboard is probably good. Maybe the free https://grafana.com/products/cloud/ offering can be used to start off (though I haven't checked).

deadmanoz:
I have started to look at this.

Any BOSS participants also welcome to look at this as well!

RazorBest:
@deadmanoz
I can also look into this if that's okay. Let me know if you already did some work on it. Otherwise I’m happy to take the first step.

deadmanoz:
[...]

PR #400, opened 2026-04-13
Generic anomaly detection with Prometheus

RazorBest:
Last update: 19.05.2026 14:59 UTC 

This is a first iteration of https://github.com/peer-observer/peer-observer/issues/304.

anomaly_2

## Methodology

This PR is based on: https://grafana.com/blog/how-to-use-prometheus-to-efficiently-detect-anomalies-at-scale/. I started [from one of their files](https://github.com/grafana/promql-anomaly-detection/blob/main/rules/adaptive.yml), and then changed it. They define some generic rules that apply for all the metrics that have the `anomaly_name` label.

From their article, I implemented:
- Establishing baselines: a first attempt
- Extreme outliers
- Low sensitivity
- Discontinuities

To summarize this, they used 2 main ways of detecting anomalies:
- When a sample is too far from the mean (no information about stddev). This is represented in `anomalies.yml` by the expressions `anomaly:adaptive:avg_1h / 1.5` (lower bound) and `anomaly:adaptive:avg_1h * 1.5` (upper bound).
- When a sample is exceeding a custom z-score, which are recorded by `lower_band_st` and `upper_band_st`.

Then, for example, for the final `lower_band`, we merge the 2 metrics above:
```
min(anomaly:adaptive:avg_1h / 1.5, lower_band_st) // Pseudocode
```

Then, I took these metrics, that I found interesting, and selected them for anomaly detection:
- https://demo.peer.observer/monitoring/d/bdmeshfmf2i9sb/connections-closed-connections-per-network?orgId=1&from=now-24h&to=now&timezone=browser
- https://demo.peer.observer/monitoring/d/ddukz6s4vr1moc/addr-inbound-addr-entries?orgId=1&from=now-12h&to=now&timezone=browser
- https://demo.peer.observer/monitoring/d/cdhprcwaivta8c/connections-current-inbound-connections-per-host?orgId=1&from=now-7d&to=now&timezone=browser
- https://demo.peer.observer/monitoring/d/a44a6f0b-0c19-41e6-a0de-bb6b27428c42/36cddba?orgId=1&from=now-7d&to=now&timezone=utc&var-networks=$__all
- https://demo.peer.observer/monitoring/d/784115ce-faf6-4beb-9ea2-603dd0d74e68/46628b0?orgId=1&from=now-7d&to=now&timezone=utc

## Labeling a metric for anomaly check

I think it's best to choose for what metrics we need to measure anomalies. One way of doing this is through a custom label (e.g. `anomaly_name`). Our anomaly rules would consider all the metrics with that label. You can check `monotonic.yml` for some examples.

There is a small issue with using rules, though. Let's look at:
```
      - record: peerobserver_conn_inbound_current_anomaly
        expr: peerobserver_conn_inbound_current
        labels:
          anomaly_name: peerobserver_conn_inbound_current_anomaly
          anomaly_type: latency
```

This creates a new metric named `peerobserver_conn_inbound_current_anomaly`, that has the same values as `peerobserver_conn_inbound_current`. The only difference is in the labels.

So, an alternative to using rules is to label those metrics directly. This can be done from the metrics-tools. I know that `prometheus.yml` lets you add custom labels at the scraping step, but I don't think it's granular. You can't say "Add this label only to peerobserver_conn_inbound_current".

## Setting up Prometheus

I am aware of the nix infra-libray, but I was not able to build only service.prometheus, to test that the configurations translate to the correct yml files. If anyone is able to show an example of an iteration cycle (modify the appropriate nix file, build, then get the yml files), it would be appreciated.

There are 4 relevant items:
- `tools/metrics/prometheus/prometheus.yml`: prometheus root config
- `tools/metrics/prometheus/rules/`: a directory where all the prometheus rules are stored in yml files
- `tools/metrics/template-compose-yml`: template example for docker compose
- `tools/metrics/prometheus/backfill_rules.sh`: script used for backfilling

I'm running my processes with Docker compose, so I also included `tools/metrics/template-compose-yml`, with an example of how to set up prometheus with compose. It doesn't contain all the necessary services, because I don't know the environments of other people.

You need to tell prometheus where to find the `prometheus.yml` file. This config file contains a field named `rule_files` which specifies the paths to the recording and alerting rules (relative to `prometheus.yml`). So, if I have `/etc/prometheus/prometheus.yml`, and the value of `rule_files` is `"rules/*"`, then prometheus expects the rules directory to be at `/etc/prometheus/rules`.

**Backfilling:**

When you add new recording rules, Prometheus applies them periodically. The result of these rules are new metrics. However, these metrics only start from where the new rules were added. If you want the rules to apply to previous data, you need to do this manually. Fortunately, `promtool` is capabale of backfilling, but this is external from the TSDB. The tool does the same queries you would imagine if you wrote such a tool by yourself.

As a consequence, there are some issues:
- The tool is a little slow: on my slow dev machine, backfilling 7 days takes about 2 minutes. This scales linearly, so 90 days of past data would take 25 minutes on my machine
- Each rule is done separately, the time also scales with the number of rules
- When you have rules that depend on other rules, the order in which rules are filled matters. This can be solved by running `promtool` multiple times
- From my tests, the behaviour is non-deterministic when you have metrics that started after the `--start` timestamp given to `promtool`. Changing the value `--start` solved this.

Now that you are aware of the instability of this, I can tell you that you can try to run `tools/metrics/prometheus/backfill_rules.sh`. This will create a `data` directory with the new blocks. If you use docker, then you can run this at `/prometheus`, and it will write directly at `/prometheus/data`, where the TSDB is stored.

Reference: https://prometheus.io/docs/prometheus/latest/storage/#backfilling-for-recording-rules

## Grafana dashboards

There are two new dashboards:
- One where the anomalies are determined from the recording rules
- One that's identical to the one above, but doesn't use recording rules. Basically, it turns everything into big expressions that do what the recording rules would do.

You can use the second dashboard (no recording rules) to play with the anomaly parameters, and see how the bands change.
vareiables

Here is the meaning of the variables, from my understanding:
- `THRESHOLD_BY_COVAR`: we only consider the standard deviation when it's more than `mean * THRESHOLD_BY_COVAR`. This means that small standard deviations will not affect the z-score.
- `STDDEV_MULTIPLIER`: the constant used for the z-score
- `MARGIN_MULTIPLIER`: the constant used by the first method of detecting anomalies. A sample is considered anomaly if it's outside the interval `[mean / MARGIN_MULTIPLIER, mean * MARGIN_MULTIPLIER]`. The devs of [promql-anomaly-detection](https://github.com/grafana/promql-anomaly-detection) use `[mean -  mean * MARGIN_MULTIPLIER, mean + mean * MARGIN_MULTIPLIER]`, which seems wrong to me.
- `INCREASE_INTERVAL`: the range used for computing the increase of our monotonic metrics
- `SHORT_LOOKBACK`: the mean and the simple standard deviation are computed from samples of the interval of time determined by this variable. The authors used `1h` in their article.
- `LONG_LOOKBACK`: this is what's used to customize the z-score. When we have spikes in the standard deviation that pass the `STDDEV_MULTIPLIER` threshold, we take that and expand it for the entire `LONG_LOOKBACK` interval. This means that short spikes will have a long lasting effect, and increase the bands. The authors used `26h` for this.
- `HOST`:  the name of the node

**17 May Update**

### Example 1
total_addrv2_example1

Plot of total addrv2 received over 10 hours.

- red: a scaled down line of `peerobserver_p2p_addrb2_addresses_sum`
- green: the increase of the red metric
- gray: the area between the upper and lower bands

When the node is started, there is a spike in the amount of received addresses, but the increase is mostly constant. The first fall is probably caused by a restart. After a couple of hours, there are 4 blips in the inrease. All of them are identified as anomalies. The third blip is so high that its decrease is reported as an anomaly as well.

### Example 2.1
total_inbound_addr_example2

Plot of total received addresses in inbound addr messages across 10 hours. `THRESHOLD_BY_COVAR` is 0.5.

There are two detected anomalies. The first, at a steep decrease. The, second, at another steep decrease. There is a discontinuity at the gray bands at around 6:50. This comes from `anomaly:adaptive:stddev_st`, which looks at the last 6 hours of `anomaly:adaptive:stddev_1h:filtered`. If the `filtered` metric doesn't have any samples in a window of 6 hours (which happened in this case), then `stddev_st` will also not register any sample.

## Example 2.2
total_inbound_addr_example3

Plot of total received addresses in inbound addr messages across 10 hours. `THRESHOLD_BY_COVAR` is 0.25.

The green line is the same as in the previous example, where `THRESHOLD_BY_COVAR` was 0.5. There is one detected anomaly. The effect of lowering that variable was that the lower band went lower, so the first steep decrease is not detected anymore. Moreover, the second steep decrease is almost undetected. This is quite an unexpected result, which I believe that comes from how `THRESHOLD_BY_COVAR` and `LONG_LOOKBACK` interact with each other. Indeed, when I take this eexample and change `LONG_LOOKBACK` from 6h to 1h, the first anomly is detected again. This case needs more analysis.

## Example 3.1
closed_connections_total_addresses_example4

Two plots: closed connections and total addrv2 received. `INCREASE_INTERVAL` is 30m.

In both plots, there an anomaly happening between 11:00 and 12:00: an increased number of closed connections and a increased number of received addrv2 addresses. This could be explained by the host disconnecting from some previous peers, connecting to new peers, and as a consequence, the new peers send addrv2 messages at the beginning.

In the second plot, 4 hours after the first anomaly, there is a second anomaly that can't be correlated with any change in the first plot: a number of increased addrv2 addresses without an increase in the close connections. This could be explained by the fact that the addrv2 anomaly can be triggered by a single peer that sends a lot of addresses.

In the second plot, the width of the blips is not related to the quantity of the increase, but to `INCREASE_INTERVAL`.

## Example 3.2
closed_connections_total_addresses_example5

Similar to the previous example - two plots: closed connections and total addrv2 received. `INCREASE_INTERVAL` is 5m.

In the first plot (closed connections), a lot more anomalies are happening. One observation is that it has little entropy, with values between 0 and 4.

In the second plot, we can observe 4 blips, out of which 3 are identified as anomalies. The values that define the range of this metric are higher: between 200 and 2000.

This time, there's no observable correlation between the anomalies in the two plots.


dutenrapha:
Hey, @RazorBest! 

I went through the rules, dashboards, and your notes, and I think the overall structure makes a lot of sense.

A few things I thought were really nice:
- the use of recording rules to standardize anomaly inputs via `anomaly_name`
- the hybrid approach (margin-based + stddev-based), which handles both low and high variance metrics
- the separation between monotonic transformations and anomaly logic

That said, I think there are a few limitations we could explore further.

---

### 1. Anomaly definition is purely local (per-metric)

Right now, anomalies are defined as deviations from a local baseline (mean/stddev of the same time series).

This works well for:
- spikes
- sudden drops

But I think it might not capture:
- coordinated shifts across nodes
- gradual regime changes
- network-wide degradation

For example:
If all nodes lose ~20% of inbound connections at the same time, the baseline would adapt and we might not flag anything.

---

### 2. No cross-node or cross-metric awareness

All anomaly detection is currently univariate.

In practice, I feel like some of the more interesting behaviors might only appear when combining signals:
- drop in peers + drop in addr messages
- traffic imbalance + connection churn

I think we could explore:
- cross-node comparisons (e.g. deviation from cluster median)
- multi-metric signals

---

### 3. Baseline is short-term only

Using `avg_over_time(...[1h])` works well for short-term anomalies, but it might miss:
- daily patterns
- longer-term trends

This could potentially lead to:
- false positives during natural fluctuations
- missed anomalies during slow drifts

---

### 4. Stddev filtering is heuristic

The condition:

stddev > mean * THRESHOLD_BY_COVAR

feels like a practical heuristic, but I’m not sure how stable it is across different metrics, especially since it’s not scale-invariant.

Maybe we could explore:
- normalized scores (like z-score directly)
- or percentile-based thresholds

---

### 5. Alerting is binary and per-series

Currently alerts are:
- per metric
- boolean (inside/outside band)

I wonder if this makes it harder to:
- prioritize anomalies
- aggregate signals across the system

---

### Possible next steps (just ideas)

I think this PR is already a strong base, and maybe we can iterate on top of it.

Some things that could be interesting to explore:

1. A simple cross-node anomaly  
   - comparing each node to something like a cluster median

2. A “relative change” signal  
   - comparing short-term vs longer-term averages

3. Introducing a score instead of just a boolean  
   - for example, distance from the band or from the mean

4. Looking at a few multi-metric scenarios  
   - connection drops  
   - addr message anomalies  
   - traffic distribution shifts  

---

Curious what you think about the cross-node angle. does that align with what you had in mind for future iterations?

If you think it makes sense, I’d be happy to take a pass at implementing a simple version of that first idea (e.g. comparing each node to something like a cluster median) so we can see how it behaves in practice.

Happy to explore this further together.

RazorBest:
@dutenrapha 

Thank you for the feedback!

> coordinated shifts across nodes
gradual regime changes
network-wide degradation

I'd be happy to think more about this if you provide me some concrete examples where this behavior can be observed. I can't think of any.

dutenrapha:
A couple of scenarios came to mind (not saying they’re frequent, but maybe interesting to consider).

I think there are really two different types of problems here:

---

**1. Global shifts across all nodes**

This would include cases like:
- network-wide degradation
- protocol-level changes or behavior shifts
- data collection / instrumentation issues

Mathematically, I think these are all the same kind of problem: most or all nodes move together in the same direction.

To make it concrete, imagine each number below represents the number of inbound connections for a node at a given time:

before:
node A = 100  
node B = 100  
node C = 100  
node D = 100  
node E = 100  

after:
node A = 80  
node B = 80  
node C = 80  
node D = 80  
node E = 80  

---

**Why the current approach might not detect this**

Each node builds its own baseline from recent history.

So for node A:
- previous average ≈ 100  
- new values ≈ 80 → the baseline quickly adapts toward 80  

After a short time:
- mean ≈ 80  
- current value ≈ 80  

→ no strong deviation → no anomaly

The same happens for all nodes.

---

**How a simple alternative could detect it**

If instead we compare short-term vs longer-term behavior:

- short-term average (last 1h) ≈ 80  
- long-term average (last 24h) ≈ 100  

Then:

ratio = 80 / 100 = 0.8  

→ this gives a clear signal that the whole system has shifted downward

So even though each node looks “normal” locally, the system-level change becomes visible.

---

**2. Partial degradation (where cluster comparison might help)**

This seems like a different type of problem: only a subset of nodes diverges from the rest.

Again, using inbound connections as an example:

node A = 100  
node B = 100  
node C = 100  
node D = 70  
node E = 65  

---

**Why the current approach might miss it**

For node D:
- its recent baseline might already have adapted toward ~70  

So:
- mean ≈ 70  
- current value ≈ 70  

→ no deviation → no anomaly

Same idea for node E.

---

**How cluster comparison could detect it**

Now we compare nodes to each other at the same point in time.

Cluster median:

[100, 100, 100, 70, 65] → median = 100  

Now compute deviation:

- node D: 70 - 100 = -30  
- node E: 65 - 100 = -35  

→ both clearly stand out relative to the rest of the cluster

So in this case:
- per-node baseline → might miss it  
- cluster comparison → highlights the divergence  

---

I agree these might not be super common, and maybe the current approach is already enough for most cases.

I was mostly thinking that, since we already have multiple nodes, it could be interesting to experiment with these as complementary signals.

Curious if that resonates with your intuition or if you think these cases are too edge to worry about.

0xB10C:
Hi @RazorBest! I've just read through OP. Thank you for diving into this! Commenting on a few things that came to mind while reading:

- I think what we'd want to ship is `tools/metrics/prometheus/rules/` with a set of prometheus rules. To keep it infrastructure agnostic (I use nix, you docker, someone else just `tmux` to run their stuff), I think `prometheus.yaml` and `tools/metrics/template-compose-yml` aren't needed in a final PR. Buf feel free to leave them in the draft for now.
- Thanks for the backfillings script. That looks handy. I think we can add this to peer-observer and should document it, similar to the way you did it.
- On the infra-library side, we shoudl be able to use `services.prometheus.ruleFiles` and just point it at the files in `tools/metrics/prometheus/rules/`. See https://search.nixos.org/options?channel=25.11&include_modular_service_options=0&include_nixos_options=1&query=services.prometheus.ruleFiles&show=option:services.prometheus.ruleFiles. An alternative could also be `services.prometheus.rules` which accepts a multi-line string of rules. 
- I think I should be able to copy your rule files, add them to my nodes, and run the backfill to see them in action? Will put that on my list for next week. Can give more feedback then!

0xB10C:
I'm running this on the demo nodes now. For future reference, I these were the required commits: https://github.com/0xB10C/nix/commit/33cd754ce70c573b0d2afefce665ea3a07f11ac7, https://github.com/peer-observer/infra-library/commit/145dc467c66d6cd6aa7f4da87edbd7e5885ebc0e, https://github.com/peer-observer/infra-demo/commit/54b2b27c8c504fb17c824445504feee7b5ed355e.

I haven't run the backfilling yet. Will do later. The dashboards can be found here:
- https://demo.peer.observer/monitoring/d/adhjlw5/anomalies?orgId=1&from=now-24h&to=now&timezone=browser&var-HOST=hal
-https://demo.peer.observer/monitoring/d/adtn9gz/anomalies-without-recording-rules?orgId=1&from=now-24h&to=now&timezone=browser&var-THRESHOLD_BY_COVAR=0.5&var-STDDEV_MULTIPLIER=2&var-MARGIN_MULTIPLIER=1.5&var-INCREASE_INTERVAL=1h&var-SHORT_LOOKBACK=1h&var-LONG_LOOKBACK=6h&var-HOST=hal

0xB10C:
On my NixOS host I tried to backfill with:

```
$ pwd
/nix/store/xll0iwkkp8c19pxb5lxln4f3idrx1fx3-peer-observer/prometheus

$ ls
backfill_rules.sh  prometheus.yml  rules

$ END="${END:-$(date +%s)}"
$ START="${START:-$(( ${END} - 345600 ))}" # 4 days ago
$ PROMETHEUS_URL="${PROMETHEUS_URL:-http://127.0.0.1:9090}"

$ sudo -u prometheus mkdir -p /tmp/backfill
$ sudo -u prometheus promtool tsdb create-blocks-from rules \
  --start "${START}" --end "${END}" \
  --url "${PROMETHEUS_URL}" \
  --output-dir /tmp/backfill \
  rules/monotonic.yml rules/anomalies.yml

$ sudo mv /tmp/backfill/* /var/lib/prometheus2/data/
$ sudo chown -R prometheus:prometheus /var/lib/prometheus2/data/
```

This seems to have backfilled the metric, but not the error bars and alerts. Running the backfill took only a few seconds.

grafik

Will revisit backfilling another time.

RazorBest:
Thanks! Yeah, backfilling is pretty unstable. It might work if you run it multiple times again. I think, doing this will also overpopulate the already filled metrics.

I saw that there's a discontinuity in the error bands, at the beginning, here:
discontinuity

The larger band only appears at 12:42, which seems to be about 1 hour since the recording started.
Also, I'm not sure why the yellow highlight doesn't include the part where the metric resets to 0.

0xB10C:
Yeah, I think back-filling is a nice-to-have feature, but it's not too important. Can always just wait a while.

0xB10C:
@RazorBest what do you think is a good way of moving this foward from my side? I guess me looking at the recoding rules would be a good start, but was wondering if you had anything more concrete in mind.

I saw you added "analyze_increase.md", which I think is probably not meant to be merged here.

RazorBest:
Hi @0xB10C! I started looking at one of the plots in your demo, and I took a detour. I was wondering how to look at the recording rules that use `increase()`, and whether the standard deviation changes if we change the query window. That turned into a writeup that analyses some examples. I'm still adding some final touches to that write-up. The quick conclusion is that the query range doesn't affect the standard deviation, so our z-score thresholds can stay as they are. Yeah, I'll delete the analyze_increase.md

The reason the drop to 0 was not caught by the anomaly highlight (the comment above) is because we record the rules with the "> 0" condition. That's a way to ignore counter resets. But I gave it some thought and I think it's a good idea to just remove that condition. The reasons:
- I don't think it's a good enough rule for catching resets. You could have a counter that resets to 0, and then immediately increases to 1 between two sampling events. This will bypass the condition.
- There are cases where you might want to consider counter resets to be anomalies. For example, if the active inbound connections counter drops to 0, it might indicate that you restarted bitcoin-core. But it might also indicate that the network went down. Whether you believe a "network down" event should be considered an anomaly or not is not that important, but I think it's a nice direction to have something that can make that distinction. One way I'm thinking about this is to correlate multiple metrics, which might be possible with PromQL.

> @RazorBest what do you think is a good way of moving this foward from my side? 

Right now, I feel like the bottleneck is at justifying decisions. It's possible to turn this into a mergeable PR without a lot of modifications. But I also don't want the alerts to become noise. I think the decision process can benefit from some subjective perspectives. So, I think, one think that could be helpful for me is if you can look at a few of the anomalies (the yellow highlights), and see if it they make sense for you, or you think they shouldn't be anomalies. I'm more interested in the latter.

RazorBest:
- Changed some of the parameters
- Added a `-5` offset to the lower band to ignore anomalies that use values below 5. Metrics that have such low values don't have much information anyway
- Added 5 examples in the PR description
- Increased the alert wait time from 5m to 30m
- Set the increase interval in transfomations.yml to 30m for all the rules
- Fixed the alerting rule. The expression was using some old names.

I also looked at the discontinuities in the bands. They are normal. They happen whenever the metric becomes more unstable, after 6 hours of stableness (I can provide more info about this, if it's still unclear).

0xB10C:
I feel I have to spend a bit more time and focus on this than I have today, but a bit of feedback on the examples:

- Example 1.: These increases are very squared and don't look natural at all to me. So they might be anomalies. Could they also be artifacts of how the increase is calcuated? edit: on a second thought: It's completly normal for us to sometimes receive 1000 addresses at once! We ask our peers with GETADDR sometimes and then get back up to 1000. So this isn't a real anomaly here. Maybe looking at this over a much longer timeframe (e.g. a week) would show someone starting to spam addr?
- Example 2.1: These don't seem to be anomalies to me, at least not the drop being the anomaly, but possible the increase being the anomaly.
- Example 2.2: So lower THRESHOLD_BY_COVAR might be better here to not to detect 2.1. as an anomaly?
- Example 3.1: Comparing this with the other numbers of closed connections in 30min, it seems to me like this isn't an anomaly either. I think for closed connections an anomaly is when we drop 20, 30, 40, 50 or more inbound connections at the same time maybe? Or if we drop all of our outbound connections at the same time. Maybe a better metric to look at here is the number of connections and not the change in number of connections? I.e. if we are always full with connections (around 125) and then suddently only have 85 or so, then maybe that's an anomaly. I think `peerobserver_conn_inbound_current_anomaly` would be interesting to look at.
- Example 3.2: yeah neither of these are anomalies I think.

Also, feel free to post these as comments. Had to open two tabs side-by-side to avoid scolling back and forth 😄

Should I update the demo deployment with the newest version of this PR?

I also think it makes sense for me to go through some of your picks for metrics and think about what are good first candidates for merging this. E.g. address message _entries_ might not be, but maybe number of addr messages received is? Number of closed connections might not be (due to one connection replacing another), but a drop in current conenctions might be.

Additonally, I feel like we might want to have different settings for different metrics. For example, lower `MARGIN_MULTIPLIER` for the following (which is an anomaly, but caused by me restarting the node).

image


The host field could be a HOST field could be a Grafana variable listing all hosts. We have this in some dashboards. For example in https://demo.peer.observer/monitoring/d/c7065555-68cb-477f-8fbf-367c0a99f104/p2p-p2p-messages-per-node?orgId=1&from=now-7d&to=now&timezone=browser&var-msg=$__all&refresh=1m for p2p messages, but the same can be done for hosts too.

image

0xB10C:
Also, I've been wondering. What is your perfered approach based on the two dashboards? Using the recording rules or don't use them?

My understanding is that if we can do the same without the recoding rules, including alerting on anomalies, and it being a lot easier for us to change the parameters on the fly, we might not even want to do the recoding rules? Am I missing something?

RazorBest:
I appreciate taking time to look through the examples! I'm going to answer the easier questions first.

> Should I update the demo deployment with the newest version of this PR?

Yes

---

> My understanding is that if we can do the same without the recording rules, including alerting on anomalies, and it being a lot easier for us to change the parameters on the fly, we might not even want to do the recoding rules? Am I missing something?

The only difference I'm aware of is that for the on-the-fly dashboards, Prometheus needs to execute the query every time, which is proportional to the query window. Last time I checked, the dashboards were loading pretty fast, with the exception of the LinkingLion dashboard. The worst thing that could happen is that the Prometheus instance behind the public Grafana UI gets DoS-ed (or at least the component that has to answer queries).

Personally, I prefer the on-the-fly dashboards for the reason you already mentioned, and I'm ok with removing the recording rules.

---

> Also, feel free to post these as comments. Had to open two tabs side-by-side to avoid scolling back and forth 😄

Sure

---

> Additonally, I feel like we might want to have different settings for different metrics. For example, lower MARGIN_MULTIPLIER for the following (which is an anomaly, but caused by me restarting the node).

I'll address this from a different perspective: how much would it bother you if we considered resets to be anomalies? In my case, I reset my node about every day, but I'm not bothered by all the anomalies, because they all happen at the same time, and I know why they happen.

I'm asking this because I want to see how far we can go with one set of parameters for all the metrics, and I think it's a trade-off worth taking into account, if it shows that only resets are the exception to the rule.

I'm wondering if we could think about these anomalies using the following black-box mindset: as long as they don't trigger alarm fatigue, they're good enough. For example, we might not want to consider anomalies that are triggered about every hour, not necessarily because they're not anomalies, but because you don't even have time to check all of them.

---

> I also think it makes sense for me to go through some of your picks for metrics and think about what are good first candidates for merging this. E.g. address message entries might not be, but maybe number of addr messages received is? Number of closed connections might not be (due to one connection replacing another), but a drop in current conenctions might be.

I agree.

0xB10C:
Sounds good!

> I'll address this from a different perspective: how much would it bother you if we considered resets to be anomalies? In my case, I reset my node about every day, but I'm not bothered by all the anomalies, because they all happen at the same time, and I know why they happen.

Agree that all resets should be an anomaly, other than them being triggered by a maintaince restart.

I was thinking about an alerting rule: Only alert if connections dropped/reset, but uptime DID NOT also resets.

Uptime (via e.g. `uptime` RPC) is a clear indication of "node just restarted" and I do have a seperate alerting for this already.

---

I'll redeploy soonish and dig through the metrics. Both is on my list.

RazorBest:
> I was thinking about an alerting rule: Only alert if connections dropped/reset, but uptime DID NOT also resets.

That sounds good.

> Example 1.: These increases are very squared and don't look natural at all to me. So they might be anomalies. Could they also be artifacts of how the increase is calcuated? edit: on a second thought: It's completly normal for us to sometimes receive 1000 addresses at once! We ask our peers with GETADDR sometimes and then get back up to 1000. So this isn't a real anomaly here. Maybe looking at this over a much longer timeframe (e.g. a week) would show someone starting to spam addr?

Sharing some thoughts while also exploring the problem:

I'm not exactly sure how a longer timeframe would show someone starting to spam. Is having someone that sends 6000 addrs every 6 hours for 30 days considered spam? What if it sends 1000 every hour? What if it does 36000 every 2 days for the same interval? The effective addr/hour is the same in all 3 cases, but they come in different bursts.

What I can try to do is increase the LONG_LOOKBACK, SHORT_LOOKBACK and INCREASE_INTERVAL, which will result in a larger band.

If the norm is to receive 1000-2000 periodically, I don't see how a new peer that starts sending 1000 addrs periodically would be an anomaly.

Maybe the mean and standard deviation are more stable for a longer period of time (one or more weeks). Then, the mean/stddev of that week could be compared to the mean/stddev of the last weeks.

One issue is that this might depend on the number of active connections. A spammer among 10 peers is easier to detect, than if you had 100 peers. I wonder whether this one is a limitation of the fact that the metric is a sum of all the peers.

This makes me think that we can have 3 types of anomalies:
- short: depending on the last 6 hours
- medium: depending on the last 48 hours
- long: depending on the last 2 weeks

But again, any of the changes I'm thinking about would just result in a larger band, not a "smarter" one.

RazorBest:
Regarding the same metric from Example 1, I saw this spike on May 10th for len.
Screenshot_20260524_215129-1

Apart from that, I think the bands are ok for this one.

It's also interesting to see how it looks with an `INCREASE_INTERVAL` of 24h:
Screenshot_20260524_215732

It's less noisy, so here we might be able to have narrower bands. The spike is still there, but it's smaller and wider.

It gets even more interesting when you set `INCREASE_INTERVAL` to 1w. I think you can observe a long-term trend:
Screenshot_20260524_220425-1

The reason the first half is smaller is probably because the node was down for those couple of days, as the contribution to the metric is additive. 

Each of this plots is emphasizing different types of anomalies (the ones I was mentioning before - short, medium, and long). Looking at this, I think it makes sense to look at the same metric using multiple intervals for the `increase` function.

0xB10C:
> I'll redeploy soonish 

Did that last night. The https://demo.peer.observer/monitoring/d/adhjlw5/anomalies?orgId=1&from=now-24h&to=now&timezone=browser&var-HOST=len dashboard seems to be broken, the https://demo.peer.observer/monitoring/d/adtn9gz/anomalies-without-recording-rules?orgId=1&from=now-24h&to=now&timezone=browser&var-THRESHOLD_BY_COVAR=0.5&var-STDDEV_MULTIPLIER=2&var-MARGIN_MULTIPLIER=1.5&var-INCREASE_INTERVAL=30m&var-SHORT_LOOKBACK=1h&var-LONG_LOOKBACK=6h&var-HOST=hal dashboard seems fine.

> The host field could be a HOST field could be a Grafana variable listing all hosts. We have this in some dashboards. For example in https://demo.peer.observer/monitoring/d/c7065555-68cb-477f-8fbf-367c0a99f104/p2p-p2p-messages-per-node?orgId=1&from=now-7d&to=now&timezone=browser&var-msg=$__all&refresh=1m for p2p messages, but the same can be done for hosts too.

I still think we should add this. Let me know if you want me to do it :)

I'll respond to your other comments soon.

RazorBest:
> Did that last night. 

Thanks, I'll check later why the others are broken.

> I still think we should add this. Let me know if you want me to do it :)

Give me 24 hours - if not, you can do it :D

RazorBest:
I don't know why the "Anomalies" dashboard doesn't work.

I inspected the addrv2 dashboard, and one of the rules in transformations.yml, `peerobserver_p2p_addrv2_addresses_sum_inbound_increase_30m`, is not recorded. The rule's expression is `increase(peerobserver_p2p_addrv2_addresses_sum{direction="inbound"}[30m])`. If you run that expression separately in PromQL, it's fine, the data is there. This is a sign that the rule doesn't even work.

I changed the file name from `monotonic.yml` to `transformations.yml`. Maybe that's the issue?

P.S. I also changed HOST to a dynamic variable, similar to this infra dashboard: https://demo.peer.observer/monitoring/d/y4I1v6fMz/infrastructure-host-resource-usage-overview?orgId=1&from=now-7d&to=now&timezone=browser&var-server=%24__all&refresh=5m)

RazorBest:
In the "Inbound connections" dashboard, I also observed that it takes about 24 hours on both hal and fin to fill up the pool of inbound connections. Is this from a restriction in bitcoin-core, or is it from how the network propagates addrs slowly?

slow_increase

Although, slow addr propagation would not necessarily explain this. If you restart a node, you don't change its IP and port. And, if the node was previously part of the network, its address should had propagated before it restarted.

0xB10C:
Correct, that's just the network opening connections to it. We can assume that these ~40 connections that were there right after restart are actually some kind of spy/monitoring-nodes that try to reconnect when a node goes down.

0xB10C:
It seems I dropped this change from my infrastructure as I was deploying this PR + #420 together.

https://github.com/peer-observer/infra-library/commit/b65487a8d5f868e6cf54f395107154026e274733

Dashboard is now filling up again: https://demo.peer.observer/monitoring/d/adhjlw5/anomalies?orgId=1&from=now-24h&to=now&timezone=browser&var-HOST=hal

tools/metrics/dashboards/anomaly_dashboard.json
```diff
@@ -0,0 +1,1344 @@
...
+          "type": "prometheus",
+          "uid": "P1809F7CD0C75ACF3"
+        },
+        "definition": "label_values(host)",
+        "includeAll": false,
+        "multi": false,
+        "name": "HOST",
+        "options": [],
+        "query": {
+          "query": "label_values(host)",
```
0xB10C, line 1325: one tiny nit: This captures hosts that aren't "nodes" too. For example, host "jude" (Judith Milhon), is the webserver running prometheus/grafana etc. I think we can filter here by using a metric that only hosts expose. 

However, this is fine for now, we'll have to touch the dashboards again at some point, so we can fix it then.

0xB10C:
Going to merge this now. Looks clean and seems to be working. It's a good first step, and we can experiment with which metrics we care about and which not going forward - this doesn't have to happen here.

Thank you @RazorBest for moving this forward and for bearing with me on the reviews! 🧡

PR #412, opened 2026-04-21
Removing ebpf extractor addrman from protobuf definition as well as from tools 

Sansh2356:
- Addresses issue #409 .

PR #415, opened 2026-04-22
Adding lazylock initializer for regtest address in tests

### LINKED ISSUE #413: testing: Avoid generating new wallet addresses on tests
Follow up from @0xB10C in https://github.com/peer-observer/peer-observer/pull/386#discussion_r3043684469

Ref code:

https://github.com/peer-observer/peer-observer/blob/5b63d98be32022836be367cb54b85fd26238e37e/extractors/rpc/tests/integration.rs#L618-L630

Comment:

I recently learned from another project that using an address from the wallet of the node makes tests a lot slower since the wallet spends time processing each transaction to it's addresses. Since we are not spending the UTXOs here, we could use a static regtest address here and in other places. I've used the one on https://bitcoin.stackexchange.com/questions/91222/bech32-addresses-beginning-with-bcrt-on-regtest (bcrt1qs758ursh4q9z627kt3pp5yysm78ddny6txaqgw) in the past.

Sansh2356:
- Fixes #413 .
- Using `LazyLock` instead of `OnceLock` since i think we don't  require passing the initialization params during init and thus be suitable for global thread-safe initialization for regtest address used for spawned nodes in integration tests.

### DIFF
--- a/extractors/log/tests/integration.rs
+++ b/extractors/log/tests/integration.rs
@@ -4,7 +4,7 @@
 use log_extractor::Args;
 use shared::{
     async_nats,
-    bitcoin::{self, Block, consensus::Decodable, hashes::Hash, hex::FromHex},
+    bitcoin::{Block, consensus::Decodable, hashes::Hash, hex::FromHex},
     corepc_node,
     futures::StreamExt,
     log::{Level, LevelFilter, info},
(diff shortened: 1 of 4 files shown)


Sansh2356:
There are clippy warnings in this which are patched in PR #412 and can i add a utils module in `shared/src/testing/utils.rs` for keeping the utilities like this inside that ? cc: @0xB10C and @m4ycon .

0xB10C:
Looks good, thank you @Sansh2356! I didn't know about `LazyLock` before, but this seems to be the right fit here.

As for creating a `shared/src/testing/utils.rs`: Let's keep it as and create a utils next time we add more utils-like functionallity to `mod.rs`.

PR #418, opened 2026-04-25
Adding `anyhow` and replacing natively defined errors

### LINKED ISSUE #404: rust: use anyhow for error handling in applications instead of manual error.rs files
I've not been a fan of the `anyhow` as I've never understood it in use for libraries. However, it's not intended for libraries at all! Applications don't need to propagate fixed error types. Only libraries do.

I think replacing the `error.rs` files and error handling of the extractors and tools with a dependency on `anyhow` should make it a lot less boilerplate code for us to maintain and write. It might simplify our error handling too.

For someone wanting to pick this up, I'd recommend:
- start with a quick PoC of replacing the errors.rs files with anyhow
- anyhow should be a dependency of `shared`
- open a draft PR, it doesn't need to be polished yet. 
- then we can play around with it a bit, let the applications and tools run into errors to see it in action
- and then, if we think it's a net improvement, clean it up and polish it

Sansh2356:
closes #404 by switching to anyhow for error handling of the applications.


Sansh2356:
@0xB10C this is a partial work requiring a directional review from you addressing #404, in this i have not addressed all the `error.rs`  and only for `tools/metrics` .

tools/metrics/src/lib.rs
```diff
@@ -25,12 +25,11 @@ use shared::protobuf::{
 };
 use shared::tokio::sync::watch;
 use shared::util::{self, is_on_linkinglion_banlist};
-use shared::{async_nats, clap};
+use shared::{anyhow, async_nats, clap};
 use std::cmp::{max, min};
 use std::collections::{BTreeMap, BTreeSet, HashMap};
 use std::sync::{Arc, Mutex};
 
-pub mod error;
```
RazorBest, line 35: This also means that `toos/metrics/src/error.rs` can be be deleted. Are we doing this because error.rs only contains some wrappers over other errors, so they don't actually provide new information?

RazorBest (changes requested):
The errors are still printed nicely.

```
cargo run -p metrics --release -- --nats-address 127.0.0.1:8888
...
ERROR [metrics] metrics task failed: IO error: Connection refused (os error 111)
```
```
cargo run -p metrics --release -- --nats-address 127.0.0.1:8888 --nats-password-file sike --nats-username sike
...
ERROR [metrics] metrics task failed: No such file or directory (os error 2)
```

~I also tried running the tests but most of them are failing `NATS_SERVER_BINARY=nats-server cargo test --features nats_integration_tests`. I'm leaving the review here while I'm tracing to trace them.~

Tests are passing

tools/metrics/src/lib.rs
Sansh2356: Yes, i will remove the `error.rs` as well wanted a peer review on changing to `anyhow`, Thanks !

tools/metrics/tests/integration.rs
```diff
@@ -144,10 +144,9 @@ async fn publish_and_check(events: &[Event], subject: Subject, expected: &str) {
             }
 
             let args = make_test_args(nats_server.port, port);
-            match metrics::run(args, shutdown_rx.clone()).await {
-                Ok(_) => break,
-                Err(e) => match e {
-                    RuntimeError::Io(e) => match e.kind() {
+            if let Err(error) = metrics::run(args, shutdown_rx.clone()).await {
+                if let Some(inner_error) = error.downcast_ref::<std::io::Error>() {
+                    match inner_error.kind() {
```
RazorBest: I managed to trigger this branch when testing

tools/metrics/tests/integration.rs
```diff
@@ -201,10 +203,24 @@ async fn test_integration_metrics_no_nats_connection() {
...
+                break;
+            }
+        }
+        if !found_connect_error {
+            panic!(
+                "Expected async_nats::ConnectErrorKind in error chain, got: {}",
+                error
+            );
+        }
+    }
```
RazorBest: I think this should be replaced with something simpler
```suggestion
    assert!(matches!(
        result.unwrap_err().downcast::<async_nats::error::Error<async_nats::ConnectErrorKind>>(),
        Ok(_),
    ));

```

tools/metrics/src/lib.rs
RazorBest: This was also addressed in DMs. A partial conclusion was that the struct defined errors.rs might still be useful, because it shows all the custom messages in one place. I personally think it doesn't bring a lot of value, but other people might find it useful.

tools/metrics/tests/integration.rs
Sansh2356: Yes agreed, have removed the chaining iteration in [`b303356`](https://github.com/peer-observer/peer-observer/pull/418/changes/be03356967e0c5a8f22aaba2b1a5dcc97eba53b2) .

tools/metrics/tests/integration.rs
```diff
@@ -201,10 +203,13 @@ async fn test_integration_metrics_no_nats_connection() {
...
     assert!(matches!(
-        result,
-        Err(metrics::error::RuntimeError::NatsConnect(_))
-    ))
+        result
+            .unwrap_err()
+            .downcast::<async_nats::error::Error<async_nats::ConnectErrorKind>>(),
+        Ok(_),
+    ));
 }
```
0xB10C: See lint failure https://github.com/peer-observer/peer-observer/actions/runs/25002036562/job/73214896052?pr=418

0xB10C (commented):
thanks @Sansh2356 - I still plan to get back to you on this in detail

Sansh2356:
@0xB10C, @RazorBest  i have added anyhow to the rest of the tools and extractors as well, with replacing `bail!` and `context` at certain places where the errors required `params` for better error context .

tools/metrics/tests/integration.rs
Sansh2356: Resolved .

tools/logger/src/error.rs
```diff
@@ -1,70 +0,0 @@
...
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        match self {
-            RuntimeError::SetLogger(e) => write!(f, "set logger error {}", e),
-            RuntimeError::ProtobufDecode(e) => write!(f, "protobuf decode error {}", e),
-            RuntimeError::NatsSubscribe(e) => write!(f, "NATS subscribe error {}", e),
-            RuntimeError::NatsConnect(e) => write!(f, "NATS connection error {}", e),
-            RuntimeError::Io(e) => write!(f, "IO Error {}", e),
-        }
-    }
-}
```
RazorBest, line 28: The errors logged by tools-logger, after some manual testing:

Before:
```
ERROR [logger] logger task failed: NATS connection error DNS error: failed to lookup address information: Temporary failure in name resolution
ERROR [logger] logger task failed: NATS subscribe error invalid subject: invalid subject format
ERROR [logger] logger task failed: protobuf decode error failed to decode Protobuf message: invalid tag value: 0
ERROR [logger] logger task failed: IO Error No such file or directory (os error 2)
```

After
```
ERROR [logger] logger task failed: DNS error: failed to lookup address information: Temporary failure in name resolution
ERROR [logger] logger task failed: invalid subject: invalid subject format
ERROR [logger] caused by: failed to decode Protobuf message: invalid tag value: 0
ERROR [logger] caused by: No such file or directory (os error 2)
```

extractors/p2p/src/lib.rs
```diff
@@ -35,10 +35,9 @@ use std::{
     net::{IpAddr, Ipv4Addr, SocketAddr},
     time::{SystemTime, UNIX_EPOCH},
 };
+// mod error;
 
-mod error;
-
-use error::{BitcoinMsgDecodeError, RuntimeError};
+// use error::{BitcoinMsgDecodeError, RuntimeError};
```
RazorBest: TODO: delete this when the PR is ready to merge

extractors/p2p/src/lib.rs
```diff
@@ -460,15 +459,15 @@ async fn read_and_decode_message<R: AsyncRead + Unpin>(
     reader: &mut BufReader<R>,
     network: BitcoinNetwork,
     addr: &str,
-) -> Result<RawNetworkMessage, BitcoinMsgDecodeError> {
+) -> Result<RawNetworkMessage, Error> {
```
RazorBest: Nit: 
```suggestion
) -> Result<RawNetworkMessage> {
```

extractors/p2p/src/lib.rs
```diff
@@ -460,15 +459,15 @@ async fn read_and_decode_message<R: AsyncRead + Unpin>(
...
     let mut header = [0u8; 24];
     if let Err(e) = reader.read_exact(&mut header).await {
         log::debug!(
             "reading the P2P message header from '{}' failed: {}",
             addr,
             e
         );
-        return Err(BitcoinMsgDecodeError::HeaderReadError(e));
+        bail!("Failed to read message header: {:#?}", e)
```
RazorBest: This is what this looks like now:
```
WARN  [127.0.0.1:34110] error decoding message: Failed to read message header: Custom {
    kind: UnexpectedEof,
    error: "early eof",
}
```

extractors/p2p/src/lib.rs
```diff
@@ -478,10 +477,19 @@ async fn read_and_decode_message<R: AsyncRead + Unpin>(
...
+            .iter()
+            .fold(String::new(), |a, b| a + &format!("{:02x}", b));
+        let expected_bytes = header_magic
+            .iter()
+            .fold(String::new(), |a, b| a + &format!("{:02x}", b));
+        bail!(
+            "Network magic mismatch: got={}, expected={}",
+            required_bytes,
+            expected_bytes
+        );
```
RazorBest, line 466: This is how this error looks now:
```
WARN  [127.0.0.1:43164] error decoding message: Network magic mismatch: got=f9beb4d9, expected=036f59e1
```

It's reversed. Expected should be `f9beb4d9`.

extractors/ebpf/src/lib.rs
```diff
@@ -506,10 +500,12 @@ pub async fn run(args: Args, shutdown_rx: watch::Receiver<bool>) -> Result<(), R
...
                 }
-                // Restarting the bitcoind process can take some time
-                Err(RuntimeError::NoProcessWithPid(_) | RuntimeError::NoPidFile(_)) => {}
                 Err(e) => {
-                    return Err(e);
+                    if e.downcast_ref::<std::io::Error>().is_some()
+                        || e.downcast_ref::<std::num::ParseIntError>().is_some()
+                    {
+                        return Err(e);
+                    }
```
RazorBest, line 533: From my understanding, we can do this because we know what errors `try_get_running_process_pid` returns. But I think this is not future proof, because it's hard to control this invariant that comes from `try_get_running_process_pid`. We don't know what other errors are will be there.

I think this is the only place where the errors defined errors.rs had some additional value other than wrapping the existing errors.

Do you think this problem has an easy solution? I think even string matching the error messages could be fine, if it's simpler.

RazorBest (changes requested):
Hi! Only one concrete change requested. And one place where I'm not sure what we should do.

extractors/p2p/src/lib.rs
Sansh2356: Aah, i forgot will remove the commented lines .

extractors/ebpf/src/lib.rs
Sansh2356: I agree with you on this @RazorBest i think here the native errors makes sense, and i don't think other solution may exist in any case we will have to check for the arising errors and only then we can conditionally proceed .

RazorBest: Checking the errors with `downcast_ref` is fine. That's by design for `anyhow`. The issue is that the previous code was doing a negative check: for `NoProcessWithPid` and `NoPidFile`, it was ignoring the errors, and it was breaking the flow in the case of other errors.

The default path was to break. Now, it's reversed. The default path is to continue.

RazorBest: I'm personally ok with the code being merged like this. It doesn't change the potential functionality too much.

extractors/p2p/src/lib.rs
0xB10C: Ideally, no multiline errors please!

extractors/ebpf/src/lib.rs
```diff
@@ -277,7 +272,7 @@ fn bitcoind_pid(args: &Args) -> Result<i32, RuntimeError> {
         .clone()
         .expect("pid file path should be set");
 
-    let file = File::open(&path).map_err(|e| RuntimeError::NoPidFile((path.clone(), e)))?;
+    let file = File::open(&path).context("could not read pid file from - {path}")?;
```
RazorBest: I tested this and it generates the log message:
```
ERROR [ebpf_extractor] Fatal error during extractor runtime: could not read pid file from - {path}
```

extractors/ebpf/src/lib.rs
RazorBest:
```rust
                Err(e) => {
                    if e.downcast_ref::<std::io::Error>().is_some()
                        || e.downcast_ref::<std::num::ParseIntError>().is_some()
                    {
                        return Err(e);
                    }
```

I think the logic is reversed now. The extractor should continue to run even if the pidfile is not present. But, if the pidfile is not present, the `std::io::Error` is emitted.

This part of the code might be a little confusing. That's why I also kept the comment that explains why we do this:
```
// Restarting the bitcoind process can take some time
```

When bitoin-core stops, the pidfile also gets deleted. We want the ebpf-extractor to wait for the bitcoin-core to restart, and then start the reloading phase. However, in the restart period, there might be a couple of seconds where the pidfile is no yet present.

RazorBest (approved):
Hi! Just one issue.

extractors/ebpf/src/lib.rs
```diff
@@ -501,10 +501,17 @@ pub async fn run(args: Args, shutdown_rx: watch::Receiver<bool>) -> Result<(), E
...
-                    if e.downcast_ref::<std::io::Error>().is_some()
-                        || e.downcast_ref::<std::num::ParseIntError>().is_some()
-                    {
+                    if e.downcast_ref::<std::num::ParseIntError>().is_some() {
                         return Err(e);
+                    } else if e.downcast_ref::<std::io::Error>().is_some() {
+                        if let Some(inner_error) = e.downcast_ref::<std::io::Error>() {
+                            if inner_error.kind() == std::io::ErrorKind::NotFound {
+                                // Restarting the bitcoind process can take some time
+                                continue;
```
RazorBest: We shouldn't continue, because, even in this case, we still have a timeout that's checked right below this if block.

extractors/ebpf/src/lib.rs
```diff
@@ -501,10 +501,17 @@ pub async fn run(args: Args, shutdown_rx: watch::Receiver<bool>) -> Result<(), E
                     has_warned_about_no_events = false;
                 }
                 Err(e) => {
-                    if e.downcast_ref::<std::io::Error>().is_some()
-                        || e.downcast_ref::<std::num::ParseIntError>().is_some()
-                    {
+                    if e.downcast_ref::<std::num::ParseIntError>().is_some() {
                         return Err(e);
+                    } else if e.downcast_ref::<std::io::Error>().is_some() {
+                        if let Some(inner_error) = e.downcast_ref::<std::io::Error>() {
```
RazorBest: `let Some(inner_error)` implies `.is_some()`, which makes it redundant in this case.

extractors/ebpf/src/lib.rs
```diff
@@ -501,10 +501,17 @@ pub async fn run(args: Args, shutdown_rx: watch::Receiver<bool>) -> Result<(), E
...
+                            if inner_error.kind() == std::io::ErrorKind::NotFound {
+                                // Restarting the bitcoind process can take some time
+                                continue;
+                            }
+                        } else {
+                            return Err(e);
+                        }
```
RazorBest: Let's try this instead:
```rust
                Err(e) => {
                    let process_not_found =
                        e.to_string().contains("could not find process with PID");
                    let pidfile_not_found = e.to_string().contains("could not read pid file from");
                    // Restarting the bitcoind process can take some time
                    if !(process_not_found || pidfile_not_found) {
                        return Err(e);
                    }
                }

```

RazorBest (changes requested):

deadmanoz:
Hey @Sansh2356, nice one for picking this up.

I think moving these application-level errors to `anyhow` makes sense. One thing I think we should aim to preserve though is the useful context that some of the deleted enum variants were adding.

In a few places this currently either forwards the raw error with `?`, or turns the underlying error into a string with `bail!("... {e}")`. Could we use `anyhow::Context` in those spots instead? That keeps the extra human-readable context, while still preserving the original source error for downcasting / `ErrorKind` checks.

For example, in `extractors/p2p/src/lib.rs`, `read_and_decode_message()` now does:

```rust
bail!("Failed to read message header: {e}")
```

and the same pattern for the payload read. The previous `BitcoinMsgDecodeError::{HeaderReadError, PayloadReadError}` variants still carried the `std::io::Error` as typed data. With the current `bail!` form, that error gets flattened into the message string. Something like:

```rust
reader
    .read_exact(&mut header)
    .await
    .context("failed to read P2P message header")?;
```

would keep the context while preserving the IO error in the anyhow chain.

For a concrete example, if the header read fails with `early eof`, the old enum and the current PR produce basically the same visible log line:

```text
error decoding message: Failed to read message header: early eof
```

The difference is that the current `bail!("... {e}")` form has flattened the IO error into that string. With `context(...)` and the current `{}` logging, we would only see the outer context:

```text
error decoding message: failed to read P2P message header
```

But with `context(...)` plus `{:#}` at the log site, we keep both the readable context and the source error:

```text
error decoding message: failed to read P2P message header: early eof
```

More broadly, this is not just `BitcoinMsgDecodeError`. The `RuntimeError` enums being deleted across the other extractors and tools (`extractors/rpc`, `extractors/log`, `tools/metrics`, etc.) all impl `source()`, so the underlying IO / NATS / RPC errors were programmatically accessible before via the standard error chain. The `bail!("... {e}")` rewrites flatten that out, whereas `Context` would keep it intact for callers that want to inspect the source.

I don't think this means bringing the custom error enums back. It is more about using `anyhow`'s context model where the old variants were potentially carrying useful context.

Sansh2356:
@deadmanoz thanks for the suggestion, I see what are you pointing out to i will apply the changes by replacing `bail!` by attaching the `context` to concrete error types .

Sansh2356:
@deadmanoz although i think apart from `read_and_decode_message()` other places using `bail!` is fine and it is not imo required to attach `context` because at other places underlying `E` is not there to be preserved in the error-chain.
what do you think about  this ?

deadmanoz:
> @deadmanoz although i think apart from `read_and_decode_message()` other places using `bail!` is fine and it is not imo required to attach `context` because at other places underlying `E` is not there to be preserved in the error-chain. what do you think about this ?

Yeah, agreed. `bail!` doesn't need to be replaced everywhere.

The distinction I am drawing is basically the following. 

If there is no underlying error value to preserve, `bail!` is fine. But where there is a concrete `E` from a fallible call, I would avoid formatting it into the message string with `bail!("... {e}")`, because that turns the typed `E` into plain text. `.context(...)` / `.with_context(...)` feels the right thing to do: it preserves the typed `E` while adding extra context on top.

So things like `could not find the BPF program {name}`, `could not find the BPF map {name}`, or `could not find process with PID {pid}` are fine as `bail!`. There is no source error to preserve in those cases: the `{name}` and `{pid}` interpolations are just descriptive identifiers, not typed errors getting flattened.

The p2p read path is an example where the source matters: `reader.read_exact(...).await` fails with a typed `std::io::Error`, which I think is worth preserving. Something like:

```rust
reader
    .read_exact(&mut header)
    .await
    .with_context(|| format!("failed to read P2P message header from {addr}"))?;
```

Same idea for the payload read. That keeps the human-readable context while preserving the original `std::io::Error`. The current `bail!("Failed to read message header: {e}")` gives a similar log line, but loses the typed value.

So yeah, I think your proposed scope is right: `bail!` for newly-created errors, `Context` where we are wrapping an existing `E`. Admittedly we haven't been using these error types in any meaningful way today, so the practical loss is minimal (though I can imagine it would be useful in expanded testing, for example). But `Context` looks to be anyhow's idiomatic version of this, so we don't have to "throw the baby out with the bathwater" while we shrink the custom-error surface?

Just my 2 cents btw, don't need to act on this now, let's let others chime in with their thoughts!!

Sansh2356:
> > @deadmanoz although i think apart from `read_and_decode_message()` other places using `bail!` is fine and it is not imo required to attach `context` because at other places underlying `E` is not there to be preserved in the error-chain. what do you think about this ?
> 
> Yeah, agreed. `bail!` doesn't need to be replaced everywhere.
> 
> The distinction I am drawing is basically the following.
> 
> If there is no underlying error value to preserve, `bail!` is fine. But where there is a concrete `E` from a fallible call, I would avoid formatting it into the message string with `bail!("... {e}")`, because that turns the typed `E` into plain text. `.context(...)` / `.with_context(...)` feels the right thing to do: it preserves the typed `E` while adding extra context on top.
> 
> So things like `could not find the BPF program {name}`, `could not find the BPF map {name}`, or `could not find process with PID {pid}` are fine as `bail!`. There is no source error to preserve in those cases: the `{name}` and `{pid}` interpolations are just descriptive identifiers, not typed errors getting flattened.
> 
> The p2p read path is an example where the source matters: `reader.read_exact(...).await` fails with a typed `std::io::Error`, which I think is worth preserving. Something like:
> 
> ```rust
> reader
>     .read_exact(&mut header)
>     .await
>     .with_context(|| format!("failed to read P2P message header from {addr}"))?;
> ```
> 
> Same idea for the payload read. That keeps the human-readable context while preserving the original `std::io::Error`. The current `bail!("Failed to read message header: {e}")` gives a similar log line, but loses the typed value.
> 
> So yeah, I think your proposed scope is right: `bail!` for newly-created errors, `Context` where we are wrapping an existing `E`. Admittedly we haven't been using these error types in any meaningful way today, so the practical loss is minimal (though I can imagine it would be useful in expanded testing, for example). But `Context` looks to be anyhow's idiomatic version of this, so we don't have to "throw the baby out with the bathwater" while we shrink the custom-error surface?
> 
> Just my 2 cents btw, don't need to act on this now, let's let others chime in with their thoughts!!

Yeah agreed with this, i had used `Context` at one other place `bitcoind_pid()` specifically where as you mentioned underlying `E` type should be preserved same as the case of `read_and_decode_message()` where it should be preserved due to concrete error type .

Sansh2356:
@0xB10C can you check this whenever free ?

0xB10C:
> @0xB10C can you check this whenever free ?

Yes will do! I was traveling last week and still have to catch up with quite a few things, but this is still on my list to look into. Can you rebase it once to fix the conflict with #427=

tools/metrics/tests/integration.rs
```diff
@@ -160,10 +159,13 @@ async fn publish_and_check(events: &[Event], subject: Subject, expected: &str) {
...
                             *port = new_port;
                         }
-                        _ => panic!("Couldn not start metrics tool: {}", e),
-                    },
-                    _ => panic!("Couldn not start metrics tool: {}", e),
-                },
+                        _ => panic!("Couldn't not start metrics tool: {}", inner_error),
+                    }
+                } else {
+                    panic!("Couldn't not start metrics tool: {}", error);
```
RazorBest: Nit: "Could not" or "Couldn't" in both places.

extractors/ebpf/src/lib.rs
```diff
@@ -507,9 +501,15 @@ pub async fn run(args: Args, shutdown_rx: watch::Receiver<bool>) -> Result<(), R
...
                 Err(e) => {
-                    return Err(e);
+                    if e.downcast_ref::<std::num::ParseIntError>().is_some() {
+                        return Err(e);
+                    }
+                    if let Some(inner_error) = e.downcast_ref::<std::io::Error>() {
+                        if inner_error.kind() != std::io::ErrorKind::NotFound {
+                            return Err(e);
+                        }
+                    }
```
RazorBest: This is an ignore-by-default approach - meaning that we only treat some errors and we continue in the other cases. 

Nit: I thought about this once again and I still think it's better to use a return-by-default approach, instead. That's because, when changing code outside of this (e.g. in the `bitcoind_pid` function), we might silently ignore errors that weren't meant to be ignored.

RazorBest (approved):
Tested again, nothing to add. Just nits.

I agree with deadmanoz on this:
> So yeah, I think your proposed scope is right: bail! for newly-created errors, Context where we are wrapping an existing E.

0xB10C:
Thanks for the rebase! We've now merged some of the long-standing PRs that (at least in my view) had a higher priority. This is one of the next PRs I'll spend time on!

extractors/ebpf/src/lib.rs
```diff
@@ -413,7 +410,7 @@ fn init_bpf_listener<'a, 'b>(
     ))
 }
 
-pub async fn run(args: Args, shutdown_rx: watch::Receiver<bool>) -> Result<(), RuntimeError> {
+pub async fn run(args: Args, shutdown_rx: watch::Receiver<bool>) -> Result<(), Error> {
```
0xB10C: In the other files you use `Result<()>`, why `Result<(), Error>` here?

extractors/ebpf/src/lib.rs
```diff
@@ -491,9 +488,15 @@ pub async fn run(args: Args, shutdown_rx: watch::Receiver<bool>) -> Result<(), R
...
                 }
                 // Restarting the bitcoind process can take some time
-                Err(RuntimeError::NoProcessWithPid(_) | RuntimeError::NoPidFile(_)) => {}
                 Err(e) => {
-                    return Err(e);
+                    if e.downcast_ref::<std::num::ParseIntError>().is_some() {
+                        return Err(e);
+                    }
+                    if let Some(inner_error) = e.downcast_ref::<std::io::Error>() {
+                        if inner_error.kind() != std::io::ErrorKind::NotFound {
```
0xB10C: There's also `e.is::<T>()` instead of `.downcast_ref::<T>().is_some()`

0xB10C (commented):
A few review notes:

- Generally, think looks Ok.
- However, the git history is a bit messy and can be cleaned up. I might do this myself and force-push to avoid multiple round trips.
- We could use context() in a few more places I think.


I'll work on this myself a bit and force-push here in a sec.

extractors/ebpf/src/lib.rs
Sansh2356: Yes i missed this sorry for that, `anyhow::Error` can be removed from return signature here .

Sansh2356:
> A few review notes:
> 
>     * Generally, think looks Ok.
> 
>     * However, the git history is a bit messy and can be cleaned up. I might do this myself and force-push to avoid multiple round trips.
> 
>     * We could use context() in a few more places I think.
> 
> 
> I'll work on this myself a bit and force-push here in a sec.

Agreed, i have messed up git history so sorry for that, i can fix that now .

extractors/ebpf/src/lib.rs
Sansh2356: Yes, can be done like this also .

0xB10C:
> I'll work on this myself a bit and force-push here in a sec.

Alright, I've force-pushed. Each tool/extractor in a separate commit so reviewing is hopefully a bit easier. Here are the changes I made:

- we bail now when we can't initialize the logger in main.rs
- we return `Result<()>` from main so anyhow can handle the pretty printing
- added `.context()?` to the relevant places where we used `?`
- used `{:#}` where appropriate to see the full error chain
- while at it, in the RPC-extractor I refactored the fetching quite a bit so it's less repetitive code for the RPC

I've added you as a Co-Author in these commits, as you did the ground work (thanks!). Maybe I should have mentioned it in the initial issue, that this isn't a small task of just removing the error.rs files..

I think this is in a good state, but needs more review. If @deadmanoz @RazorBest @m4ycon @octaviolucca @Sansh2356 want to test this, feel free to run the extractors and tools in ways they'll error. If you see something that doesn't look right or is missing information in the errors, feel free to leave a comment.

Sansh2356:
> > I'll work on this myself a bit and force-push here in a sec.
> 
> Alright, I've force-pushed. Each tool/extractor in a separate commit so reviewing is hopefully a bit easier. Here are the changes I made:
> 
>     * we bail now when we can't initialize the logger in main.rs
> 
>     * we return `Result<()>` from main so anyhow can handle the pretty printing
> 
>     * added `.context()?` to the relevant places where we used `?`
> 
>     * used `{:#}` where appropriate to see the full error chain
> 
>     * while at it, in the RPC-extractor I refactored the fetching quite a bit so it's less repetitive code for the RPC
> 
> 
> I've added you as a Co-Author in these commits, as you did the ground work (thanks!). Maybe I should have mentioned it in the initial issue, that this isn't a small task of just removing the error.rs files..
> 
> I think this is in a good state, but needs more review. If someone wants to test something, feel free to run the extractors and tools in ways they'll error. If you see something that doesn't look right or is missing information in the errors, feel free to leave a comment.

So sorry that i might have missed the scope of the original issue, thanks for rectifying i will keep it in mind while contributing in future issues @0xB10C .

extractors/ebpf/src/lib.rs
0xB10C: I think this still needs a fix, right?

extractors/ipc/src/lib.rs
0xB10C, line 1: This is in the ebpf-extractor commit by mistake, should be in the ipc-extractor commit.

extractors/rpc/src/lib.rs
```diff
@@ -628,26 +576,19 @@ async fn estimatesmartfee(
 
     for target in BLOCK_TARGETS {
         for mode in MODES {
-            let estimate = measure_rpc_call("estimatesmartfee", metrics, || {
-                rpc_client.estimate_smart_fee_with_mode(target, mode)
-            })?;
-            let estimate = estimate.into_model().unwrap();
```
0xB10C, line 634: Previously, we would have crashed here. Now, we don't.

extractors/rpc/src/metrics.rs
```diff
@@ -50,6 +52,16 @@ impl Metrics {
         )
         .expect("Could not create rpc_fetch_errors_total metric");
 
+        let rpc_model_errors = register_int_counter_vec_with_registry!(
+            Opts::new(
+                "rpc_model_errors_total",
```
0xB10C, line 58: Previously, we wouldn't have put the into_model() errors into prometheus. We'd just log them. Now we do put them into prometheus.

0xB10C:
> The errors are still printed nicely.
> 
> ```
> cargo run -p metrics --release -- --nats-address 127.0.0.1:8888
> ...
> ERROR [metrics] metrics task failed: IO error: Connection refused (os error 111)
> ```


@RazorBest note that the output is now different:

```
$ cargo run -p metrics --release -- --nats-address 127.0.0.1:8888
    Finished `release` profile [optimized] target(s) in 0.07s
     Running `target/release/metrics --nats-address '127.0.0.1:8888'`
2026-06-05T14:08:55.686Z INFO  [main] Starting metrics-server...
2026-06-05T14:08:55.686Z DEBUG [shared::nats_util] Connecting to NATS-server at 127.0.0.1:8888 without authentification
2026-06-05T14:08:55.688Z INFO  [metrics] metrics finished
Error: connecting to NATS at 127.0.0.1:8888

Caused by:
    0: IO error: Connection refused (os error 111)
    1: Connection refused (os error 111)
```

What do you prefer? I somewhat prefer the logging from your comment.

deadmanoz:
> > I'll work on this myself a bit and force-push here in a sec.
> 
> Alright, I've force-pushed. Each tool/extractor in a separate commit so reviewing is hopefully a bit easier. Here are the changes I made:
> 
> * we bail now when we can't initialize the logger in main.rs
> * we return `Result<()>` from main so anyhow can handle the pretty printing
> * added `.context()?` to the relevant places where we used `?`
> * used `{:#}` where appropriate to see the full error chain
> * while at it, in the RPC-extractor I refactored the fetching quite a bit so it's less repetitive code for the RPC
> 
> I've added you as a Co-Author in these commits, as you did the ground work (thanks!). Maybe I should have mentioned it in the initial issue, that this isn't a small task of just removing the error.rs files..
> 
> I think this is in a good state, but needs more review. If @deadmanoz @RazorBest @m4ycon @octaviolucca @Sansh2356 want to test this, feel free to run the extractors and tools in ways they'll error. If you see something that doesn't look right or is missing information in the errors, feel free to leave a comment.

I will re-test over the next few days :)

RazorBest:
@0xB10C 

> What do you prefer? I somewhat prefer the logging from your comment.

I also prefer the first one. The second one could be useful for development/debugging, but I personally don't find it satisfying, because I still don't know what line in the code caused that error - I still have to trace stuff by hand. I'm used to Python's trace style where it shows the line number of each frame in the exception stack. But I think in Rust this can only be done with macros, and I don't think `anyhow` has that capability.

extractors/ebpf/src/lib.rs
```diff
@@ -491,9 +497,15 @@ pub async fn run(args: Args, shutdown_rx: watch::Receiver<bool>) -> Result<(), R
...
                 Err(e) => {
-                    return Err(e);
+                    if e.is::<std::num::ParseIntError>() {
+                        return Err(e);
+                    }
+                    if let Some(inner_error) = e.downcast_ref::<std::io::Error>() {
+                        if inner_error.kind() != std::io::ErrorKind::NotFound {
+                            return Err(e);
+                        }
+                    }
```
RazorBest:
```suggestion
                    let err_msg = format!("{:#}", e);
                    if !err_msg.contains("could not find process with PID")
                        && !err_msg.contains("could not open pid file from")
                    {
                        return Err(e);
                    }
```

I think this is more correct and simple.

RazorBest: This is a place where it was making sense to keep some error structs. But I'm personally ok with string matching.

tools/metrics/src/main.rs
```diff
@@ -1,29 +1,26 @@
...
             log::info!("Received Ctrl+C. Stopping...");
-            let _ = shutdown_tx.send(true);
+            shutdown_tx.send(true).context("sending shutdown signal")
         }
         result = metrics_handle => {
-            match result.unwrap() {
-                Ok(_) => log::info!("metrics task completed."),
-                Err(e) => log::error!("metrics task failed: {e}"),
-            }
+            log::info!("metrics finished");
```
deadmanoz: Nit: this will always be logged no matter what the outcome. So maybe:

```
    let run_result = result.context("metrics runtime")?;
    if run_result.is_ok() {
        log::info!("metrics finished");
    }
    run_result
```

deadmanoz: This impacts all other similar sites too

tools/websocket/src/main.rs
```diff
@@ -19,13 +18,11 @@ async fn main() {
...
             log::info!("Received Ctrl+C. Stopping...");
-            let _ = shutdown_tx.send(true);
+            shutdown_tx.send(true).context("sending shutdown signal")
         }
         result = websocket_handle => {
-            match result.unwrap() {
-                Ok(_) => log::info!("websocket task completed."),
-                Err(e) => log::error!("websocket task failed: {e}"),
-            }
+            log::info!("websocket finished");
```
deadmanoz: Also, always logged, for example

extractors/ipc/src/metrics.rs
```diff
@@ -25,9 +26,9 @@ pub struct Metrics {
 }
 
 impl Metrics {
-    pub fn new() -> Self {
+    pub fn new() -> Result<Self> {
```
deadmanoz, line 29: `rpc` is still just `Self` and then `expect` for metrics, any reason for ipc to change and rpc to not?

deadmanoz (commented):
Some minor non-blocking comments

RazorBest (commented):
Hi! I think I prefer pushing all the messages through logs. Otherwise, for errors that reach `main`, we won't see the timestamp. What I copied in the examples below is exactly how they were displayed on the terminal.

Here are some examples that I tried:

## ebpf

**Process with pid doesn't exist**
```
Error: Fatal error during ebpf-extractor runtime

Caused by:
    could not find process with PID 111111
```
 
**Inexistent bitcoind path**
```
Error: Fatal error during ebpf-extractor runtime

Caused by:
    0: failed to attach USDT
    1: No such file or directory (os error 2)
```

**--bitcoind-path is a directory**
```
Error: Fatal error during ebpf-extractor runtime

Caused by:
    0: failed to attach USDT
    1: Unknown error 4001 (os error 4001)
```

**--bitcoind-path is a text file**
```
Error: Fatal error during ebpf-extractor runtime

Caused by:
    0: failed to attach USDT
    1: Bad file descriptor (os error 9)
```

**--bitcoind-path s the wrong binary (bitcoin-cli)**
```
Error: Fatal error during ebpf-extractor runtime

Caused by:
    0: failed to attach USDT
    1: No such file or directory (os error 2)
```

**the path of --bitcoind-pid-file is inexistent**
```
Error: Fatal error during ebpf-extractor runtime

Caused by:
    0: could not open pid file from test
    1: No such file or directory (os error 2)
```

**--bitcoin-pid-file can't be parsed**
```
Error: Fatal error during ebpf-extractor runtime

Caused by:
    0: parsing the pid from pid file /home/btc/.bitcoin/bitcoin.conf
    1: invalid digit found in string
```

**bitcoind is stopped and not restarted**
No trace because the Err didn't end up in main. This is an expected behaviour, handled entirely by lib.rs.
```
2026-06-08T08:53:45.938Z WARN  [ebpf_extractor] No events received in the last 10s. Is bitcoind or the network down?
2026-06-08T08:53:55.951Z ERROR [ebpf_extractor] No events received in the last 20s.
2026-06-08T08:53:55.951Z WARN  [ebpf_extractor] The bitcoind process might be down, has restarted and changed PIDs, or the network might be down.
2026-06-08T08:53:55.951Z WARN  [ebpf_extractor] The extractor will exit. Please restart it
2026-06-08T08:53:55.951Z DEBUG [ebpf_extractor] Dropped links vector
2026-06-08T08:53:56.738Z DEBUG [ebpf_extractor] Dropped ring buffers
2026-06-08T08:53:56.792Z DEBUG [ebpf_extractor] Dropped loaded skel
```

**Inexistent BPF map**
```
Error: Fatal error during ebpf-extractor runtime

Caused by:
    could not find the BPF map mempool_new_function_inexistent
```

**Insufficient privileges**
```
Error: Fatal error during ebpf-extractor runtime

Caused by:
    0: loading the BPF programs and maps into the kernel
    1: Operation not permitted (os error 1)
```

## log-extractor

**No NATS server**
```
Error: connecting to NATS at 127.0.0.1:4222

Caused by:
    0: IO error: Connection refused (os error 111)
    1: Connection refused (os error 111)
```

## p2p-extractor

**Wrong network**
```
2026-06-08T12:57:36.313Z DEBUG [127.0.0.1:55140] mismatch in the P2P message magic: this could mean the node is on a different network or it's attempting a P2Pv2 connection..
2026-06-08T12:57:36.313Z WARN  [127.0.0.1:55140] error decoding message: Network magic mismatch: got=d3868afd, expected=0b110907
```

**Wrong p2p message**
```
2026-06-08T13:12:58.227Z WARN  [127.0.0.1:47388] error decoding message: decoding the P2P network message: invalid checksum: expected 5df6e0e2, actual 0000000
2026-06-08T13:12:58.227Z INFO  [p2p_extractor] closing connection: '127.0.0.1:47388'
```

Sansh2356:
@0xB10C can i address the suggestions and nits above in a commit ?

0xB10C:
Sure, otherwise I'll get to it soonish.

extractors/ebpf/src/lib.rs
Sansh2356: Acknowledged, thanks.

tools/metrics/src/main.rs
Sansh2356: Acknowledged, thanks .

Sansh2356:
@0xB10C as @deadmanoz mentioned that `Metrics::new` for rpc returns `Self` instead like in ipc which returns `Result<Self>` and instead of excepts that follows attaches `context`, it can similarly be done for rpc also, any specific reason ?

extractors/ipc/src/metrics.rs
0xB10C: yes, good point. This probably was just forgotten

extractors/ipc/src/lib.rs
0xB10C: resolved.

0xB10C:
Now using this for all tools and extractors:

```rust
if let Err(e) = result {
    log::error!("xyz failed: {:#}", e);
    std::process::exit(1);
}
Ok(())
```

which produces something like this again:

```bash
$ cargo run -p metrics --release -- --nats-address 127.0.0.1:8888
2026-06-26T11:30:22.723Z INFO  [main] Starting metrics-server...
2026-06-26T11:30:22.724Z DEBUG [shared::nats_util] Connecting to NATS-server at 127.0.0.1:8888 without authentification
2026-06-26T11:30:22.725Z ERROR [metrics] metrics failed: connecting to NATS at 127.0.0.1:8888: IO error: Connection refused (os error 111): Connection refused (os error 111)
```

Also addresses the remaining comments and cleaned up the ipc-extractor changes from the ebpf-extractor.

This is ready for review again. I think it's close to done and plan to merge next week, unless anything else comes up.

Sansh2356:
> Now using this for all tools and extractors:
> 
> ```rust
> if let Err(e) = result {
>     log::error!("xyz failed: {:#}", e);
>     std::process::exit(1);
> }
> Ok(())
> ```
> 
> which produces something like this again:
> 
> ```shell
> $ cargo run -p metrics --release -- --nats-address 127.0.0.1:8888
> 2026-06-26T11:30:22.723Z INFO  [main] Starting metrics-server...
> 2026-06-26T11:30:22.724Z DEBUG [shared::nats_util] Connecting to NATS-server at 127.0.0.1:8888 without authentification
> 2026-06-26T11:30:22.725Z ERROR [metrics] metrics failed: connecting to NATS at 127.0.0.1:8888: IO error: Connection refused (os error 111): Connection refused (os error 111)
> ```
> 
> Also addresses the remaining comments and cleaned up the ipc-extractor changes from the ebpf-extractor.
> 
> This is ready for review again. I think it's close to done and plan to merge next week, unless anything else comes up.

Thanks @0xB10C .

0xB10C:
Rebased on master to resolve conflict.

0xB10C:
Just did another pass on self-reviewing this. I think it looks good. If there's anything we missed or we find to be annoying, we can iterate on it in a followup. 

Thanks @Sansh2356 for picking this up! And thanks to all the reviewers and testers!

Issue #405, opened 2026-04-18
testing: Reduce and fix intermitting integration test failures


0xB10C:
The peer-observer integration test can intermittently fail due to e.g. different timings and race conditions on different machines running the tests.

While some have been fixed already in e.g. #365, #358, there are still some remaining. We now have #360, which runs the tests multiple times in CI and make it more likely for us to run into an intermittent failure.

To track the ongoing efforts of catching and tracking them, this issue will be the parent issue for concrete reports of intermitting failures going forward.

edilmedeiros:
Probably the report from https://github.com/0xB10C/nix/pull/326#issuecomment-4269257990 counts as another instance of this.

Issue #334, opened 2026-01-06
Add dashboards for new metrics (# 2)


0xB10C:
Similar to #257, add dashboards for recently added metrics:

- [ ] https://github.com/peer-observer/peer-observer/pull/312
- [ ] https://github.com/peer-observer/peer-observer/pull/335
- [ ] https://github.com/peer-observer/peer-observer/pull/321
- [ ] https://github.com/peer-observer/peer-observer/pull/319
- [ ] #340 
- [ ] ...

0xB10C:
from @b10claude

---

I diffed the metrics defined in tools/metrics/src/metrics.rs on current master (84ba89a) against every peerobserver_* name referenced in tools/metrics/dashboards/**/*.json (histograms matched on _bucket/_sum/_count; the recording rules in tools/metrics/prometheus/rules/ only consume metrics that dashboards already use). 114 of 184 metrics have no dashboard. Grouped by the PR that added them:

PRs already listed in this issue

#312 getchaintxstats (none of the 6 have a dashboard)
- [ ] rpc_chaintxstats_tx_count
- [ ] rpc_chaintxstats_tx_rate
- [ ] rpc_chaintxstats_window_block_count
- [ ] rpc_chaintxstats_window_final_block_height
- [ ] rpc_chaintxstats_window_interval
- [ ] rpc_chaintxstats_window_tx_count

#335 2026 LinkingLion IPs
- [ ] rpc_peer_info_list_peers_linkinglion_pre2026 (the existing LinkingLion dashboards only use rpc_peer_info_list_peers_linkinglion)

#321 getblockchaininfo (none of the 10)
- [ ] rpc_blockchaininfo_blocks
- [ ] rpc_blockchaininfo_headers
- [ ] rpc_blockchaininfo_difficulty
- [ ] rpc_blockchaininfo_verification_progress
PRs already listed in this issue

#312 getchaintxstats (none of the 6 have a dashboard)
- [ ] rpc_chaintxstats_tx_count
- [ ] rpc_chaintxstats_tx_rate
- [ ] rpc_chaintxstats_window_block_count
- [ ] rpc_chaintxstats_window_final_block_height
- [ ] rpc_chaintxstats_window_interval
- [ ] rpc_chaintxstats_window_tx_count

#335 2026 LinkingLion IPs
- [ ] rpc_peer_info_list_peers_linkinglion_pre2026 (the existing LinkingLion dashboards only use rpc_peer_inf

[...]

Issue #257, opened 2025-09-16
add dashboards for recently added metrics


0xB10C:
The following PRs recently added metrics and I have a few temp dashboards for them. It would be good to add them here:

- https://github.com/0xB10C/peer-observer/pull/249
- https://github.com/0xB10C/peer-observer/pull/252
- https://github.com/0xB10C/peer-observer/pull/251
- https://github.com/0xB10C/peer-observer/pull/256
- https://github.com/0xB10C/peer-observer/pull/269 (there are dashboards for similar mempool metrics from an external rpc-extractor. These should be adapted for the new internal getmempoolinfo metrics)
- ...

PR #420, opened 2026-04-28
Watching compact block reconstructions via log-extractor

m4ycon:
No, this description wasn't AI generated, so don't simply skip it, it took quite some time :)

Inspired by @0xB10C post on delving [Stats on compact block reconstructions](https://delvingbitcoin.org/t/stats-on-compact-block-reconstructions/1052).

## What this PR does?

- Add log matchers for "Saw new header..." and "Successfully reconstructed block..." logs, which are the "start" and "end" points for measuring time;
- Add prometheus metrics for `log_compact_block_reconstruction_requested_bytes`, `log_compact_block_reconstruction_count` and `log_compact_block_reconstruction_time`.

## What kind of question could this PR answer?

- What is the proportion of compact block reconstructions that do not require transactions?
- What is the average size of requested transactions?
- What is the proportion of compact block reconstructions that happen in high-bandwidth connections?
- Is something affecting the reconstruction time of compact blocks? (anomaly/attacks)

## How reconstruction time is measured?

Bitcoin core outputs some logs when a block arrives, in this PR two of them are important for this, "Saw new header..." and "Successfully reconstructed block..." (with `-debug=cmpctblock` flag).

```
2026-04-06T15:02:09.010720Z Saw new header hash=00000000000000000001405ace6e6e4abd39fd1e13d1b3434468ba99986b6ad5 height=943929 peer=498
...
2026-04-06T15:02:09.404125Z [cmpctblock] Successfully reconstructed block 00000000000000000001405ace6e6e4abd39fd1e13d1b3434468ba99986b6ad5 with 1 txn prefilled, 2714 txn from mempool (incl at least 2 from extra pool) and 4 txn (886 bytes) requested
```

With them we know when the block was first seen and when our node have everything from that block. In that way we can take the diff between the dates and measure the elapsed time, a better measure is done with `-logtimemicros=1` as it will log the timestamp in microseconds.

## How do we know if the reconstruction was via low or high-bandwidth connection?

The heuristic here to know if it's a low or high bandwidth is if the "Saw new header..." has "cmpctblock" ([core ref code](https://github.com/bitcoin/bitcoin/blob/859215218667ca9f35d5adae0289e4a125798087/src/net_processing.cpp#L3543-L3549)). As [BIP-152](https://github.com/bitcoin/bips/blob/master/bip-0152.mediawiki#specification-for-version-1) illustrates, the difference is if we receive first a `cmpctblock` message or `headers`/`inv`, if it's the first one, we are in a high-bandwidth connection.

## Using a gauge metric, how do we know if a new reconstruction happened?

As prometheus is a pull-based monitoring system, we can't just "emit" an event when a reconstruction happens, we'll update a value somewhere (gauge in this case). This generates a problem, what if `log_compact_block_reconstruction_time` measure the same time in a row, how do we know if it was the same reconstruction or a new one?

The solution for this is to use a counter metric, `log_compact_block_reconstruction_count`, that will be incremented every time we measure a reconstruction time, so if we see that the count increased, we know that the time measured is from a new reconstruction. This metric wasn't initially created to solve this problem, but it serves this purpose. It could be the block height but I saw some "zero" values on it (probably when I was messing with infra), so I sticked with a simple counter.

And of course, I consider this pretty rare, even more with the fact that we are measuring the time in microseconds, so the chances of two reconstructions taking the same microsecond time in a row are pretty low.

## How do we measure elapsed time if logs can arrive in different timings or even out of order?

We use a `HashMap` state in metrics tool to store "start" timestamp, "end" timestamp, how many transactions were requested, bandwidth mode and when this record was created. When we see a "Saw new header..." log, we store the "start" timestamp and bandwidth mode. When we see a "Successfully reconstructed block..." log, we store the "end" timestamp and how many transactions were requested.

The key is the block hash, when all data is there (start and end logs arrived), we can calculate the elapsed time and emit some metrics, then we remove this record from the `HashMap`. Also, if a record is there for more than 5 minutes, we remove it as well, this could be common to happen if node doesn't have `-debug=cmpctblock` flag as it will never receive the "end" log.

### DIFF
--- a/extractors/log/tests/integration.rs
+++ b/extractors/log/tests/integration.rs
@@ -15,11 +15,11 @@ use shared::{
         log_extractor::{LogDebugCategory, log},
     },
     simple_logger::SimpleLogger,
-    testing::{REGTEST_ADDRESS, nats_server::NatsServerForTesting},
+    testing::{REGTEST_ADDRESS, nats_server::NatsServerForTesting, node::mine_and_wait_for_tip},
     tokio::{
         self, select,
         sync::watch,
-        time::{Duration, Instant, sleep},
+        time::{Duration, sleep},
     },
 };
 use std::io::Write;
@@ -193,7 +193,7 @@ fn setup_two_connected_nodes(node1_args: Vec<&str>) -> (bitcoind::BitcoinD, bitc
 ///    exit before returning.
 async fn check(
     args: Vec<&str>,
-    test_setup: impl AsyncFn(&bitcoind::Client, &bitcoind::Client),
+    test_setup: fn(&bitcoind::Client, &bitcoind::Client),
     check_event: fn(PeerObserverEvent) -> bool,
 ) {
     setup();
@@ -237,7 +237,7 @@ async fn check(
     write_marker(&log_path_main, &marker);
     let buffered = wait_for_marker(&mut sub, &marker, Duration::from_secs(10)).await;
 
-    test_setup(&node1.client, &node2.client).await;
+    test_setup(&node1.client, &node2.client);
 
     sleep(Duration::from_secs(1)).await;
 
@@ -285,7 +285,7 @@ async fn test_integration_logextractor_log_events() {
 
     check(
         vec![],
-        async |_node1, _node2| {},
+        |_node1, _node2| {},
         |event| match event {
             PeerObserverEvent::LogExtractor(r) => r.log_event.is_some(),
             _ => panic!("unexpected event {:?}", event),
@@ -300,7 +300,7 @@ async fn test_integration_logextractor_unknown_log_events() {
 
     check(
         vec![],
-        async |_node1, _node2| {},
+        |_node1, _node2| {},
         |event| {
             match event {
                 PeerObserverEvent::LogExtractor(r) => {
@@ -326,7 +326,7 @@ async fn test_integration_logextractor_block_connected() {
 
     check(
         vec!["-debug=validation"],
-        async |node1, _node2| {
+        |node1, _node2| {
             node1.generate_to_address(1, &REGTEST_ADDRESS).unwrap();
         },
         |event| {
@@ -354,7 +354,7 @@ async fn test_integration_logextractor_logtimemicros() {
 
     check(
         vec!["-logtimemicros=1"],
-        async |_node1, _node2| {},
+        |_node1, _node2| {},
         |event| {
             match event {
                 PeerObserverEvent::LogExtractor(r) => {
@@ -382,7 +382,7 @@ async fn test_integration_logextractor_extralogging() {
             "-logips=1",
             "-logtimemicros=1",
         ],
-        async |node1, _node2| {
+        |node1, _node2| {
             node1.generate_to_address(1, &REGTEST_ADDRESS).unwrap();
         },
         |event| {
@@ -410,7 +410,7 @@ async fn test_integration_logextractor_block_checked() {
 
     check(
         vec!["-debug=validation"],
-        async |node1, _node2| {
+        |node1, _node2| {
             node1.generate_to_address(1, &REGTEST_ADDRESS).unwrap();
         },
         |event| {
@@ -438,7 +438,7 @@ async fn test_integration_logextractor_mutated_block_bad_witness_nonce_size() {
 
     check(
         vec!["-debug=validation"],
-        async |node1, _node2| {
+        |node1, _node2| {
             let block = node1
                 .generate_block(&REGTEST_ADDRESS.to_string(), &[], false)
                 .unwrap();
@@ -486,7 +486,7 @@ async fn test_integration_logextractor_mutated_block_bad_txnmrklroot() {
 
     check(
         vec!["-debug=validation"],
-        async |node1, _node2| {
+        |node1, _node2| {
             let block = node1
                 .generate_block(&REGTEST_ADDRESS.to_string(), &[], false)
                 .unwrap();
@@ -532,7 +532,7 @@ async fn test_integration_logextractor_unknown_with_threadname() {
 
     check(
         vec!["-logthreadnames=1"],
-        async |_node1, _node2| {},
+        |_node1, _node2| {},
         |event| {
             match event {
                 PeerObserverEvent::LogExtractor(r) => {
@@ -560,7 +560,7 @@ async fn test_integration_logextractor_unknown_with_category() {
 
     check(
         vec!["-debug=net"],
-        async |_node1, _node2| {},
+        |_node1, _node2| {},
         |event| {
             match event {
                 PeerObserverEvent::LogExtractor(r) => {
@@ -587,7 +587,7 @@ async fn test_integration_logextractor_unknown_with_threadname_and_category() {
 
     check(
         vec!["-logthreadnames=1", "-debug=net"],
-        async |_node1, _node2| {},
+        |_node1, _node2| {},
         |event| {
             match event {
                 PeerObserverEvent::LogExtractor(r) => {
@@ -618,7 +618,7 @@ async fn test_integration_logextractor_unknown_with_all_metadata() {
 
     check(
         vec!["-logthreadnames=1", "-logsourcelocations=1", "-debug=net"],
-        async |_node1, _node2| {},
+        |_node1, _node2| {},
         |event| {
             match event {
                 PeerObserverEvent::LogExtractor(r) => {
@@ -650,7 +650,7 @@ async fn test_integration_logextractor_testsshouldtimeout() {
 
     check(
         vec![],
-        async |_node1, _node2| {},
+        |_node1, _node2| {},
         |event| {
             match event {
                 PeerObserverEvent::LogExtractor(_) => {
@@ -670,7 +670,7 @@ async fn test_integration_logextractor_log_bytes() {
 
     check(
         vec!["-debug=net"],
-        async |_node1, _node2| {},
+        |_node1, _node2| {},
         |event| match event {
             PeerObserverEvent::LogExtractor(r) => {
                 if let Some(log::LogEvent::UnknownLogMessage(_)) = r.log_event
@@ -697,16 +697,8 @@ async fn test_integration_logextractor_saw_new_header() {
 
     check(
         vec![],
-        async |node1, node2| {
-            // first chain tip
-            node2.generate_to_address(1, &REGTEST_ADDRESS).unwrap();
-            // wait for node1 to finish IBD
-            let loop_start = Instant::now();
-            while loop_start.elapsed() < Duration::from_secs(30)
-                && node1.get_blockchain_info().unwrap().initial_block_download
-            {
-                sleep(Duration::from_millis(50)).await;
-            }
+        |node1, node2| {
+            mine_and_wait_for_tip(node1, node2);
             // this block should trigger the SawNewHeaderLog
             node2.generate_to_address(1, &REGTEST_ADDRESS).unwrap();
         },
@@ -738,16 +730,8 @@ async fn test_integration_logextractor_compact_block_reconstruction() {
 
     check(
         vec!["-debug=cmpctblock"],
-        async |node1, node2| {
-            // first chain tip
-            node2.generate_to_address(1, &REGTEST_ADDRESS).unwrap();
-            // wait for node1 to finish IBD
-            let loop_start = Instant::now();
-            while loop_start.elapsed() < Duration::from_secs(30)
-                && node1.get_blockchain_info().unwrap().initial_block_download
-            {
-                sleep(Duration::from_millis(50)).await;
-            }
+        |node1, node2| {
+            mine_and_wait_for_tip(node1, node2);
             // this block should trigger the CompactBlockReconstructedLog
             node2.generate_to_address(1, &REGTEST_ADDRESS).unwrap();
         },
--- a/extractors/log/tests/integration.rs
+++ b/extractors/log/tests/integration.rs
@@ -19,7 +19,7 @@ use shared::{
     tokio::{
         self, select,
         sync::watch,
-        time::{Duration, sleep},
+        time::{Duration, Instant, sleep},
     },
 };
 use std::io::Write;
@@ -182,9 +182,9 @@ fn setup_two_connected_nodes(node1_args: Vec<&str>) -> (bitcoind::BitcoinD, bitc
 ///    into `debug.log` and waits for it to arrive via NATS, buffering any
 ///    events consumed before the marker. This proves the full
 ///    `debug.log -> tail -> pipe -> extractor -> NATS` path is live.
-/// 5. Calls `test_setup` against node1's RPC client so the test can trigger
-///    the specific node behaviour it wants to observe (mine a block, submit a
-///    transaction, etc.).
+/// 5. Calls `test_setup` against node1's and node2's RPC client so the test can
+///    trigger the specific node behaviour it wants to observe (mine a block,
+///    submit a transaction, etc.).
 /// 6. Replays buffered pre-marker events through `check_event`, then polls
 ///    live NATS messages. The loop breaks as soon as `check_event` returns
 ///    `true`.
@@ -193,11 +193,11 @@ fn setup_two_connected_nodes(node1_args: Vec<&str>) -> (bitcoind::BitcoinD, bitc
 ///    exit before returning.
 async fn check(
     args: Vec<&str>,
-    test_setup: fn(&bitcoind::Client),
+    test_setup: impl AsyncFn(&bitcoind::Client, &bitcoind::Client),
     check_event: fn(PeerObserverEvent) -> bool,
 ) {
     setup();
-    let (node1, _node2) = setup_two_connected_nodes(args);
+    let (node1, node2) = setup_two_connected_nodes(args);
     let nats_server = NatsServerForTesting::new(&[]).await;
     let nats_port = nats_server.port;
     let (shutdown_tx, shutdown_rx) = watch::channel(false);
@@ -237,7 +237,7 @@ async fn check(
     write_marker(&log_path_main, &marker);
     let buffered = wait_for_marker(&mut sub, &marker, Duration::from_secs(10)).await;
 
-    test_setup(&node1.client);
+    test_setup(&node1.client, &node2.client).await;
 
     sleep(Duration::from_secs(1)).await;
 
@@ -285,7 +285,7 @@ async fn test_integration_logextractor_log_events() {
 
     check(
         vec![],
-        |_node1| (),
+        async |_node1, _node2| {},
         |event| match event {
             PeerObserverEvent::LogExtractor(r) => r.log_event.is_some(),
             _ => panic!("unexpected event {:?}", event),
@@ -300,7 +300,7 @@ async fn test_integration_logextractor_unknown_log_events() {
 
     check(
         vec![],
-        |_node1| (),
+        async |_node1, _node2| {},
         |event| {
             match event {
                 PeerObserverEvent::LogExtractor(r) => {
@@ -326,7 +326,7 @@ async fn test_integration_logextractor_block_connected() {
 
     check(
         vec!["-debug=validation"],
-        |node1| {
+        async |node1, _node2| {
             node1.generate_to_address(1, &REGTEST_ADDRESS).unwrap();
         },
         |event| {
@@ -354,7 +354,7 @@ async fn test_integration_logextractor_logtimemicros() {
 
     check(
         vec!["-logtimemicros=1"],
-        |_node1| {},
+        async |_node1, _node2| {},
         |event| {
             match event {
                 PeerObserverEvent::LogExtractor(r) => {
@@ -382,7 +382,7 @@ async fn test_integration_logextractor_extralogging() {
             "-logips=1",
             "-logtimemicros=1",
         ],
-        |node1| {
+        async |node1, _node2| {
             node1.generate_to_address(1, &REGTEST_ADDRESS).unwrap();
         },
         |event| {
@@ -410,7 +410,7 @@ async fn test_integration_logextractor_block_checked() {
 
     check(
         vec!["-debug=validation"],
-        |node1| {
+        async |node1, _node2| {
             node1.generate_to_address(1, &REGTEST_ADDRESS).unwrap();
         },
         |event| {
@@ -438,7 +438,7 @@ async fn test_integration_logextractor_mutated_block_bad_witness_nonce_size() {
 
     check(
         vec!["-debug=validation"],
-        |node1| {
+        async |node1, _node2| {
             let block = node1
                 .generate_block(&REGTEST_ADDRESS.to_string(), &[], false)
                 .unwrap();
@@ -486,7 +486,7 @@ async fn test_integration_logextractor_mutated_block_bad_txnmrklroot() {
 
     check(
         vec!["-debug=validation"],
-        |node1| {
+        async |node1, _node2| {
             let block = node1
                 .generate_block(&REGTEST_ADDRESS.to_string(), &[], false)
                 .unwrap();
@@ -532,7 +532,7 @@ async fn test_integration_logextractor_unknown_with_threadname() {
 
     check(
         vec!["-logthreadnames=1"],
-        |_node1| {},
+        async |_node1, _node2| {},
         |event| {
             match event {
                 PeerObserverEvent::LogExtractor(r) => {
@@ -560,7 +560,7 @@ async fn test_integration_logextractor_unknown_with_category() {
 
     check(
         vec!["-debug=net"],
-        |_node1| {},
+        async |_node1, _node2| {},
         |event| {
             match event {
                 PeerObserverEvent::LogExtractor(r) => {
@@ -587,7 +587,7 @@ async fn test_integration_logextractor_unknown_with_threadname_and_category() {
 
     check(
         vec!["-logthreadnames=1", "-debug=net"],
-        |_node1| {},
+        async |_node1, _node2| {},
         |event| {
             match event {
                 PeerObserverEvent::LogExtractor(r) => {
@@ -618,7 +618,7 @@ async fn test_integration_logextractor_unknown_with_all_metadata() {
 
     check(
         vec!["-logthreadnames=1", "-logsourcelocations=1", "-debug=net"],
-        |_node1| {},
+        async |_node1, _node2| {},
         |event| {
             match event {
                 PeerObserverEvent::LogExtractor(r) => {
@@ -650,7 +650,7 @@ async fn test_integration_logextractor_testsshouldtimeout() {
 
     check(
         vec![],
-        |_node1| {},
+        async |_node1, _node2| {},
         |event| {
             match event {
                 PeerObserverEvent::LogExtractor(_) => {
@@ -670,7 +670,7 @@ async fn test_integration_logextractor_log_bytes() {
 
     check(
         vec!["-debug=net"],
-        |_node1| {},
+        async |_node1, _node2| {},
         |event| match event {
             PeerObserverEvent::LogExtractor(r) => {
                 if let Some(log::LogEvent::UnknownLogMessage(_)) = r.log_event
@@ -681,6 +681,91 @@ async fn test_integration_logextractor_log_bytes() {
                         "Received log event with log_line_bytes={}",
                         r.log_line_bytes
                     );
+                    return true;
+                }
+                false
+            }
+            _ => panic!("unexpected event {:?}", event),
+        },
+    )
+    .await;
+}
+
+#[tokio::test]
+async fn test_integration_logextractor_saw_new_header() {
+    println!("test that we can parse saw new header logs");
+
+    check(
+        vec![],
+        async |node1, node2| {
+            // first chain tip
+            node2.generate_to_address(1, &REGTEST_ADDRESS).unwrap();
+            // wait for node1 to finish IBD
+            let loop_start = Instant::now();
+            while loop_start.elapsed() < Duration::from_secs(30)
+                && node1.get_blockchain_info().unwrap().initial_block_download
+            {
+                sleep(Duration::from_millis(50)).await;
+            }
+            // this block should trigger the SawNewHeaderLog
+            node2.generate_to_address(1, &REGTEST_ADDRESS).unwrap();
+        },
+        |event| match event {
+            PeerObserverEvent::LogExtractor(r) => {
+                if let Some(log::LogEvent::SawNewHeaderLog(block)) = r.log_event
+                    && r.category == LogDebugCategory::Unknown as i32
+                {
+                    assert!(
+                        !block.block_hash.is_empty(),
+                        "block_hash should not be empty"
+                    );
+                    assert_eq!(block.block_height, 2);
+                    assert!(block.peer_id == 0 || block.peer_id == 1); // in rare cases, peer_id might be 1
+                    info!("SawNewHeaderLog event {}", block);
+                    return true;
+                }
+                false
+            }
+            _ => panic!("unexpected event {:?}", event),
+        },
+    )
+    .await;
+}
+
+#[tokio::test]
+async fn test_integration_logextractor_compact_block_reconstruction() {
+    println!("test that we can parse compact block reconstruction logs");
+
+    check(
+        vec!["-debug=cmpctblock"],
+        async |node1, node2| {
+            // first chain tip
+            node2.generate_to_address(1, &REGTEST_ADDRESS).unwrap();
+            // wait for node1 to finish IBD
+            let loop_start = Instant::now();
+            while loop_start.elapsed() < Duration::from_secs(30)
+                && node1.get_blockchain_info().unwrap().initial_block_download
+            {
+                sleep(Duration::from_millis(50)).await;
+            }
+            // this block should trigger the CompactBlockReconstructedLog
+            node2.generate_to_address(1, &REGTEST_ADDRESS).unwrap();
+        },
+        |event| match event {
+            PeerObserverEvent::LogExtractor(r) => {
+                if let Some(log::LogEvent::CompactBlockReconstructedLog(block)) = r.log_event
+                    && r.category == LogDebugCategory::Cmpctblock as i32
+                {
+                    assert!(
+                        !block.block_hash.is_empty(),
+                        "block_hash should not be empty"
+                    );
+                    assert_eq!(block.prefilled_txn_count, 1);
+                    assert_eq!(block.mempool_txn_count, 0);
+                    assert_eq!(block.extra_pool_txn_count, 0);
+                    assert_eq!(block.requested_txn_count, 0);
+                    assert_eq!(block.requested_txn_bytes, 0);
+                    info!("CompactBlockReconstructedLog event {}", block);
                     return true;
                 }
                 false
--- a/protobuf/log_extractor.proto
+++ b/protobuf/log_extractor.proto
@@ -20,6 +20,8 @@ message log {
     UnknownLogMessage unknown_log_message = 4;
     BlockConnectedLog block_connected_log = 5;
     BlockCheckedLog block_checked_log = 6;
+    SawNewHeaderLog saw_new_header_log = 9;
+    CompactBlockReconstructedLog compact_block_reconstructed_log = 10;
   }
 }
 
@@ -72,4 +74,23 @@ message BlockCheckedLog {
   required string block_hash = 1;
   required string state = 2;
   required string debug_message = 3;
+}
+
+// 2026-04-06T15:02:09.010720Z Saw new header hash=00000000000000000001405ace6e6e4abd39fd1e13d1b3434468ba99986b6ad5 height=943929 peer=498
+// 2026-04-30T14:42:51.622131Z Saw new cmpctblock header hash=00000000000000000001f143db1e9a20f4ab523c2e9e2367d4780eae09cba1fa height=947303 peer=6709
+message SawNewHeaderLog {
+  required string block_hash = 1;
+  required uint32 block_height = 2;
+  required uint32 peer_id = 3;
+  required bool is_cmpctblock = 4;
+}
+
+// 2026-04-06T15:02:09.404125Z [cmpctblock] Successfully reconstructed block 00000000000000000001405ace6e6e4abd39fd1e13d1b3434468ba99986b6ad5 with 1 txn prefilled, 2714 txn from mempool (incl at least 2 from extra pool) and 4 txn (886 bytes) requested
+message CompactBlockReconstructedLog {
+  required string block_hash = 1;
+  required uint32 prefilled_txn_count = 2;
+  required uint32 mempool_txn_count = 3;
+  required uint32 extra_pool_txn_count = 4;
+  required uint32 requested_txn_count = 5;
+  required uint32 requested_txn_bytes = 6;
 }
--- a/shared/Cargo.toml
+++ b/shared/Cargo.toml
@@ -16,7 +16,7 @@ log = "0.4"
 async-nats = { version = "0.49.0", default-features = false }
 prometheus = "0.14.0"
 lazy_static = "1.5.0"
-tokio = { version = "1.51.0", features = ["rt-multi-thread", "process", "signal"] }
+tokio = { version = "1.51.0", features = ["rt-multi-thread", "process", "signal", "test-util"] }
 tokio-util = { version = "0.7.18", features = ["compat"] }
 futures = "0.3.31"
 rand = "0.10.1"
--- a/shared/src/log_matchers.rs
+++ b/shared/src/log_matchers.rs
@@ -1,6 +1,7 @@
 use crate::protobuf::log_extractor::log::LogEvent;
 use crate::protobuf::log_extractor::{
-    BlockCheckedLog, BlockConnectedLog, Log, LogDebugCategory, LogLevel, UnknownLogMessage,
+    BlockCheckedLog, BlockConnectedLog, CompactBlockReconstructedLog, Log, LogDebugCategory,
+    LogLevel, SawNewHeaderLog, UnknownLogMessage,
 };
 use lazy_static::lazy_static;
 use regex::Regex;
@@ -78,6 +79,18 @@ lazy_static! {
         VALIDATION_STATE_PATTERN
     ))
     .unwrap();
+
+    static ref SAW_NEW_HEADER_REGEX: Regex = Regex::new(&format!(
+        r"Saw new (?:cmpctblock )?header hash=({}) height=(\d+) peer=(\d+)",
+        BLOCK_HASH_PATTERN
+    ))
+    .unwrap();
+
+    static ref COMPACT_BLOCK_RECONSTRUCTED_REGEX: Regex = Regex::new(&format!(
+        r"Successfully reconstructed block ({}) with (\d+) txn prefilled, (\d+) txn from mempool \(incl at least (\d+) from extra pool\) and (\d+) txn \((\d+) bytes\) requested",
+        BLOCK_HASH_PATTERN
+    ))
+    .unwrap();
 }
 
 trait LogMatcher {
@@ -122,6 +135,47 @@ impl LogMatcher for BlockCheckedLog {
     }
 }
 
+impl LogMatcher for SawNewHeaderLog {
+    fn parse_event(line: &str) -> Option<LogEvent> {
+        let caps = SAW_NEW_HEADER_REGEX.captures(line)?;
+
+        let block_hash = caps.get(1)?.as_str().to_string();
+        let block_height = caps.get(2)?.as_str().parse::<u32>().ok()?;
+        let peer_id = caps.get(3)?.as_str().parse::<u32>().ok()?;
+        let is_cmpctblock = line.contains("cmpctblock");
+
+        Some(LogEvent::SawNewHeaderLog(SawNewHeaderLog {
+            block_hash,
+            block_height,
+            peer_id,
+            is_cmpctblock,
+        }))
+    }
+}
+
+impl LogMatcher for CompactBlockReconstructedLog {
+    fn parse_event(line: &str) -> Option<LogEvent> {
+        let caps = COMPACT_BLOCK_RECONSTRUCTED_REGEX.captures(line)?;
+
+        let block_hash = caps.get(1)?.as_str().to_string();
+        let prefilled_txn_count = caps.get(2)?.as_str().parse::<u32>().ok()?;
+        let mempool_txn_count = caps.get(3)?.as_str().parse::<u32>().ok()?;
+        let extra_pool_txn_count = caps.get(4)?.as_str().parse::<u32>().ok()?;
+        let requested_txn_count = caps.get(5)?.as_str().parse::<u32>().ok()?;
+        let requested_txn_bytes = caps.get(6)?.as_str().parse::<u32>().ok()?;
+        Some(LogEvent::CompactBlockReconstructedLog(
+            CompactBlockReconstructedLog {
+                block_hash,
+                prefilled_txn_count,
+                mempool_txn_count,
+                extra_pool_txn_count,
+                requested_txn_count,
+                requested_txn_bytes,
+            },
+        ))
+    }
+}
+
 impl BlockCheckedLog {
     pub fn is_mutated_block(&self) -> bool {
         matches!(
@@ -144,8 +198,12 @@ pub fn parse_log_event(line: &str) -> Log {
         message,
     } = parse_common_log_data(line);
 
-    let matchers: Vec<fn(&str) -> Option<LogEvent>> =
-        vec![BlockConnectedLog::parse_event, BlockCheckedLog::parse_event];
+    let matchers: Vec<fn(&str) -> Option<LogEvent>> = vec![
+        BlockConnectedLog::parse_event,
+        BlockCheckedLog::parse_event,
+        SawNewHeaderLog::parse_event,
+        CompactBlockReconstructedLog::parse_event,
+    ];
     for matcher in &matchers {
         if let Some(event) = matcher(&message) {
             return Log {
--- a/shared/src/log_matchers.rs
+++ b/shared/src/log_matchers.rs
@@ -716,4 +716,47 @@ mod tests {
         let log_event = parse_log_event(log);
         assert_eq!(log_event.log_level, LogLevel::Warning as i32);
     }
+
+    #[test]
+    fn test_log_matcher_saw_new_header() {
+        let log = "2026-04-06T15:02:09.010720Z Saw new header hash=00000000000000000001405ace6e6e4abd39fd1e13d1b3434468ba99986b6ad5 height=943929 peer=498";
+        let log_event = parse_log_event(log);
+
+        assert_eq!(log_event.log_timestamp, 1775487729010720);
+        assert_eq!(log_event.category, LogDebugCategory::Unknown as i32);
+
+        if let Some(LogEvent::SawNewHeaderLog(event)) = log_event.log_event {
+            assert_eq!(
+                event.block_hash,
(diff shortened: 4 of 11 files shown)


tools/metrics/src/lib.rs
```diff
@@ -1463,5 +1476,84 @@ fn handle_log_event(log: &Log, metrics: metrics::Metrics) {
...
+        state.compact_block_reconstruction.remove(&block_hash);
     }
+
+    // TODO: add a test for this, we need to mock "now"
+    // if entries are older than 5 minutes, remove them to avoid unbounded
+    // growth of the hashmap in case of missing events
+    let now = time::OffsetDateTime::now_utc().unix_timestamp();
+    state
+        .compact_block_reconstruction
+        .retain(|_, reconstruction| now <= reconstruction.created_at + 5 * 60);
```
m4ycon: 1. How do you guys would recommend to mock "now"? I was thinking about creating a wrapper class to it 🤔
2. Also, is this the best place to keep a "cleanup" routine? Where should it be?

extractors/log/tests/integration.rs
```diff
@@ -690,3 +690,72 @@ async fn test_integration_logextractor_log_bytes() {
...
+async fn test_integration_logextractor_compact_block_reconstruction() {
+    println!("test that we can parse compact block reconstruction logs");
+
+    check(
+        vec!["-debug=cmpctblock"],
+        |_node1, node2| {
+            node2.generate_to_address(1, &REGTEST_ADDRESS).unwrap(); // this block will trigger SENDCMPCT trade between nodes
+            node2.generate_to_address(1, &REGTEST_ADDRESS).unwrap(); // this block will trigger the actual CompactBlockReconstructedLog
+        },
+        |event| match event {
```
m4ycon, line 738: Both tests here cause intermitent errors and I don't wanna increase #405 list. I think a sleep could solve that but I would prefer to solve this in another way. Sometimes "SENDCMPCT trade" doesn't happen, which is a p2p message that I think I can't trigger it specifically via corepc 🤔.

tools/metrics/dashboards/playlist/compactblock-average-size-of-requested-txs-per-day.json
```diff
@@ -0,0 +1,166 @@
...
+      },
+      "pluginVersion": "12.4.1",
+      "targets": [
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "P1809F7CD0C75ACF3"
+          },
+          "editorMode": "code",
+          "expr": "avg_over_time(\r\n    (\r\n        peerobserver_log_compact_block_reconstruction_requested_bytes\r\n            and\r\n        changes(peerobserver_log_compact_block_reconstruction_requested_bytes[1d]) > 0\r\n    )[1d:]\r\n)",
```
m4ycon: Also, it would be interesting if you guys could review those promql queries, just search for `"expr"` on these `.json` files that you find them.

m4ycon (commented):
Some specific code comments that I would appreciate suggestions/opinions on it.

edilmedeiros:
> As prometheus is a pull-based monitoring system, we can't just "emit" an event when a reconstruction happens, we'll update a value somewhere (gauge in this case).

Just a quick comment about this for future self. I've been planning to experiment with [InfluxDB](https://www.influxdata.com/products/influxdb3-enterprise/) which is a push-based alternative to Prometheus and there are native plugins for Grafana. That would allow us to have more precise timestamps for events, that we have discussed as desirable in a recent call.

edilmedeiros:
> * Second panel line show them aggregated in a histogram;

I would love to have them in a common scale, but not sure this is possible in Grafana since they are independent sources.

m4ycon:
> I would love to have them in a common scale, but not sure this is possible in Grafana since they are independent sources.

I had an idea to use a query to define those min/max values but grafana doesn't let me use them on "soft min" and "soft max" fields, only a static value (those are the ones that define the scales). I don't know if there's another way 🤔.

protobuf/log_extractor.proto
```diff
@@ -73,3 +75,21 @@ message BlockCheckedLog {
   required string state = 2;
   required string debug_message = 3;
 }
+
+// 2026-04-06T15:02:09.010720Z Saw new header hash=00000000000000000001405ace6e6e4abd39fd1e13d1b3434468ba99986b6ad5 height=943929 peer=498
+message SawNewHeaderLog {
+  required string block_hash = 1;
+  required uint32 block_height = 2;
+  required uint32 peer_id = 3;
+  required bool is_cmpctblock = 4;
```
0xB10C, line 85: Maybe add the compact block version of the log message here too just to make readers aware this covers both.

shared/src/log_matchers.rs
```diff
@@ -78,6 +79,18 @@ lazy_static! {
...
     .unwrap();
+
+    static ref SAW_NEW_HEADER_REGEX: Regex = Regex::new(&format!(
+        r"Saw new (?:cmpctblock )?header hash=({}) height=(\d+) peer=(\d+)",
+        BLOCK_HASH_PATTERN
+    ))
+    .unwrap();
+
+    static ref COMPACT_BLOCK_RECONSTRUCTED_REGEX: Regex = Regex::new(&format!(
+        r"Successfully reconstructed block ({}) with (\d+) txn prefilled, (\d+) txn from mempool \(incl at least (\d+) from extra pool\) and (\d+) txn \((\d+) bytes\) requested",
```
0xB10C, line 90: Just noting this here:

I know this is a weird edge-case, but this log message changed with the v30 release of Bitcoin Core. https://github.com/bitcoin/bitcoin/commit/3733ed2dae3d8a6825af9f3dbc6495a238b455da https://github.com/bitcoin/bitcoin/pull/32582

This means, the Bitcoin Knots v29.x and the Bitcoin Core node v29.x I run will give you a different log message. I'm not sure we want to support that going forward - these will be upgraded at some point.

0xB10C:
Gave this a first read, thank you @m4ycon - I think having real-time stats on compact-block reconstructions is very nice. I haven't looked at your comments in detail yet,

> This generates a problem, what if log_compact_block_reconstruction_time measure the same time in a row, how do we know if it was the same reconstruction or a new one?
> The solution for this is to use a counter metric, log_compact_block_reconstruction_count, that will be incremented every time we measure a reconstruction time, so if we see that the count increased, we know that the time measured is from a new reconstruction. 

If I'm not misunderstanding you, I think the Prometheus has a primitive for this https://prometheus.io/docs/practices/histograms/ - does this do the same thing you've implemented?

> Just a quick comment about this for future self. I've been planning to experiment with [InfluxDB](https://www.influxdata.com/products/influxdb3-enterprise/) which is a push-based alternative to Prometheus and there are native plugins for Grafana. That would allow us to have more precise timestamps for events, that we have discussed as desirable in a recent call.

Let's explore a influx tool (in general, not part of this PR!). I agree that Prometheus, while it works well for most counters and gauges, doesn't really work well for individual measurements.


> Average size of requested transactions per day

As you show in the screenshot, is average might hide that there were a bunch of blocks were we had to request 4 MB. Maybe we can also have additional metrics on median, and 25th/75th percentile? Maybe also a histogram?

> First panel line shows each reconstruction measure over time, its averages and the block height;
Second panel line show them aggregated in a histogram;
Third panel line show them in a heatmap. I'm leaning to remove this one;
The last ones are candlesticks showing the distribution of reconstruction times per day, with the mean, 25th and 75th percentiles (open/close) and the min/max values. That's another one I'm leaning to remove but not so much, I want opinions on it.

I agree. Keep the scatterplot, histogram. Drop heatmap. I'm leaning towards keeping the candlestick one too.

Additionally, it might be interesting to have metrics on number of transaction requested an number of transactions we had in our extra pool (when we reject a transaction from mempool, we briefly store it in the extra pool and use this pool when trying to reconstruct a block).

extractors/log/tests/integration.rs
0xB10C: I'm not sure I've understood why you need to generate two blocks. 

Does this test only work if a node is considered high-bandwidth?

extractors/log/tests/integration.rs
```diff
@@ -690,3 +690,72 @@ async fn test_integration_logextractor_log_bytes() {
...
+async fn test_integration_logextractor_saw_new_header() {
+    println!("test that we can parse saw new header logs");
+
+    check(
+        vec![],
+        |_node1, node2| {
+            // it needs 2 blocks to be sent so one of them triggers the
+            // SawNewHeaderLog, maybe it also relates to the compact block flow
+            node2.generate_to_address(1, &REGTEST_ADDRESS).unwrap();
+            node2.generate_to_address(1, &REGTEST_ADDRESS).unwrap();
```
0xB10C: Same as in https://github.com/peer-observer/peer-observer/pull/420/changes/BASE..e2574a96fd5a531e4ca8e4f5d6c07c1c34f172d4#r3161942129 

> I'm not sure I've understood why you need to generate two blocks.
> Does this test only work if a node is considered high-bandwidth?

tools/metrics/src/lib.rs
```diff
@@ -67,13 +67,26 @@ impl Args {
     }
 }
 
+/// Compact block reconstruction state tracking
+#[derive(Clone, Default, Debug)]
+struct CompactBlockReconstruction {
+    started_at_microsec: Option<u64>,
+    ended_at_microsec: Option<u64>,
+    requested_txn_count: u32,
+    bandwidth_mode: Option<String>,
```
0xB10C: Could this be an enum?

- Unknown
- High
- Low

tools/metrics/src/lib.rs
```diff
@@ -67,13 +67,26 @@ impl Args {
...
+    created_at: i64,
```
0xB10C: Can we use an https://doc.rust-lang.org/std/time/struct.Instant.html for this? I think this type is more idiomatic for measuring internal duration.

Edit: I think a `tokio::time::Instant` would be even better. This way we can mock time with `tokio::time::advance(Duration::from_secs(60)).await` and move it forward. See https://docs.rs/tokio/latest/tokio/time/fn.advance.html

shared/src/log_matchers.rs
m4ycon: Hmm that could be a good sample for #285...
> How to organize a LogMatcher to support many implementations of a same log

I just don't know if it's good to fit this here (?), maybe it is.

tools/metrics/src/lib.rs
0xB10C: Instead of using `5 * 60`, use a https://doc.rust-lang.org/std/time/struct.Duration.html. 

> 1. How do you guys would recommend to mock "now"? I was thinking about creating a wrapper class to it 🤔

Use tokio's Instant and then https://docs.rs/tokio/latest/tokio/time/fn.advance.html !

> 2. Also, is this the best place to keep a "cleanup" routine? Where should it be?

I think it's OK here.

extractors/log/tests/integration.rs
m4ycon: It is supposed to work for either but I can only trigger high-bandwidth, and sometimes not even that, maybe nodes are using `headers/inv` trade initially, maybe it take some time to `SENDCMPCT` trade happen (?). Or are nodes low-bandwidth by default and I'm missing something here?

0xB10C: If this happens to be a high-bandwidth relay (you can find out the status via getpeerinfo), then the node1 might first need to tell node2 that node2 should give node blocks in high-bandwidth. We could mock this by sending a `sendcmpct(1)` via the hidden `sendmsgtopeer` RPC (https://github.com/rust-bitcoin/corepc/issues/333) - but I'm not sure if node1 receiving a high-bandwidth compact block without requesting one (sendmsgtopeer is not setting it in node1's state), is going to cause problems.

shared/src/log_matchers.rs
0xB10C: Let's skip it here for now.. I don't think it's too important here.

m4ycon:
Answering @0xB10C 

> If I'm not misunderstanding you, I think the Prometheus has a primitive for this https://prometheus.io/docs/practices/histograms/ - does this do the same thing you've implemented?

Yes, it does. Initially I was using it but after some problems with grafana I decided to abandon it. It was very limiting to do time range filtering in grafana histogram, in the other hand, a gauge did solve this problem and gave the possibility to build a scatterplot one, which I think it's the most cool/interesting one.

> As you show in the screenshot, is average might hide that there were a bunch of blocks were we had to request 4 MB. Maybe we can also have additional metrics on median, and 25th/75th percentile? Maybe also a histogram?

Something similar to the time measuring one, right? Yeah agreed, I even have a hidden query in that panel showing the exact measures, with the intention to debug, but I think it was a signal that it was a bad view for it.

m4ycon:
### Updates:
- Solved the "mock now problem" with `Instant` and `Advance` from `tokio::time`;
- Added sample of "saw new header..." compact block version log in `log_extractor.proto`;
- Added `BandwidthMode` enum;
- Dropped heatmap panel on "Compact block reconstruction time" dashboard;
- Refactored "Average size of requested transactions per day" to "Requested transactions stats":
  - Including size, number of transactions requested and number of transactions in extra pool;
  - For those I thought that it was better to see the sum of a period than the "single" points of each event -- most of them are low values but when a "outlier" appears, it messes up the view;
  - But there's also the panel with each event separated, similar to "Compact block reconstruction time" first panels;
  - The histogram for size of transactions requested also suffers from the same "outlier" problem.

WIP: I'm still working on this [comment](https://github.com/peer-observer/peer-observer/pull/420#discussion_r3162143469) about intermitent errors that I've added in log extractor integration tests.

### Requested transactions stats

Captura de tela_5-5-2026_145039_5 189 165 114

m4ycon:
Solved last pending fix about intermitent errors on log extractor integration tests ([comment](https://github.com/peer-observer/peer-observer/pull/420#discussion_r3156324543)).

In "Saw new..." log case, it happened that if node was still doing IBD, this would only be logged if `-debug=validation` was activated ([core ref](https://github.com/bitcoin/bitcoin/blob/2b7f5914c427f58deb92d52b9ffc4cbee72bc3a3/src/net_processing.cpp#L3548-L3552)), so I wait for IBD to complete.

```rust
async |node1, node2| {
    // first chain tip
    node2.generate_to_address(1, &REGTEST_ADDRESS).unwrap();
    // wait for node1 to finish IBD
    while node1.get_blockchain_info().unwrap().initial_block_download {
      sleep(Duration::from_millis(50)).await;
    }
    // this block should trigger the SawNewHeaderLog
    node2.generate_to_address(1, &REGTEST_ADDRESS).unwrap();
},
```

And for "Succesfully reconstructed...", it seems to be also related to IBD but in a more complex way, it depends on how fast node1 requests for blocks, the first one will arrive as a `BLOCK` message and the node must mark IBD as completed, so it won't have a tip that is too old from the actual one (which is actually none), then it asks for the second block and this one will arrive as a `CMPCTBLOCK` message. Much of it relates to [CanDirectFetch](https://github.com/bitcoin/bitcoin/blob/2b7f5914c427f58deb92d52b9ffc4cbee72bc3a3/src/net_processing.cpp#L1346-L1349) response.

Fail case logs:

```
2026-05-07T14:36:41Z [net] received: getheaders (69 bytes) peer=0
2026-05-07T14:36:41Z [net] getheaders -1 to end from peer=0
2026-05-07T14:36:41Z [net] sending headers (1 bytes) peer=0
2026-05-07T14:36:41Z [net] received: feefilter (8 bytes) peer=0
2026-05-07T14:36:41Z [net] received: feefilter of 0.09936506 BTC/kvB from peer=0
2026-05-07T14:36:41Z [net] received: pong (8 bytes) peer=0
2026-05-07T14:36:41Z [net] received: headers (1 bytes) peer=0
2026-05-07T14:36:41Z [net] pcp: Timeout
2026-05-07T14:36:41Z [net] pcp: Retrying (1)
2026-01-01T00:00:00Z PIPELINE_READY:1778164601876751855
2026-05-07T14:36:41Z [net] received: inv (37 bytes) peer=0
2026-05-07T14:36:41Z [net] got inv: block 1e10c9d427f153f00678a5822ba959c70138105eff9bf2e8d3953fa2c7efe9d3  new peer=0
2026-05-07T14:36:41Z [net] sending getheaders (69 bytes) peer=0
2026-05-07T14:36:41Z [net] getheaders (0) 1e10c9d427f153f00678a5822ba959c70138105eff9bf2e8d3953fa2c7efe9d3 to peer=0
2026-05-07T14:36:41Z [net] received: feefilter (8 bytes) peer=0
2026-05-07T14:36:41Z [net] received: feefilter of 0.00000100 BTC/kvB from peer=0
2026-05-07T14:36:41Z [net] received: headers (82 bytes) peer=0
2026-05-07T14:36:41Z Synchronizing blockheaders, height: 1 (~100.00%)
2026-05-07T14:36:41Z [net] sending sendheaders (0 bytes) peer=0
2026-05-07T14:36:41Z [net] Requesting block 1e10c9d427f153f00678a5822ba959c70138105eff9bf2e8d3953fa2c7efe9d3 (1) peer=0
2026-05-07T14:36:41Z [net] sending getdata (37 bytes) peer=0
2026-05-07T14:36:41Z [net] received: headers (82 bytes) peer=0
2026-05-07T14:36:41Z Synchronizing blockheaders, height: 2 (~100.00%)
2026-05-07T14:36:41Z [net] Requesting block 207db7d3380b6eafcc4afdf3630f6ee0d3c720eeb70109711172057f9ef59fc5 (2) peer=0
2026-05-07T14:36:41Z [net] sending getdata (37 bytes) peer=0
2026-05-07T14:36:41Z [net] received: block (248 bytes) peer=0
2026-05-07T14:36:41Z [net] received block 1e10c9d427f153f00678a5822ba959c70138105eff9bf2e8d3953fa2c7efe9d3 peer=0
2026-05-07T14:36:41Z Leaving InitialBlockDownload (latching to false)
2026-05-07T14:36:41Z UpdateTip: new best=1e10c9d427f153f00678a5822ba959c70138105eff9bf2e8d3953fa2c7efe9d3 height=1 version=0x20000000 log2_work=2.000000 tx=2 date='2026-05-07T14:36:41Z' progress=0.769231 cache=0.3MiB(1txo)
2026-05-07T14:36:41Z [net] sending feefilter (8 bytes) peer=0
2026-05-07T14:36:41Z [net] received: block (248 bytes) peer=0
2026-05-07T14:36:41Z [net] received block 207db7d3380b6eafcc4afdf3630f6ee0d3c720eeb70109711172057f9ef59fc5 peer=0
2026-05-07T14:36:41Z [net] sending sendcmpct (9 bytes) peer=0
2026-05-07T14:36:41Z UpdateTip: new best=207db7d3380b6eafcc4afdf3630f6ee0d3c720eeb70109711172057f9ef59fc5 height=2 version=0x20000000 log2_work=2.584963 tx=3 date='2026-05-07T14:36:42Z' progress=1.000000 cache=0.3MiB(2txo)
2026-05-07T14:36:42Z [net] pcp: Timeout
2026-05-07T14:36:42Z [net] pcp: Retrying (2)
2026-05-07T14:36:43Z [net] pcp: Timeout
2026-05-07T14:36:43Z [net] pcp: Giving up after 3 tries
2026-05-07T14:36:43Z [net] portmap: Could not determine IPv6 default gateway
```

Success case logs:

```
2026-05-07T14:40:01Z [net] received: ping (8 bytes) peer=0
2026-05-07T14:40:01Z [net] sending pong (8 bytes) peer=0
2026-05-07T14:40:01Z [net] received: getheaders (69 bytes) peer=0
2026-05-07T14:40:01Z [net] getheaders -1 to end from peer=0
2026-05-07T14:40:01Z [net] sending headers (1 bytes) peer=0
2026-05-07T14:40:01Z [net] received: feefilter (8 bytes) peer=0
2026-05-07T14:40:01Z [net] received: feefilter of 0.09936506 BTC/kvB from peer=0
2026-05-07T14:40:01Z [net] received: pong (8 bytes) peer=0
2026-05-07T14:40:01Z [net] received: headers (1 bytes) peer=0
2026-01-01T00:00:00Z PIPELINE_READY:1778164802299061689
2026-05-07T14:40:02Z [net] pcp: Timeout
2026-05-07T14:40:02Z [net] pcp: Retrying (1)
2026-05-07T14:40:02Z [net] received: inv (37 bytes) peer=0
2026-05-07T14:40:02Z [net] got inv: block 0f00027dc2cf7d93e9680ed33c15446a7c08a4d2984773beda547e16dda08a92  new peer=0
2026-05-07T14:40:02Z [net] sending getheaders (69 bytes) peer=0
2026-05-07T14:40:02Z [net] getheaders (0) 0f00027dc2cf7d93e9680ed33c15446a7c08a4d2984773beda547e16dda08a92 to peer=0
2026-05-07T14:40:02Z [net] received: feefilter (8 bytes) peer=0
2026-05-07T14:40:02Z [net] received: feefilter of 0.00000100 BTC/kvB from peer=0
2026-05-07T14:40:02Z [net] received: headers (82 bytes) peer=0
2026-05-07T14:40:02Z Synchronizing blockheaders, height: 1 (~100.00%)
2026-05-07T14:40:02Z [net] sending sendheaders (0 bytes) peer=0
2026-05-07T14:40:02Z [net] Requesting block 0f00027dc2cf7d93e9680ed33c15446a7c08a4d2984773beda547e16dda08a92 (1) peer=0
2026-05-07T14:40:02Z [net] sending getdata (37 bytes) peer=0
2026-05-07T14:40:02Z [net] received: block (248 bytes) peer=0
2026-05-07T14:40:02Z [net] received block 0f00027dc2cf7d93e9680ed33c15446a7c08a4d2984773beda547e16dda08a92 peer=0
2026-05-07T14:40:02Z Leaving InitialBlockDownload (latching to false)
2026-05-07T14:40:02Z UpdateTip: new best=0f00027dc2cf7d93e9680ed33c15446a7c08a4d2984773beda547e16dda08a92 height=1 version=0x20000000 log2_work=2.000000 tx=2 date='2026-05-07T14:40:02Z' progress=1.000000 cache=0.3MiB(1txo)
2026-05-07T14:40:02Z [net] sending feefilter (8 bytes) peer=0
2026-05-07T14:40:02Z [net] received: headers (82 bytes) peer=0
2026-05-07T14:40:02Z Saw new header hash=5e40f7271d321566afc89fc9bdfc30f6ba498b563b3572d77e40ceeb5b8885bb height=2 peer=0
2026-05-07T14:40:02Z [net] Requesting block 5e40f7271d321566afc89fc9bdfc30f6ba498b563b3572d77e40ceeb5b8885bb from  peer=0
2026-05-07T14:40:02Z [net] sending getdata (37 bytes) peer=0
2026-05-07T14:40:02Z [net] received: cmpctblock (258 bytes) peer=0
2026-05-07T14:40:02Z [cmpctblock] Initializing PartiallyDownloadedBlock for block 5e40f7271d321566afc89fc9bdfc30f6ba498b563b3572d77e40ceeb5b8885bb using a cmpctblock of 258 bytes
2026-05-07T14:40:02Z [cmpctblock] Initialized PartiallyDownloadedBlock for block 5e40f7271d321566afc89fc9bdfc30f6ba498b563b3572d77e40ceeb5b8885bb using a cmpctblock of 258 bytes
2026-05-07T14:40:02Z [cmpctblock] Successfully reconstructed block 5e40f7271d321566afc89fc9bdfc30f6ba498b563b3572d77e40ceeb5b8885bb with 1 txn prefilled, 0 txn from mempool (incl at least 0 from extra pool) and 0 txn (0 bytes) requested
2026-05-07T14:40:02Z [net] sending sendcmpct (9 bytes) peer=0
2026-05-07T14:40:02Z UpdateTip: new best=5e40f7271d321566afc89fc9bdfc30f6ba498b563b3572d77e40ceeb5b8885bb height=2 version=0x20000000 log2_work=2.584963 tx=3 date='2026-05-07T14:40:03Z' progress=1.000000 cache=0.3MiB(2txo)
```

extractors/log/tests/integration.rs
m4ycon: Solved, explanation in https://github.com/peer-observer/peer-observer/pull/420#issuecomment-4398363940.

extractors/log/tests/integration.rs
m4ycon: I've tested `sendmsgtopeer` but it didn't seem to work, so I went deeper on core code around this and it is solved, explanation in https://github.com/peer-observer/peer-observer/pull/420#issuecomment-4398363940.

And about bandwidth, in my last tests, the second block is relayed via low-band, I guess that a third block would be high-band -- as seen in `success case logs`, the `sendcmpct` message is only sent after the second block.

m4ycon:
I think I've solved the main issues here, I'll mark it as ready for review.

And I would like more opinions about the dashboards, more specifically about the recent one [Requested transactions stats](https://github.com/peer-observer/peer-observer/pull/420#issuecomment-4381682157), maybe it has too much info on it? Do you guys agree with the way I'm showing data? Is there a better one?

And just a reminder, it would be good if reviewers could take a look at dashboard expressions, because of [Using a gauge metric, how do we know if a new reconstruction happened?](https://github.com/peer-observer/peer-observer/pull/420#issue-4345296826).

extractors/log/tests/integration.rs
```diff
@@ -690,3 +690,82 @@ async fn test_integration_logextractor_log_bytes() {
...
+
+    check(
+        vec![],
+        async |node1, node2| {
+            // first chain tip
+            node2.generate_to_address(1, &REGTEST_ADDRESS).unwrap();
+            // wait for node1 to finish IBD
+            while node1.get_blockchain_info().unwrap().initial_block_download {
+                sleep(Duration::from_millis(50)).await;
+            }
```
0xB10C: Good idea, but we should not loop here forever as we'll never see an error in the CI logs for this. I think you can exit after 30 seconds or so of looping here.

m4ycon:
Updates: Rebased with main and prevented infinite verification loop in integration tests with a timeout.

octaviolucca:
Tested locally, log parsing and metrics emission working as expected. Dashboards looking good.

m4ycon:
Rebased with main.

Also, I was about to run `cargo build` to check if I didn't break anything before `push -f`, but IPC extractor wasn't building because I didn't have its libs installed, I took a look at its readme and there wasn't anything saying that I was supposed to run `sudo apt-get install libcapnp-dev capnproto` to get it to work. Which leads me to think if that kind of information is really meant to be **implicit** -- same for ebpf extractor. Any thoughts on this @0xB10C? Maybe that's not the better place to talk about it, a issue maybe, idk...

0xB10C:
I'll have a look at this in details soon. It might also make sense to deploy on the demo hosts for now.

> Also, I was about to run cargo build to check if I didn't break anything before push -f, but IPC extractor wasn't building because I didn't have its libs installed, I took a look at its readme and there wasn't anything saying that I was supposed to run sudo apt-get install libcapnp-dev capnproto to get it to work. Which leads me to think if that kind of information is really meant to be implicit -- same for ebpf extractor. Any thoughts on this @0xB10C? Maybe that's not the better place to talk about it, a issue maybe, idk...

Maybe having a setup documentation for common OS is reasonable, yes. A place that might give a hint on what's needed is what the CI is installing on each run. I feel a docs section might get outdated fast and I'm not sure yet how to best check this yet. Could we improve the documentation on what the CI installs and why, and then link to it?

m4ycon:
> Maybe having a setup documentation for common OS is reasonable, yes. A place that might give a hint on what's needed is what the CI is installing on each run. I feel a docs section might get outdated fast and I'm not sure yet how to best check this yet. Could we improve the documentation on what the CI installs and why, and then link to it?

I agree that creating two places with an information that must be synced is pretty bad and hard to maintain. CI is a good source, we don't have to detail too much about it on readmes, a simple link works fine. If necessary we could let a comment in the CI file, after each package and if that package is needed for a specific case we have something pointing that out. e.g.

```bash
        sudo apt-get install -y \
          protobuf-compiler \
          ...
          elfutils \
          gcc-multilib \
          libcapnp-dev \ # ipc-extractor
          capnproto # ipc-extractor
```

extractors/log/tests/integration.rs
```diff
@@ -690,3 +690,88 @@ async fn test_integration_logextractor_log_bytes() {
...
+        vec!["-debug=cmpctblock"],
+        async |node1, node2| {
+            // first chain tip
+            node2.generate_to_address(1, &REGTEST_ADDRESS).unwrap();
+            // wait for node1 to finish IBD
+            let loop_start = Instant::now();
+            while loop_start.elapsed() < Duration::from_secs(30)
+                && node1.get_blockchain_info().unwrap().initial_block_download
+            {
+                sleep(Duration::from_millis(50)).await;
```
RazorBest: Why does this function need to be async? The only `await` is on this line. Is there a different task running while this one sleeps?

tools/metrics/dashboards/playlist/compactblock-reconstruction-time.json
```diff
@@ -0,0 +1,771 @@
...
+          ]
+        },
+        "datasource": {
+          "type": "prometheus",
+          "uid": "P1809F7CD0C75ACF3"
+        },
+        "definition": "label_values(host)",
+        "includeAll": true,
+        "multi": true,
+        "name": "host",
```
RazorBest: The host variable is not used in the panels. Specifically, in:
- Reconstruction time dispersion (tx requested=true/false)
- Aggregated time elapsed (tx requested=true/false)

tools/metrics/dashboards/playlist/compactblock-reconstruction-time.json
```diff
@@ -0,0 +1,771 @@
...
+          "range": true,
+          "refId": "dots"
+        },
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "P1809F7CD0C75ACF3"
+          },
+          "editorMode": "code",
+          "expr": "avg_over_time(\r\n  (\r\n    peerobserver_log_compact_block_reconstruction_time{tx_requested=\"$tx_requested\", bandwidth=~\"$bandwidth\"}\r\n      and\r\n    changes(peerobserver_log_compact_block_reconstruction_count{tx_requested=\"$tx_requested\", bandwidth=~\"$bandwidth\"}[1d]) > 0\r\n  )[1d:]\r\n)",
```
RazorBest: (This is related to the screenshot in the review description)

I think the query here should have:
```
peerobserver_log_compact_block_reconstruction_count{...}[$__interval]
```

With a range of `$__interval`, instead of `1d`, just like in the previous query. If Grafana's step is 1m, then that will also be 1m.

I think this would make the avg_over_time more snappy.

RazorBest (commented):
Regarding edimedeiros' comment
> I would love to have them in a common scale, but not sure this is possible in Grafana since they are independent sources.

You can do a hack: plot both variants in both panels, and with Grafana overrides, make the first variant invisible (alpha 0 or thickness 0) in the first panel, and the second variant invisible in the second panel.

---

I looked at a closer example of the average, in the demo dashboard. The yellow dots are low reconstruction times of a node. The orange line is the average over time of those dots. After the first dot appears, the line is horizontal, for a while, which is expected - the average of one value is just that value. However, after the second dot, I was expecting the average to decrease instantly to the mid-value between the y-values of the 2 dots. Instead, it has a gradual decrease. This is directly related to the comment I made about one of the query expressions.
Image

---

Also, here: https://demo.peer.observer/monitoring/d/asdfdgsdfgsdfgsredg/compact-block-requested-transactions-stats?var-per_time=1h&orgId=1&from=2026-05-28T19%3A50%3A37.917Z&to=2026-05-29T03%3A26%3A51.128Z&timezone=browser&refresh=5m

if you look at the last two panels ("Size of requested transactions (without zeroes)" and "Aggregated size of requested transactions (without zeroes)"), there's a 168kb transaction to `len`, but the histogram only goes to 31kB.

---

Regarding this section from your PR description: `Using a gauge metric, how do we know if a new reconstruction happened?`

I think, no matter what we use, we will be bounded by Prometheus' scrape interval. Even if the dashboard uses `peerobserver_log_compact_block_reconstruction_count` to detect new reconstructions, if between two Prometheus pulls, 2 reconstructions happen one after another, `peerobserver_log_compact_block_reconstruction_time` still has only the last value.

extractors/log/tests/integration.rs
m4ycon: I would say that the other task is `node1` completing its IBD, and to avoid an unecessary cpu usage, I added this sleep. It was also done in #449 originated from this. Unfortunatelly this is the unique reason for this callback to be async.

tools/metrics/dashboards/playlist/compactblock-reconstruction-time.json
m4ycon: Good catch! I knew I had left something out, demo observer made it clear.

tools/metrics/dashboards/playlist/compactblock-reconstruction-time.json
m4ycon: It did work well, that's the final query and a result:

```promql
avg_over_time(
  (
    peerobserver_log_compact_block_reconstruction_time{tx_requested="$tx_requested", bandwidth=~"$bandwidth", host=~"$host"}
      and
    changes(peerobserver_log_compact_block_reconstruction_count{tx_requested="$tx_requested", bandwidth=~"$bandwidth", host=~"$host"}[$__interval]) > 0
  )[1d:]
)
```

image

m4ycon:
Rebased with main and made some fixes to the dashboards with @RazorBest comments, thank you for your review!

---

> You can do a hack: plot both variants in both panels, and with Grafana overrides, make the first variant invisible (alpha 0 or thickness 0) in the first panel, and the second variant invisible in the second panel.

Pretty clever! I did know about "override" but I hadn't thought about using it to make things "invisible". I managed to achieve it, with some different/additional overrides, but those generated some strange visualization when we filter a specific query (like in the below image). So I solved it by creating a "proportional" switch (dash variable) that you can use to turn on/off this "invisible plot".

with "proportional" on:

image

problematic case of "proportional" on:

image

and with "proportional" off:

image


---

> if you look at the last two panels ("Size of requested transactions (without zeroes)" and "Aggregated size of requested transactions (without zeroes)"), there's a 168kb transaction to len, but the histogram only goes to 31kB.

About this, I didn't see nothing in the histogram with your link (pre set range), but zooming a little out I did see the same problem. idk exactly why this happens, I did zoom out a bit more and saw the 168kb transaction in the histogram, I think this is related to `Query options`, but I don't know how to refine it correctly to avoid those cases... I also see those type of problems in time series panels, within a time range you see some dots but with a shorter/longer one you don't. If you have any idea, please let me know.

I think I answered everything, thanks again @RazorBest!

extractors/log/tests/integration.rs
m4ycon: Now that I'm reviewing #449 more closely, I see that he did a `std::sleep` without requiring `await`, now I get what you said! 😓 Sorry, I'll still update this code, to use this solution and undo the `async` changes.

EDIT: Done in 7a61539 :)

m4ycon:
Just an update to use the new `mine_and_wait_for_tip` for integration tests (#449).

Another, [rare case of intermittent error](https://github.com/peer-observer/peer-observer/actions/runs/26691652644/job/78669620431?pr=420). I've ran this test specifically more than 100 times in loop before this solution and didn't get this error, but CI on the first run does...

Solved with:

```rs
assert!(block.peer_id == 0 || block.peer_id == 1); // in rare cases, peer_id might be 1
```

0xB10C:
~I'm working on rebasing this (to be able to deploy the most recent version on the demo nodes) and will force-push. Please pull before you make changes! Thanks.~

edit: did not get far and had to take care of something else. Feel free to rebase yourself. I won't force-push here.

m4ycon:
Rebased with master 👍

0xB10C:
https://demo.peer.observer/monitoring/d/addg86j32/compact-block-compact-block-reconstruction-time?orgId=1&from=now-7d&to=now&timezone=browser&var-host=$__all&var-bandwidth=$__all&var-tx_requested=$__all&var-proportional=1&refresh=5m

I've been looking at the dashboards a bit. Noticed three things:

- the query for the host variable picks up the webserver host and we get empty panels: 
image
- with two hosts, this panel is already quite full. the prod version has 15+ nodes, so it might get too full? Is there anything we can do cut this down? maybe don't show dots? only show median? show median with error bars?
image
- "Reconstruction time by day" panels use `avg_over_time` once and `quantile_over_time` multiple times. I don't think we should mix median and average here. Instead of avg_over_time, maybe let's use quantile_over_time(0.5, ... (i.e. median?)

m4ycon:
Agreed with the other ones @0xB10C, but before making the adjustments, just some questions about this:

> with two hosts, this panel is already quite full. the prod version has 15+ nodes, so it might get too full? Is there anything we can do cut this down? maybe don't show dots? only show median? show median with error bars?

Yes, I agree that this will be flooded with dots and probably surpass "Max data points" from grafana's "Query options". And yeah, I think keeping only averages is a good one.

1. Do we hide the dots or effectively remove the query? And after that, is "height" still useful? Maybe we should also remove it (or hide).
2. You mentioned median, do you think median over average is better in this case? IMO average seems to be better 🤔.

0xB10C:
> Do we hide the dots or effectively remove the query? And after that, is "height" still useful? Maybe we should also remove it (or hide).

I don't know - maybe hide it by default (if that's possible). Or remove it, and we'll see if we need it again. Maybe something like this?

image

Might also make sense to have this split by bandwidth AND tx-requested. So essentially have four charts. 

However, I'm not actually sure what you are trying to show with this chart. What do you mean by "dispersion"? Is there a better name for it?

And what does the chart tell us, I'm not sure. It's the top chart, so I assume it's the most important one? If not, what is the most important one?


> You mentioned median, do you think median over average is better in this case? IMO average seems to be better 🤔.

Sure, I think the problem is that I don't yet fully understand the chart. See above. It's not clear to me: What does this show? What happens if there's a big spike (is it a problem)?

tools/metrics/dashboards/playlist/compactblock-reconstruction-time.json
```diff
@@ -0,0 +1,923 @@
...
+            "uid": "P1809F7CD0C75ACF3"
+          },
+          "editorMode": "code",
+          "expr": "max_over_time(peerobserver_log_compact_block_reconstruction_time{tx_requested!=\"false\", bandwidth=\"$bandwidth\", host=\"$host\"}[$__interval])\r\n  and on() (vector($proportional) == 1)",
+          "instant": false,
+          "interval": "",
+          "legendFormat": "max 2",
+          "range": true,
+          "refId": "max !tx_req"
+        }
```
0xB10C: why have `"legendFormat": "max",` (above) and `"legendFormat": "max 2",` here again?

0xB10C:
As mentioned offline, I didn't have the time yet to review this in-depth. It probably requires me to sit down and look at the metrics first, and then go through the dashboard and panels in detail. This requires time and focus, which is a limited resource for all of us.

My general recommendation would be to make this as easy as possible to understand and review. Thinking through the following might help:

- What are the most important things this dashboard should show?
- Have these at the top, make them clean and understandable. Try to avoid a panel with too much dots/information at the top.
- Try to avoid mixing different things, but group by host. For example, showing high-bandwidth and low-bandwidth on the same chart might be confusing and hard to read, as the timings for these differ vastly. However, normally, the timings for high-bandwidth reconstructions without tx-request should be fairly similar across hosts. That would be a good thing to show.

0xB10C:
Maybe a panel with "median block reconstruction time by host" would be the most important one at the top. This one would mix together high- and low-bandwidth and tx-requested and not-requested. The others are for "more information" and drilling down.

0xB10C:
>  It probably requires me to sit down and look at the metrics first, and then go through the dashboard and panels in detail.

For this PR, it might actually have made sense to PR the metrics first, show a minimal PoC dashboard to test it, and then iterate on a dashboard. With +2450 added lines this PR is quite big, and will naturally take longer to be merged.

deadmanoz:
@m4ycon I will get this up and running on one of my VPSs so I can thoroughly review. Apologies it has taken me so long to get around to looking at this!

tools/metrics/dashboards/playlist/compactblock-reconstruction-time.json
m4ycon: Maybe that's not the best name, but this is to allow "proportional" switch be possible. See that one of the labels are `tx_requested!=\"false\"`, the opposite of `max`.

m4ycon:
> I don't know - maybe hide it by default (if that's possible). Or remove it, and we'll see if we need it again. Maybe something like this?

I'll "remove" it, it will be hidden but the query will still exists (`Hide response` from grafana).

> However, I'm not actually sure what you are trying to show with this chart. What do you mean by "dispersion"? Is there a better name for it?

Dispersion, I was thinking more about the dots, but it's just a fancy name, it doesn't add nothing to the understanding. I guess that just removing it already makes it better (i.e. "Reconstruction time" or "Block reconstruction time").

> And what does the chart tell us, I'm not sure. It's the top chart, so I assume it's the most important one? If not, what is the most important one?

I think it is, but the dots were misleading what could draw attention, even more if there were more nodes in the panel.

> > You mentioned median, do you think median over average is better in this case? IMO average seems to be better 🤔.
>
> Sure, I think the problem is that I don't yet fully understand the chart. See above. It's not clear to me: What does this show? What happens if there's a big spike (is it a problem)?

Some strange behaviors happened to me (not sure why, maybe bad vps neighbors), I can test what would have draw our attention more on those cases (median or average). About spikes, well, I think after sometime we can realize what is "normal" or not.

image

> Maybe a panel with "median block reconstruction time by host" would be the most important one at the top. This one would mix together high- and low-bandwidth and tx-requested and not-requested. The others are for "more information" and drilling down.

> Might also make sense to have this split by bandwidth AND tx-requested. So essentially have four charts.

ACK for both.

> For this PR, it might actually have made sense to PR the metrics first, show a minimal PoC dashboard to test it, and then iterate on a dashboard. With +2450 added lines this PR is quite big, and will naturally take longer to be merged.

I don't think the number of added lines is such a problem as ~1700 of it are from dashboards and maybe we only really care for ~100 of it (?), but of course, 850 still big haha. I agree, this is a really good idea, most of the recent discussions are about the dashboards.

I'll push here a simple version of these dashboards (at some point), then create a follow up PR for the "complex" ones. Btw the code isn't supposed to change at all, if anyone wants to review while it.

m4ycon:
Updates:
- Simplified "Compact block reconstruction time", it just shows the average of those measures with no filters (btw median didn't seems to be too different, even for my weird cases);
- Also remove some panels from "Requested transactions stats", there's only number/size of requested txs, and number of txs from extra pool.

I think this is enough to test the PR, probably tomorrow I open other PR (pointing to this one) with the complex/detailed dashboards as a follow up, also with the adjustments pointed by b10c.

deadmanoz:
This is, overwhelmingly, pretty great work!! Nicely done @m4ycon

However, I think there's one minor issue in that the `requested-tx` stats dashboard is using the wrong metric type for its interval totals.

`peerobserver_log_compact_block_reconstruction_requested_bytes`, `peerobserver_log_compact_block_reconstruction_txs_requested`, and `peerobserver_log_compact_block_reconstruction_txs_extra_pool` are registered as `IntGauge`s. This means that each scrape records whichever value was set most recently, not an accumulated interval total.

Then `increase(...[$__interval])` is used on those gauges. [`increase`](https://prometheus.io/docs/prometheus/latest/querying/functions/#increase) calculates the increase over a range, with counter-reset handling. The docs also say that `increase` "should only be used with counters." It does not sum the per-event values in that range; it treats the sample stream as monotonic, with **decreases interpreted as counter resets.** This is decrease-is-effectively-a-reset, is the issue.

Example: suppose four compact-block reconstructions request `777`, `0`, `444`, and `886` bytes. The gauge series stores the current value after each `.set(...)`, while a counter keeps the running total:

```text
gauge:   0 -> 777 -> 0 -> 444 -> 886
counter: 0 -> 777 -> 777 -> 1221 -> 2107
```

The total, given by `increase(counter[...])`, is `2107`. `increase(gauge[...])` sees the `777 -> 0` drop as a counter reset, then counts only the upward steps (`0 -> 777`, `0 -> 444`, `444 -> 886`), which is `1663`, not the sum of the reconstruction values.

I tested this on a synced mainnet node by adding three temporary counters in the same `CompactBlockReconstructedLog` arm and incrementing them from the exact same event fields. After roughly 75 compact-block reconstruction events in the last ~24h, the dashboard-style gauge query was already below the counter-backed value:

| series | `increase(gauge[24h])` | `increase(counter[24h])` | undercount |
|---|---:|---:|---:|
| `requested_bytes` | `210,937.05` | `257,140.94` | `17.97%` |
| `txs_requested` | `751.19` | `838.22` | `10.38%` |
| `txs_extra_pool` | `787.20` | `1,028.26` | `23.44%` |

There were no counter resets in that window, so this is not a restart artefact. Also, here's a screenshot to show what the differences look like:

Screenshot 2026-06-12 at 4 21 18 pm

I think the fix is to make these three aggregate stats counters and `inc_by(...)` them per reconstruction. The existing `log_compact_block_reconstruction_count` metric already has the right counter use for event counts; these requested-tx aggregate fields seem like the same kind of interval-total metric?

m4ycon:
wow, thank you for the lesson/review @deadmanoz 😃! You dissected the problem very well.

Updates ([diff here](https://github.com/peer-observer/peer-observer/compare/1620e3715d12d22a169a3815c5e889b64b8ae6ff..172d30aa9df438468f24080f3982068148661817)):
- `log_compact_block_reconstruction_requested_bytes`, `log_compact_block_reconstruction_txs_requested` and `log_compact_block_reconstruction_txs_extra_pool` were updated to be a counter instead of a gauge, as pointed out in his [comment](https://github.com/peer-observer/peer-observer/pull/420#issuecomment-4689137752), `increase` is not supposed to be used with a gauge.
- Also, simplified "Requested transactions stats" queries.

tools/metrics/src/lib.rs
```diff
@@ -1480,3 +1548,55 @@ fn handle_ipc_event(e: &ipc_extractor::ipc::IpcEvent, metrics: metrics::Metrics)
...
+            ])
+            .set(duration_microsec as i64);
+        state.compact_block_reconstruction.remove(&block_hash);
+    }
+
+    // if entries are older than 5 minutes, remove them to avoid unbounded
+    // growth of the hashmap in case of missing events
+    state
+        .compact_block_reconstruction
+        .retain(|_, reconstruction| reconstruction.created_at.elapsed() < Duration::minutes(5));
```
0xB10C: Think this could be `std::time::Duration::from_mins(5)` then we don't need the re-export in shared.

shared/src/lib.rs
```diff
@@ -13,6 +13,7 @@ pub extern crate prost;
 pub extern crate rand;
 pub extern crate serde;
 pub extern crate simple_logger;
+pub extern crate time;
```
0xB10C: can drop this, if we do https://github.com/peer-observer/peer-observer/pull/420/changes/274462551378547dfc447eb7b6e4e428279910c9#r3485954148

shared/Cargo.toml
```diff
@@ -16,7 +16,7 @@ log = "0.4"
 async-nats = { version = "0.49.0", default-features = false }
 prometheus = "0.14.0"
 lazy_static = "1.5.0"
-tokio = { version = "1.51.0", features = ["rt-multi-thread", "process", "signal"] }
+tokio = { version = "1.51.0", features = ["rt-multi-thread", "process", "signal", "test-util"] }
```
0xB10C, line 19: I think this is added to be able to mock time in the test, correct? 

A short description in the commit message body would be helpful on why this is added and needed!

tools/metrics/tests/integration.rs
```diff
@@ -172,7 +173,13 @@ async fn wait_for_metrics_ready(
     }
 }
 
-async fn publish_and_check(events: &[Event], subject: Subject, expected: &str) {
+async fn start_metrics_service() -> (
```
0xB10C, line 175: would be nice to explain in the commit message why this is refactored

0xB10C:
I've code-reviewed the Rust changes. I haven't looked at the changed dashboards (as I don't have this deployed anywhere) nor re-tested this. If you want, we can split out and merge the Rust code change (once comments are addressed) and then merge this PR with the dashboard. What do you think?

Additionally, I can recommend https://cbea.ms/git-commit/ for have slightly more context (what and why) in commit messages. These make the life of a reviewer easier and result in this getting merged faster!

tools/metrics/src/lib.rs
0xB10C: (feel free to use `use std::time::Duration;` and just write `Duration::from_mins(5);` there).

m4ycon:
Updates:
- Removed `pub extern crate time;` from shared;
- Updated 7ee9cb7 description about `test-util` and why `publish_and_check` was refactored;
- 7ee9cb7 and 77e1033 tries to folllow more cbeams git recommendations.

---

> I've code-reviewed the Rust changes. I haven't looked at the changed dashboards (as I don't have this deployed anywhere) nor re-tested this. If you want, we can split out and merge the Rust code change (once comments are addressed) and then merge this PR with the dashboard. What do you think?

Regarding this, I think I've simplified dashboards enough for a POC as you suggested in a [past comment](https://github.com/peer-observer/peer-observer/pull/420#issuecomment-4669529172) -- they are more straightfoward now. I was thinking about introducing those "complex/detailed" dashboards in a follow-up PR and keeping this one with the minimun or simplier dashs that wouldn't require a deep review.

Do you think that keeping these dashboards here, would require a long "observe window" in your demo nodes? Or maybe a deep review? I think that most, if not all, of the problems that it had people already commented idk 🤔.

0xB10C:
> Do you think that keeping these dashboards here, would require a long "observe window" in your demo nodes? Or maybe a deep review? 

It would be nice to have a look at the dashboards before merging them. While I was running this PR for a while on the demo nodes, it's a lot of work to do this for me on every change here:

- keep an up-to-date version of this PR in a peer-observer nix branch like: https://github.com/0xB10C/nix/tree/2026-06-peer-observer-pr420
- which then needs to be used and kept uptodate in a peer-observer/infra-library branch like this https://github.com/peer-observer/infra-library/tree/2026-06-pr420
- This then needs to be used in the peer-observer/demo-infra in e.g. https://github.com/peer-observer/infra-demo/compare/2026-05-pr400
- then, deploy that change to the nodes

While running this (possibly for multiple weeks to collect data for the dashboards to fill up), I can't run anything else on there, and can't test any other PRs or infrastructure upgrades. While I abuse them as dev nodes, these are actually demo nodes.


Merging the backend changes without the dashboards, would allow me to just run peer-observer master, which does require a lot less of my time to setup and run than running a PR. We can then collect data  and add temporary (as in: not tracked in Git, but stored on Grafana) dashboards there. Once we merged a dashboard, they are "provisioned" and can't be changed on my nodes (provisioning makes them immutable). That's why I've been reviewing and merging backend/metric changes first, then collecting data, and then finalizing the dashboards. See for example https://github.com/peer-observer/peer-observer/issues/257 and https://github.com/peer-observer/peer-observer/issues/334. 

That's why I've been suggesting to split this up and merge the backend first. It would be a big time saver for me, and allow me to get back to more meaning full work than keeping the dev-branches of the infrastructure rebased each time there's a change here...

0xB10C (approved):
Re-reviewed the code. Thanks for the commit message, that makes it clear.

m4ycon:
I see... I didn't imagine the ton of work to setup a single PR on one of yours demo nodes. I think that's the kind of info that would be good to have in a `CONTRIBUTING.md` or maybe a little section on readme, it's a very strong argument in favor of faster review/merges.

That said, I will remove the last commit that adds the dashboards and create a follow up with it.

EDIT: updated description accordingly.

0xB10C:
> I see... I didn't imagine the ton of work to setup a single PR on one of yours demo nodes.

Maybe that's also an area where I can improve. Or maybe it makes sense to have an dev-instance that we can use for something like this.. Maybe not even using nix, so it's faster to switch out commits.

Since last review, this only removed the dashboards. So this is ready to merge! Thanks for hanging in here. Feel free to  open the PR for the dashboards right away. I'll try to deploy the metrics changes tonight or tomorrow so we can start collecting this data.

PR #423, opened 2026-05-05
fix(rpc-extractor): use OS-assigned port in integration tests

deadmanoz:
## Summary

Apply the OS-assigned-port + readiness-oneshot pattern from #358 (p2p) and #365 (websocket) to the rpc-extractor integration tests, closing the rpc-extractor portion of #382.

`shared::metricserver::start` now returns `Result<SocketAddr, io::Error>` so callers can discover the bound port, and `rpc_extractor::run` accepts an optional `oneshot::Sender<SocketAddr>` that fires once the metrics server is listening. The body of `run()` is reordered so RPC client construction and NATS connect happen *before* `metricserver::start`, matching the p2p shape. The readiness barrier now catches RPC and NATS startup failures, not just the metricserver bind. (Deliberate ordering choice, different from #417.)

Tests pick up the p2p-style `tokio::select!` race against the extractor task handle (so startup failures surface as real errors, not `channel closed` panics) in both `check()` and the five `metrics_integration.rs` tests. The old `get_available_port()` helper and the now-unused `prometheus_address` parameter on `make_test_args` are removed.

The check() predicate refactor, state-readiness guards, and `TEST_TIMEOUT_SECONDS` bump that were part of into the original second commit are split out into a follow-up PR (per review feedback).

Fixes #382 (rpc-extractor).

## Testing

rpc-extractor crate stress on NixOS x86_64, 4-core: 20/20 green at `--test-threads=2`, 20/20 green at `--test-threads=4`. Zero `AddrInUse`, zero readiness-barrier panics.

PR #417, opened 2026-04-24
Fix remaining tests that can fail intermittently due to metrics server port reuse

0xB10C:
Closes #382 (rpc-extractor) and fixes it for the metrics tool tests too. This is similar to https://github.com/peer-observer/peer-observer/pull/365 (websocket tool) and https://github.com/peer-observer/peer-observer/pull/358 (p2p extractor).

PR #424, opened 2026-05-05
fix(metrics): use OS-assigned port in integration tests

### LINKED ISSUE #382: rpc-extractor: intermittent failure: "Address already in use"
I saw a few failures of the rpc-extractor integration tests recently due to the port being in use:

```
thread 'test_integration_rpc_getpeerinfo' (61108) panicked at extractors/rpc/tests/integration.rs:67:14:
rpc extractor failed: Io(Os { code: 98, kind: AddrInUse, message: "Address already in use" })
```

- https://github.com/peer-observer/peer-observer/actions/runs/23047602382/job/66940725178
- https://github.com/peer-observer/peer-observer/actions/runs/23047602382/job/66941544256
- https://github.com/peer-observer/peer-observer/actions/runs/23047641364/attempts/1?pr=381

We can fix this with a similar approach as taken in e.g. #365 and #358 by using an OS-assinged port for it.

deadmanoz:
## Summary

Same OS-assigned-port fix as #423 (rpc-extractor), #358, and #365, applied to the `tools/metrics` integration tests. Closes the metrics-tool portion of #382.

The metrics tool subscribes to `*` on NATS and updates Prometheus from whatever the test publisher sends; the test then scrapes `/metrics` and asserts on the output. Port-zero handles the bind race by letting the OS assign the metrics port and returning the actual bound address to the test through a oneshot.

Receiving that bound address only proves the HTTP metrics endpoint is listening. It does not prove the metrics tool's NATS subscription is already receiving publishes from the test publisher's separate NATS connection. The test now uses an end-to-end marker loop: it publishes `Uptime(u32::MAX)` until the corresponding `peerobserver_rpc_uptime 4294967295` line appears in the Prometheus output, then publishes the real test events.

The marker uses `Uptime(u32::MAX)` because it updates a single overwriteable gauge and produces a distinctive line, `peerobserver_rpc_uptime 4294967295`, that is outside any plausible real node uptime. This proves the NATS -> metrics -> Prometheus scrape path is live end-to-end without polluting counters or labelled metrics.

`metrics::run` now connects to NATS, subscribes, starts the metrics server, then sends the bound address, so the address only goes out after every fallible startup step has succeeded. `wait_for_metrics_ready` races the marker loop against the task handle so a post-bind failure surfaces as the real error rather than a generic timeout.

Fixes #382 (metrics tool). Alternative to #417's metrics-tool changes.

## Testing

Stress: 50 consecutive runs of the 56-test integration suite on Linux, all green. Pre-push gates clean.

### DIFF
--- a/tools/metrics/tests/integration.rs
+++ b/tools/metrics/tests/integration.rs
@@ -1,10 +1,9 @@
 #![cfg(feature = "nats_integration_tests")]
 
-use metrics::error::RuntimeError;
 use metrics::Args;
 
 use shared::{
-    log::{debug, warn, Level, LevelFilter},
+    log::{debug, Level, LevelFilter},
     nats_subjects::Subject,
     nats_util::NatsArgs,
     prost::Message,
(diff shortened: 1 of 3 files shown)


0xB10C:
With #423 merged, I rebased this and dropped the first two commits.

0xB10C:
The commit message states:

> The readiness order in run() is connect -> subscribe -> nc.flush() ->
metricserver::start -> send oneshot. Sending the bound address only
after flush() guarantees the SUB has been processed by the NATS server,
so a test that publishes immediately on receipt of the address cannot
beat the subscription. RuntimeError gains a NatsFlush variant for the
new flush() call.

But I don't think the flush() does much (maybe makes the test slightly faster, IDK). My understanding is that the Uptime marker is the actual fix here?

Can you clarify? Also, if we don't really need the flush for the fix, might be best to drop it here to keep the change simpler (also for #418 which will conflict on this).

deadmanoz:
Yeah, you're right. The marker is the actual barrier, the flush is really just giving a fail fast(er) path - it's not essential and can be removed. PR to be updated.

This means a follow up PR to similarly remove from `tools/websocket/src/lib.rs`. The flush seemed like it was part of the solution initially (I think I did the websockets as one of the first fixes?), but it's now apparent it's not necessary.

deadmanoz:
Addressed and rebased, ready for re-review @0xB10C

tools/metrics/tests/integration.rs
```diff
@@ -35,141 +34,174 @@ use shared::{
...
+        let line = line.trim();
+        if line.is_empty() {
+            continue;
+        }
+        if !metrics_raw.contains(line) {
+            return false;
+        }
+    }
+    true
+}
```
0xB10C, line 89: This looks like an unrelated refactor not mentioned in the commit message?

I'd recommend to do these in a extra commit.

0xB10C: (I'm seeing now that you use it below, but it wasn't immediately clear why this refactor is part of this commit, hence the above comment). I think sometimes doing these in a prefactor commit is helpful for reviewers.

deadmanoz: Yeah, the rebase squashed the 3 commits I had. Sorry about that, just looked at the final diff

I'll restore the original 3x commits so it is more legible, and comes through in the mainline history when merged

0xB10C: Alright, I pause my review then and continue once you push. Thanks.

deadmanoz: Hmmm maybe I am mistaken for the earlier form that had a few other things bundled in with this (I don't think I pushed).,,

Anyway, yes I will break this up

0xB10C:
LGTM! Thanks @deadmanoz!

PR #425, opened 2026-05-05
test(rpc-extractor): replace brittle getchaintxstats constants with relational check

deadmanoz:
## Summary

`test_integration_rpc_getchaintxstats` previously pinned `stats.window_interval == Some(34)` and `stats.tx_rate == Some(5.882352941176471)`. Both fields derive from block timestamp deltas in the rolling window, and regtest generates blocks with real wall-clock timestamps, so the values drift with host load. A stress run on a slower host saw `window_interval=38, tx_rate~=5.263`, tripping the pinned values.

Replace them with the semantic invariant: `tx_rate` must equal `window_tx_count / window_interval` within f64 precision, plus `window_interval > 0`. The numerator `window_tx_count` is already asserted at `Some(200)` and is a transaction count rather than a timing derivative, so it is stable across host load.

## Testing

- `cargo fmt -- --check`
- `NATS_SERVER_BINARY=$(which nats-server) cargo test -p rpc-extractor --features nats_integration_tests,node_integration_tests test_integration_rpc_getchaintxstats -- --nocapture`

### DIFF
--- a/extractors/rpc/tests/integration.rs
+++ b/extractors/rpc/tests/integration.rs
@@ -330,8 +330,23 @@ async fn test_integration_rpc_getchaintxstats() {
                             assert_eq!(stats.window_final_block_height, 201);
                             assert_eq!(stats.window_block_count, 200);
                             assert_eq!(stats.window_tx_count, Some(200));
-                            assert_eq!(stats.window_interval, Some(34));
-                            assert_eq!(stats.tx_rate, Some(5.882352941176471));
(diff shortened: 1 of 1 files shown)


0xB10C:
Thanks for the rebase! I think this looks good and makes sense to do.

Generally, I don't think we have to be to pedantic about exact numbers, but rather make sure they are in a sensible range and aren't flat out 0 at some point, because something broke.

PR #429, opened 2026-05-12
fix: archiver build.rs use GIT_HASH or fallback

0xB10C:
In the build.rs of the archiver, we use `git rev-parse HEAD` to get the current commit of the Git repository. However, in a Nix build using fetchFromGitHub, the package is fetched as a tarball without the .git information, not a repository. Having .git information in there is a common way of introducing non-determinism into the binary.

To fix this, we can now set GIT_HASH as env variable OR we fall back to "00000000".

### DIFF
--- a/tools/archiver/build.rs
+++ b/tools/archiver/build.rs
@@ -1,13 +1,25 @@
 use std::process::Command;
 
 fn main() {
-    let output = Command::new("git")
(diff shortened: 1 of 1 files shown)


0xB10C:
Tested this in a nix build on my infra.

PR #435, opened 2026-05-20
Update custom corepc branch to support bitcoind v31.0 download (and remove manual download from CI)

0xB10C:
This is a followup to #379, which added a workaround to manually fetch Bitcoin Core for the Ubuntu CI as it wasn't avaliable in corepc yet. I've pushed a commit (https://github.com/0xb10c/corepc/commit/3ceda10f34e945bbf8532f881d6dbc13d98a5d88) to the custom corepc branch we maintain anyway to be able to download it. Upgrading to this commit here allows us to drop the manual download. 

We can drop this the next time we update to a newer corepc release with this merged.


0xB10C:
@xyzconstant do you want to have a look at this?

xyzconstant (approved):
Seems good to me, I've tested exactly these changes before but using my `corepc` fork instead.

ACK cb117b875a4dd71fd757b3c0dbfc277f2669954d

Issue #436, opened 2026-05-20
ipc-extractor: tracking issue and followups


0xB10C:
#379 added an ipc-extractor. This issue is a place to discuss next steps and followups.

- [x] @ViniciusCestarii's comment on https://github.com/peer-observer/peer-observer/pull/379#pullrequestreview-4154925468: ipc_fetch_duration / ipc_fetch_errors on a local Prometheus endpoint is a good followup too. Done in https://github.com/peer-observer/peer-observer/pull/452
- [x] https://github.com/peer-observer/peer-observer/pull/379#discussion_r3265443316 cleanup when corepc supports v31.0: done in https://github.com/peer-observer/peer-observer/pull/435
- [x] https://github.com/peer-observer/peer-observer/pull/379#discussion_r3265604669 cleanup of true, true, true, true, ... in #438 
- [x] https://github.com/peer-observer/peer-observer/pull/379#discussion_r3266253521: move `tokio-util` to shared. It's already a dependency for p2p-extractor - done in https://github.com/peer-observer/peer-observer/pull/437
- [ ]  "is it possible to support multiple versions of Bitcoin Core's IPC interface"?
- [ ] A next step is to implement more IPC events in the extractor. #379 was only a PoC and added `getTip` - but we want to extend on that.

PR #437, opened 2026-05-20
clean: move tokio-util to shared and re-export

0xB10C:
Both the p2p-extractor and the ipc-extractor have a dependency on tokio-util. We can combine these in `shared` and re-export from there.

This is a followup to #379 and tracked in #436.


m4ycon:
ACK d7e19b4

It seems pretty straight foward, with github actions passing I don't think there's much to test.

PR #438, opened 2026-05-20
cleanup: In tests, pass structs for the checks we want to enable instead of bools

0xB10C:
While the bools where fine initially with very few args, it got unreadable and bug prone with more. Additionally, we need to touch unrelated tests when we add a new feature.

Pass structs instead. Similar to the RPC extractor tests in https://github.com/peer-observer/peer-observer/blob/dc0af9251d41b54facef1c9b1077b76ad0e40170/extractors/rpc/tests/common/mod.rs#L26-L60

This PR also drops our implementations of Args::new() and constructs an Args struct in the test. This has the benefit of being a lot clearer on which argument ends up at which position in the struct.

### DIFF
--- a/extractors/p2p/tests/integration.rs
+++ b/extractors/p2p/tests/integration.rs
@@ -55,22 +55,22 @@ struct EnabledChecksInTest {
 }
 
 fn make_test_args(nats_port: u16, p2p_address: String, checks: EnabledChecksInTest) -> Args {
-    Args::new(
-        NatsArgs {
+    Args {
+        nats: NatsArgs {
             address: format!("127.0.0.1:{}", nats_port),
             username: None,
             password: None,
             password_file: None,
         },
-        log::Level::Trace,
+        log_level: log::Level::Trace,
         p2p_address,
-        Network::Regtest,
-        PING_INTERVAL_SECONDS,
-        !checks.ping,
-        !checks.addrv2,
-        !checks.invs,
-        !checks.feefilter,
-    )
+        p2p_network: Network::Regtest,
+        ping_interval: PING_INTERVAL_SECONDS,
+        disable_ping: !checks.ping,
+        disable_addrv2: !checks.addrv2,
+        disable_invs: !checks.invs,
+        disable_feefilter: !checks.feefilter,
+    }
 }
 
 fn setup_node(conf: bitcoind::Conf) -> bitcoind::BitcoinD {
--- a/extractors/p2p/tests/integration.rs
+++ b/extractors/p2p/tests/integration.rs
@@ -46,14 +46,15 @@ fn setup() {
     });
 }
 
-fn make_test_args(
-    nats_port: u16,
-    p2p_address: String,
-    disable_ping: bool,
-    disable_addrv2: bool,
-    disable_invs: bool,
-    disable_feefilter: bool,
-) -> Args {
+#[derive(Default)]
+struct EnabledChecksInTest {
+    ping: bool,
+    addrv2: bool,
+    invs: bool,
+    feefilter: bool,
+}
+
+fn make_test_args(nats_port: u16, p2p_address: String, checks: EnabledChecksInTest) -> Args {
     Args::new(
         NatsArgs {
             address: format!("127.0.0.1:{}", nats_port),
@@ -65,10 +66,10 @@ fn make_test_args(
         p2p_address,
         Network::Regtest,
         PING_INTERVAL_SECONDS,
-        disable_ping,
-        disable_addrv2,
-        disable_invs,
-        disable_feefilter,
+        !checks.ping,
+        !checks.addrv2,
+        !checks.invs,
+        !checks.feefilter,
     )
 }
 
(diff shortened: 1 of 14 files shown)


0xB10C:
Draft as I want to look into the alerts tool's args too.

0xB10C:
This is ready for review now. I just did another pass on this and think it's fine. If there's no further review, I'll merge in a couple of days.

extractors/p2p/tests/integration.rs
```diff
@@ -46,14 +46,27 @@ fn setup() {
...
-) -> Args {
+#[derive(Default)]
+struct EnabledChecksInTest {
+    ping: bool,
+    addrv2: bool,
+    invs: bool,
+    feefilter: bool,
+}
+
+#[allow(dead_code)]
```
deadmanoz: I guess this is sensible to define now, even if currently unused!

deadmanoz (approved):
LGTM

RazorBest (commented):
Looks good. The only issue could come from copying mistakes. I did some manual fuzzing on the modified code, to see whether I can trigger test failures. However, these results only reflect how sensitive the tests are. 

## p2p

Negating the subfields of `checks` in `make_test_args` has an effect, with the exception of `!checks.addrv2`. 

## alerts

Changing in `AlertSettingsInTest::default()`, the values of `ping_window_secs`, `ping_window_secs`, `addr_threshold`, `addr_window_secs`, `peer_stale_secs` doesn't have an effect on the tests. Changing the default of `ping_threshold` to 0 fails one test.

In alerts integration tests, `make_test_args`, changing the value passed to `ping_window_secs` doesn't have an effect in the tests. Changing the other fields has an effect.

- `test_ping_spammer` reacts to the change of `ping_threshold` and `port`
- `test_addr_spammer` reacts to the change of `addr_threshold`
- `test_addrv2_spammer` reacts to the change of `addr_threshold`
- `test_ping_spammer_outbound_ignored` **doesn't react** to the change of `ping_threshold` (expected)
- `test_integration_alerts_fail_if_no_nats` **doesn't react** to the change of the port (expected)
- `test_closed_connection_cleans_state` reacts to the change of `port`
- `test_ping_spammer_fires_only_once` reacts to the change of `ping_threshold`
- `test_flagged_peer_disconnect_emits_disconnect_alert` reacts to the change of `ping_threshold`
- `test_stale_peer_cleanup_emits_disconnect_alert` fails when `peer_stale_secs` is 0, **but not** when it's increased by 1
- `test_undecodable_payload_does_not_kill_tool` **doesn't react** to `ping_threshold` being decreased by 1

## logger

Changing any one field `true -> false` in `EnabledLoggersInTest::all` causes at least one test to fail. Changing all the fields to false **doesn't cause any test to fail**.

Negating any of the `loggers` members in `make_test_args` causes at least one test to fail. Negating all of them **doesn't cause any test to fail**.

Replacing `EnabledLoggersInTest::all()` with `EnabledLoggersInTest::default()` **doesn't cause any test to fail**.

extractors/p2p/tests/integration.rs
0xB10C: actually, I dropped it!

0xB10C:
Thanks for testing the tests @RazorBest! I'm not sure if you're recommending a change or not.

tools/logger/tests/integration.rs
```diff
@@ -86,14 +102,14 @@ fn make_test_args(
...
-        log_extractor,
-        ipc_extractor,
+        loggers.messages,
+        loggers.connections,
+        loggers.mempool,
+        loggers.validation,
+        loggers.rpc,
+        loggers.ipc,
+        loggers.p2p_extractor,
+        loggers.log_extractor,
```
0xB10C: Hm, actually, we are only moving the problem here...

tools/alerts/tests/integration.rs
```diff
@@ -26,28 +26,41 @@ use shared::{
...
-        addr_threshold,
-        addr_window_secs,
+        ping_threshold: settings.ping_threshold,
+        ping_window_secs: settings.ping_window_secs,
+        addr_threshold: settings.addr_threshold,
+        addr_window_secs: settings.addr_window_secs,
         log_level: log::Level::Trace,
-        peer_stale_secs,
+        peer_stale_secs: settings.peer_stale_secs,
     }
```
0xB10C, line 64: It's fine here, as we are populating the Args fields for the alerts tool.

extractors/p2p/tests/integration.rs
```diff
@@ -65,10 +66,10 @@ fn make_test_args(
...
         PING_INTERVAL_SECONDS,
-        disable_ping,
-        disable_addrv2,
-        disable_invs,
-        disable_feefilter,
+        !checks.ping,
+        !checks.addrv2,
+        !checks.invs,
+        !checks.feefilter,
     )
```
0xB10C: But also problematic here.

RazorBest:
> I'm not sure if you're recommending a change or not.

No, not recommending anything. I checked again - the only sus thing is if you set all the arguments for the logger to false, the tests still pass. This comes from the fact that "no fields enabled" is interpreted as "all fields enabled" (check here: https://github.com/peer-observer/peer-observer/blob/6b68379c84298827063f00e982bd63e179925916/tools/logger/src/lib.rs#L73).

0xB10C:
I went ahead and removed all `Args::new()` functions in a separate commit and I don't think we should use them going forward.

RazorBest (approved):
I didn't know you could bypass Rust's lack of keyword arguments like this. Default also made things easier. Pretty neat.

LGTM

PR #442, opened 2026-05-21
fix: metric description for distinct source ASN

0xB10C:
This was likely just copy pasted from the line above and the description wasn't updated.


0xB10C:
Merging, I think this is trivial and does not require much review.

Issue #397, opened 2026-04-09
archiver-replayer tool: create sequence diagrams from recorded P2P messages


0xB10C:
With the archiver tool proposed in #373 by @octaviolucca, we record a bunch of events to disk for archival. The replayer tool can be used to replay an archive. 

I think it would be cool to have a way of creating a sequence diagram from such an archive. This allows us to get a good visual overview of which message were send and received from one/multiple/all peers during a certain timeframe. The issue #396 is similar in the way that the goal is to show a sequence diagram, but that's for real-time websocket data. Here, we want to get it from an archive.

Two approaches came to mind:
1. Mix replayer and python: add a feature (with a flag) to the replayer that extracts this data into a file that can be used by e.g. a python script. Possibly a CSV file or a JSON file. Then, have a python script that can produce the sequence diagram. I'm wondering if there are any good sequence diagram libraries for python
2. Implement everything in the replayer: Again, with a feature flag, the replayer can print a sequence diagram per peer. This might only be ascii though, or something that can be consumed by https://sequencediagram.org/ or https://mermaid.ai/open-source/syntax/sequenceDiagram.html or something else/better.

Both approches need a bit of exploration. I'm happy to provide a few sample archives.

octaviolucca:
The way I see it, approach 1 is a prerequisite for approach 2, once the replayer can export structured data (JSON/CSV), that same data can feed into different outputs: CLI text syntax for sequencediagram.org/mermaid, a Python script, or even a hosted page like the existing peers.html but for archived data.

For the CLI, I'm thinking of something like --list-peers to see what's in an archive, and --export-messages with filters for peer, timerange, and command types. From there we can build the diagram generation on top or use the data with Python scripts.

Let me know if you want to discuss any implementation details before I start.

0xB10C:
I ended up working on a web-based archive viewer https://peer-observer.github.io/archive-viewer/ which also includes a sequence diagram: 

<img width="775" height="706" alt="Image" src="https://github.com/user-attachments/assets/7d9d8358-79b8-4753-9091-54d6b2c77678" />

I think this resolves this issue for now.

Issue #439, opened 2026-05-20
archiver: tracking and followup issue


0xB10C:
#373 implemented a first experimental version of #121. Using this issue to track followups and potential next steps from this PR.

- [x] replayer needs features `strict = []` and enforcing of this https://github.com/peer-observer/peer-observer/pull/373#discussion_r3224846773
- [x] use marker for test readiness check instead of waiting for 1s https://github.com/peer-observer/peer-observer/pull/373#discussion_r3224580348 and https://github.com/peer-observer/peer-observer/pull/373#discussion_r3224574815
- [ ] document archiver compression memory usage https://github.com/peer-observer/peer-observer/pull/373#discussion_r3224427871
- [x] remove dev dependecies https://github.com/peer-observer/peer-observer/pull/373#discussion_r3224395171 (done in https://github.com/peer-observer/peer-observer/pull/444)
- [x] replayer: use `clap` and `log` https://github.com/peer-observer/peer-observer/pull/373#discussion_r3224360398
- [x] log errors in replayer instead of unwrap: https://github.com/peer-observer/peer-observer/pull/373#discussion_r3224326440
- [x] Remove the manifest: "I'm still favoring removing the mainfest writing, especially session based, and having a capinfos like tool to show and generate them." https://github.com/peer-observer/peer-observer/pull/373#pullrequestreview-4269844782
- [ ] ...

0xB10C:
It might make sense to move the replayer into the `tools/archiver` folder. If we add a `capinfos`-like archive infos binary, it could live there too. This keeps the tools folder a bit leaner and the archiver related tools live closer together. Then, we could also have a `common` or `shared` folder for the archiver related tools that share e.g. a small library to read from the archives.

See https://doc.rust-lang.org/cargo/reference/cargo-targets.html#binaries and https://stackoverflow.com/a/36604610 for an example

edit: doing this in https://github.com/peer-observer/peer-observer/pull/445

deadmanoz:
Agree that the achive and replayer should be nested under `tools/archiver` as is done in https://github.com/peer-observer/peer-observer/pull/445

0xB10C:
I've started looking into what an archiver-without-a-manifest would look like in https://github.com/peer-observer/peer-observer/tree/2026-05-remove-archiver-manifest

I also noticed that the current replayer seems to read all events into memory. This works fine in tests, but with real-world files this will become a problem:

[...]

PR #443, opened 2026-05-21
add(replayer): sequence diagrams, CSV export, and P2P message filtering

octaviolucca:
Implements both approaches from #397.

  - `--sequence-diagram`: mermaid sequence diagram from archived P2P messages
  - `--html <path>`: self-contained HTML with client-side rendering
  - `--csv <path>`: export P2P messages as CSV for external tools
  - `--peer`: filter to specific peer IDs
  - `--list-peers`: peers sorted by message count
  - `--msg-type`: filter by message type (e.g. `--msg-type ping,inv,tx`)
  - `--from`/`--to`: timestamp filtering, works with all output modes

  Mermaid for diagram rendering, its open-source, well maintained,
  renders client-side (https://github.com/mermaid-js/mermaid).

  HTML output example: 
cargo run -p replayer -- /tmp/archiver-output/archive.20260520-171752-085.bin.zst --sequence-diagram --peer 56 --peer 92 --html /tmp/diagram.html
image

  Questions:
  - What should diagrams show beyond message type and timestamp?
  - Should we cap the default number of peers shown? Currently shows all.
  - Worth exploring a web UI where you drop archive files and browse interactively?

### DIFF
--- a/tools/replayer/README.md
+++ b/tools/replayer/README.md
@@ -3,6 +3,15 @@
 Reads peer-observer archive files (`.bin` or `.bin.zst`) and prints decoded
 events to stdout.
 
+## Example output
+
+```text
+header: version=1 git=abcd1234
+[1] ts=1234567890 ebpf: ...
+[2] ts=1234567891 ebpf: ...
+total: 2 events
+```
+
 ## Usage
 
 ```bash
@@ -26,6 +35,9 @@ cargo run -p replayer -- archive.0.bin.zst --sequence-diagram --html diagram.htm
 
 # write P2P events as CSV
 cargo run -p replayer -- archive.0.bin.zst --csv events.csv
+
+# filter by message type
+cargo run -p replayer -- archive.0.bin.zst --msg-type ping,pong,inv
 
 # filter events by timestamp (milliseconds)
 cargo run -p replayer -- archive.0.bin.zst --from 1703001600000 --to 1703001700000
--- a/tools/replayer/README.md
+++ b/tools/replayer/README.md
@@ -1,32 +1,34 @@
 # `replayer`
 
-Reads peer-observer archive files and prints decoded events to stdout.
-
-Supports:
-- `.bin`
-- `.bin.zst`
+Reads peer-observer archive files (`.bin` or `.bin.zst`) and prints decoded
+events to stdout.
 
 ## Usage
 
 ```bash
-cargo run -p replayer -- archive/test.0.bin
-cargo run -p replayer -- archive/test.0.bin.zst
-cargo run -p replayer -- archive/test.0.bin archive/test.1.bin.zst
+# print decoded events
+cargo run -p replayer -- archive.0.bin.zst
+
+# multiple files
+cargo run -p replayer -- archive.0.bin.zst archive.1.bin.zst
+
+# list peers sorted by message count
+cargo run -p replayer -- archive.0.bin.zst --list-peers
+
+# generate a mermaid sequence diagram
+cargo run -p replayer -- archive.0.bin.zst --sequence-diagram
+
+# filter diagram to specific peers
+cargo run -p replayer -- archive.0.bin.zst --sequence-diagram --peer 1 --peer 3
+
+# write diagram as self-contained HTML
+cargo run -p replayer -- archive.0.bin.zst --sequence-diagram --html diagram.html
+
+# write P2P events as CSV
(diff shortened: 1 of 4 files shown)


0xB10C:
Thanks for taking a stab at this @octaviolucca - I do however think fixing up some of the other followups https://github.com/peer-observer/peer-observer/issues/439., especially the replayer ones, have a higher priority for now. Then, probably fixing up some of the archiver ones too. So this PR might be a cool demo for now, but I'll only really focus on it once we have a more stable replayer and are confident about our archiver format. 


I'm also not too sure about adding a full HTML renderer in here. This could just split out the mermaid and allow you to paste it on e.g. https://mermaid.live or a static GH pages site?

> Ran: cargo run -p replayer -- /tmp/archiver-output/archive.20260520-171752-085.bin.zst --sequence-diagram --peer 1 --peer 2 --peer 3 --html /tmp/diagram.html

this seems like it would output data for three peers, but I only see one peer in the graph?

> What should diagrams show beyond message type and timestamp?

Message type is a good start I think, especially with the counts for the INVs and addr. Maybe we can add a more information about the INV type (e.g. 40x WTx) and addr message network (4x IPv4, 2 IPv6). I don't see timestamps in the sequence diagram yet. Maybe just showing time deltas would be nice?

> Worth exploring a web UI where you drop archive files and browse interactively?

Not now, but maybe in the future. Not sure if that's something that can happen easily in the browser. Especially for large archives that need to be decopressed.

--- 

What does the CSV file format look like?

0xB10C:
As a more general note: you can open PRs against your own fork if you want so check if CI passes and iterate there before opening a PR here.

octaviolucca:
> Thanks for taking a stab at this @octaviolucca - I do however think fixing up some of the other followups #439., especially the replayer ones, have a higher priority for now. Then, probably fixing up some of the archiver ones too. So this PR might be a cool demo for now, but I'll only really focus on it once we have a more stable replayer and are confident about our archiver format.

Makes sense, I'll focus on the #439 follow-ups first. Keeping this as a draft for now.

> I'm also not too sure about adding a full HTML renderer in here. This could just split out the mermaid and allow you to paste it on e.g. https://mermaid.live or a static GH pages site?

The HTML is pretty minimal, just a `<pre class="mermaid">` tag with a CDN script. No dependencies, works offline. I could change it to mermaid-only output if you prefer.

You can have that too by simply not using the --html flag:
cargo run -p replayer -- /tmp/archiver-output/archive.20260520-171752-085.bin.zst --sequence-diagram --peer 1
```
sequenceDiagram
    participant Node as Our Node
    participant P1 as Peer 1 (127.0.0.1:28213)

    Node->>P1: inv (30 items)
    Node->>P1: inv (44 items)
    Node->>P1: inv (70 items)
    Node->>P1: inv (10 items)
    P1->>Node: ping (nonce: 0x18b155a157700282)
    Node->>P1: pong (nonce: 0x18b155a157700282)
    Node->>P1: inv (34 items)
    Node->>P1: inv (1 items)
    Node->>P1: inv (3 items)
    Node->>P1: inv (18 items)
    Node->>P1: inv (9 items)
    Node->>P1: ping (nonce: 0x28a89d269965da50)
    P1->>Node: pong (nonce: 0x28a89d269965da50)
    Node->>P1: addrv2 (1 addrs)
    Node->>P1: inv (40 items)
```

> this seems like it would output data for three peers, but I only see one peer in the graph?

Wrong screenshot. Peers 1, 2, 3 didn't exist in that archive file. Updated the PR description with the correct peers.

> Message type is a good start I think, especially with the counts for the INVs and addr. Maybe we can add a more information about the INV type (e.g. 40x WTx) and addr message network (4x IPv4, 2 IPv6). I don't see timestamps in the sequence diagram yet. Maybe just showing time deltas would be nice?

Not yet. I'll add time deltas between messages

> What does the CSV file format look like?

  ```csv
  timestamp,peer_id,addr,command,direction,size
  1779297472417,4555,159.26.113.30:17203,getdata,in,37
  1779297472417,4555,159.26.113.30:17203,tx,out,222
  1779297472420,13994,itric7qmzumx6t2wbayzingobx244nhkfuhsqfdcbe2atya5thmq.b32.i2p:0,inv,in,109
  1779297472423,92,194.99.20.29:8333,getdata,in,37
  1779297472423,92,194.99.20.29:8333,tx,out,205

octaviolucca:
Closing this for now since the archiver/replayer follow-ups in #439 should come first. This was useful as a PoC, will revisit it later.

PR #444, opened 2026-05-21
remove: drop duplicate dev dependencies from archiver

0xB10C:
This is a followup to #373 and resolves a part of #439.

In Rust, dev-dependencies inherit from normal dependencies. No need to have them here twice.

PR #445, opened 2026-05-21
move: archiver and replayer into tools/archive as sub-tools

0xB10C:
Addresses https://github.com/peer-observer/peer-observer/issues/439#issuecomment-4510875052 by mostly just moving code around.

> It might make sense to move the replayer into the `tools/archiver` folder. If we add a `capinfos`-like archive infos binary, it could live there too. This keeps the tools folder a bit leaner and the archiver related tools live closer together. Then, we could also have a `common` or `shared` folder for the archiver related tools that share e.g. a small library to read from the archives.

Since we expect the binaries working (writing and reading in some form) with the archives to share most of their code, we can move them closer together while still having seperate binaries. This allows us to create a poper libary for reading archives later, which we can then base tooling (capinfos like or sequence diagram creation) without needing a new tools/XYZ entry for each.


0xB10C:
This conflicts with https://github.com/peer-observer/peer-observer/pull/444 and needs a rebase after this.

0xB10C:
>  error: `cargo run` could not determine which binary to run. Use the `--bin` option to specify a binary, or the `default-run` manifest key.
available binaries: archiver, replayer

Hm I figured this might break. Will see if we can add a special handling for this.

deadmanoz:
> > error: `cargo run` could not determine which binary to run. Use the `--bin` option to specify a binary, or the `default-run` manifest key.
> > available binaries: archiver, replayer
> 
> Hm I figured this might break. Will see if we can add a special handling for this.

Can't we just use:

```
cargo run -p archive --bin replayer -- --help
cargo run -p archive --bin replayer -- --help
```

The README would need to be updated too

0xB10C:
> Can't we just use:
> cargo run -p archive --bin archiver -- --help
> cargo run -p archive --bin replayer -- --help

Yes, the problem was more dynamically figuring out which binaries exist and which README we should check against. I've addressed this in a recent push, so CI seems to be passing now and is checking it.

```
🔍 Checking tool: logger
🔍 Checking tool: websocket
🔍 Checking tool: metrics
🔍 Checking tool: connectivity-check
🔍 Checking tool: archive
🔍 Checking tool: alerts
✅ All help outputs are documented correctly.
```

> The README would need to be updated too.

I think the per-tool READMEs are fine. And I just noticed that it's not even documented in the root README.. Will add a commit. Thanks!

octaviolucca (approved):
README examples use cargo run -p archiver/replayer but the package is now archive with two binaries, so it needs --bin to select which one to run.

tools/archive/README.md
```diff
@@ -104,3 +108,37 @@ Options:
...
+Supports:
+- `.bin`
+- `.bin.zst`
+
+### Usage
+
+```bash
+cargo run -p replayer -- archive/test.0.bin
+cargo run -p replayer -- archive/test.0.bin.zst
+cargo run -p replayer -- archive/test.0.bin archive/test.1.bin.zst
```
octaviolucca:
```suggestion
  cargo run --bin replayer -- archive/test.0.bin
  cargo run --bin replayer -- archive/test.0.bin.zst
  cargo run --bin replayer -- archive/test.0.bin archive/test.1.bin.zst
  ```

tools/archive/README.md
octaviolucca:
```suggestion
$ cargo run --bin archiver -- \
 ```

0xB10C:
Took the suggestions from @octaviolucca. Thanks!

PR #449, opened 2026-05-26
test(shared): add post-IBD tip wait helper

deadmanoz:
## Summary

This came out of reviewing the compact block reconstruction integration tests in #420. Those tests need a deterministic way to know that a regtest node has connected a mined block and left IBD before the next step tries to trigger post-IBD behaviour.

This adds `shared::testing::node::mine_and_wait_for_tip`, which mines one block on the given node, polls until that block is the active tip, verifies `initial_block_download` is false, and then drains the validation interface queue.

The existing `p2p-extractor` `address-announcement` and `rpc-extractor` `getorphantxs` tests both utilise the helper instead of generating a block and immediately asserting on IBD state. That should make the tests less sensitive to slow hardware (e.g. the hardware I actually use for testing, a VM on a 10yr old server!!) where bitcoind has not yet advanced all relevant state by the next line of test code.

UPDATE: The helper currently takes a single client because both call sites here are single-node. When the cross-node case lands (PR #420's `test_integration_logextractor_saw_new_header` is the obvious next caller, where `node2` mines and `node1` observes), the signature can grow to `(miner, observer)` or split into a dedicated `mine_and_wait_for_tip_observed_by` helper without disturbing the current callers.

## Testing

My usual stress: 50 consecutive runs of the integration suite on older hardware, all green.

### DIFF
--- a/extractors/rpc/tests/integration.rs
+++ b/extractors/rpc/tests/integration.rs
@@ -24,7 +24,7 @@ use shared::{
             },
         },
     },
-    testing::{REGTEST_ADDRESS, nats_server::NatsServerForTesting},
+    testing::{REGTEST_ADDRESS, nats_server::NatsServerForTesting, node::mine_and_wait_for_tip},
     tokio::{
         self, select,
         sync::{oneshot, watch},
@@ -480,10 +480,7 @@ async fn test_integration_rpc_getorphantxs() {
                 .collect();
 
             // The receiving node needs to be out of IBD to start accepting transactions.
-            node1
-                .client
-                .generate_to_address(1, &REGTEST_ADDRESS)
-                .expect("failed to generate to address");
+            mine_and_wait_for_tip(&node1.client);
 
             // node1 is peer=0 of node2
             const PEER_ID: u64 = 0;
--- a/shared/src/testing/node.rs
+++ b/shared/src/testing/node.rs
@@ -1,0 +1,51 @@
+use crate::{bitcoin::BlockHash, bitcoind, testing::REGTEST_ADDRESS};
+use std::{
+    thread::sleep,
+    time::{Duration, Instant},
+};
+
+const POST_IBD_TIP_TIMEOUT: Duration = Duration::from_secs(120);
+
+fn wait_for_tip_post_ibd(node: &bitcoind::Client, expected_tip: BlockHash) {
+    let start = Instant::now();
+
+    loop {
+        let info = node
+            .get_blockchain_info()
+            .expect("failed to get blockchain info while waiting for post-IBD tip")
+            .into_model()
+            .expect("failed to parse blockchain info while waiting for post-IBD tip");
+
+        if info.best_block_hash == expected_tip && !info.initial_block_download {
+            node.sync_with_validation_interface_queue()
+                .expect("failed to sync validation interface queue after post-IBD tip");
+            return;
+        }
+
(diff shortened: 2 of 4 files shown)


extractors/rpc/tests/integration.rs
```diff
@@ -480,10 +480,12 @@ async fn test_integration_rpc_getorphantxs() {
...
             // The receiving node needs to be out of IBD to start accepting transactions.
-            node1
-                .client
-                .generate_to_address(1, &REGTEST_ADDRESS)
-                .expect("failed to generate to address");
+            mine_and_wait_for_tip(
+                &node1.client,
+                &node1.client,
+                &REGTEST_ADDRESS,
+                Duration::from_secs(120),
```
0xB10C: could make the timeout a constant in `shared/src/testing/node.rs`. Don't think we are going to change configuration on a per-test level.

shared/src/testing/node.rs
```diff
@@ -0,0 +1,58 @@
...
+        sleep(Duration::from_millis(50));
+    }
+}
+
+/// Mines one block on `miner` and waits until `observer` has that block as its
+/// active post-IBD tip.
+pub fn mine_and_wait_for_tip(
+    miner: &bitcoind::Client,
+    observer: &bitcoind::Client,
+    address: &Address,
```
0xB10C:
```suggestion
```

shared/src/testing/node.rs
```diff
@@ -0,0 +1,58 @@
...
+    timeout: Duration,
+) {
+    let generated = miner
+        .generate_to_address(1, address)
```
0xB10C:
```suggestion
        .generate_to_address(1, REGTEST_ADDRESS)
```

shared/src/testing/node.rs
```diff
@@ -0,0 +1,58 @@
...
+
+        sleep(Duration::from_millis(50));
+    }
+}
+
+/// Mines one block on `miner` and waits until `observer` has that block as its
+/// active post-IBD tip.
+pub fn mine_and_wait_for_tip(
+    miner: &bitcoind::Client,
+    observer: &bitcoind::Client,
```
0xB10C: All callers currently pass the same client as miner and observer. Can simplify it for now.
```suggestion
    miner: &bitcoind::Client,
```

deadmanoz: Absolutely. This change was originally motivated by seeing `test_integration_logextractor_saw_new_header` (https://github.com/peer-observer/peer-observer/pull/420/changes#diff-a50a41ef033e3a5a57ce05a9bce73f5551b216e71f7cf1315f15215963a235e3R700) in #420, where the two node variant would be useful, hence the two clients.

But happy to collapse to one for the current state of the codebase, and extend when #420 is in.

deadmanoz:
Addressed review feedback:
- Made the 120s timeout a `POST_IBD_TIP_TIMEOUT` constant, there's no longer a timeout argument
- Dropped the `address: &Address` parameter, the helper now uses `&REGTEST_ADDRESS` directly
- Collapsed the (miner, observer) signature to a single node parameter for now, can have a new 2-signature variant as is needed for #420 later

shared/src/testing/node.rs
0xB10C: I see! I wasn't aware of that.

0xB10C:
Looks good to me, thanks!

PR #450, opened 2026-05-26
archiver: remove manifest.toml writing

0xB10C:
This removes the manifest writing from the archiver for now as mentioned in https://github.com/peer-observer/peer-observer/pull/373#pullrequestreview-4269844782 and https://github.com/peer-observer/peer-observer/issues/439#issuecomment-4537190823. The idea is to add a binary that allows us to extract archive information, similar to a manifest, soon.

I've also went ahead and added some trace logging to the archiver which is now shown in failing tests. Additionally, I'm also making sure that the archiver finalizes/flushes the files to disk before shutting down, as not doing this broke some tests for me.

closes https://github.com/peer-observer/peer-observer/pull/431
closes https://github.com/peer-observer/peer-observer/pull/434

### DIFF
--- a/tools/archive/src/archiver/mod.rs
+++ b/tools/archive/src/archiver/mod.rs
@@ -234,6 +234,7 @@ impl ArchiveFile {
 
     fn write_event(&mut self, event: &Event) -> std::io::Result<()> {
         let buf = event.encode_length_delimited_to_vec();
+        log::trace!("writing {:?} ({} bytes)", event, buf.len());
         self.writer.write_all(&buf)?;
         self.hasher.update(&buf);
         self.bytes_written += buf.len() as u64;
--- a/tools/archive/src/archiver/mod.rs
+++ b/tools/archive/src/archiver/mod.rs
@@ -4,7 +4,6 @@ mod error;
 
 pub use error::RuntimeError;
 
-use std::collections::HashSet;
 use std::fs::{self, File};
 use std::io::{BufWriter, Write};
 use std::path::{Path, PathBuf};
@@ -12,9 +11,6 @@ use std::time::{SystemTime, UNIX_EPOCH};
 
 use time::macros::format_description;
 use time::OffsetDateTime;
-
-use sha2::{Digest, Sha256};
-use shared::serde::Serialize;
 
 use shared::clap;
 use shared::clap::Parser;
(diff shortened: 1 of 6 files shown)


0xB10C:
re-triggering GH CI, which was down

tools/archive/src/archiver/mod.rs
```diff
@@ -435,43 +358,19 @@ pub async fn run(args: Args, mut shutdown_rx: watch::Receiver<bool>) -> Result<(
         }
     }
 
-    total_events += close_file(current_file, &args)?;
     total_files += 1;
```
deadmanoz, line 363: here

tools/archive/src/archiver/mod.rs
```diff
@@ -379,8 +304,7 @@ pub async fn run(args: Args, mut shutdown_rx: watch::Receiver<bool>) -> Result<(
         ArchiveFile::new(&args.output_dir, &args.base_name, args.compression_level)?;
     log::info!("Created archive file: {}", current_file.path.display());
 
-    let mut total_events: u64 = 0;
-    let mut total_files: u64 = 0;
+    let mut total_files: u64 = 1;
```
deadmanoz: Nit: Ok this is being pre-incremented now, but then we are also post-incrementing line ~361 and incrementing in the loop as well?

deadmanoz: So we probably over-report by 1? 

It's harmless as it is only logged (but then again, a heap of this project is about scraping logs.. 😛)

tools/archive/src/archiver/mod.rs
```diff
@@ -401,8 +325,7 @@ pub async fn run(args: Args, mut shutdown_rx: watch::Receiver<bool>) -> Result<(
 
                         if current_file.needs_rotation(args.max_file_size) {
                             match rotate(current_file, &args) {
-                                Ok((file_events, new_file)) => {
-                                    total_events += file_events;
+                                Ok(new_file) => {
                                     total_files += 1;
```
deadmanoz, line 331: here

deadmanoz (commented):
Also the following can be removed from the `tools/archive/README.md` (can always do as a cleanup once the churn in this tool is complete).

```
### Manifest

Each archive file has a corresponding `<base-name>.<timestamp>.manifest.toml` with its metadata
(version, NATS address, event count, uncompressed size, SHA-256 checksum, event types,
first/last timestamps).
```

0xB10C:
Yeah, the total files is off-by-one. Will fix!

While testing with `cargo run --bin archiver -- --output-dir . --max-file-size 2`, I noticed that we also don't seem to log when rotating to a new file.

tools/archive/src/archiver/mod.rs
0xB10C: Thanks, yeah an off-by-one here would clearly have been a bug!

0xB10C:
Dropped the manifest section from the README, fixed the off-by-one error, and moved the "Created archive file" logging into the `ArchiveFile::new()` function after the flush, so we always log on rotation.

octaviolucca (approved):
Very nice! Looks all good to me

PR #451, opened 2026-05-26
archiver: rework ArchiveHeader and implement an iterator-based ArchiveReader

0xB10C:
This reworks the ArchiverHeader to be a protobuf struct, which inherits nice extensible properties when we want to add more data to the header in the future. 

Additionally, in `tools/archive/src/read.rs` implements a iterator-like reader for archives. Previously, archives were read into memory as a whole, and not as individual messages. This caused high memory usage on large archives.


tools/archive/src/archiver/mod.rs
```diff
@@ -435,43 +326,19 @@ pub async fn run(args: Args, mut shutdown_rx: watch::Receiver<bool>) -> Result<(
         }
     }
 
-    total_events += close_file(current_file, &args)?;
     total_files += 1;
```
octaviolucca, line 331:
```suggestion
```

 Final file is already counted. Either from init = 1 or last rotation

octaviolucca (changes requested):
Looks way simpler now, nice.
Tested locally and it all works well, just caught the file count being off by one and the README still mentions the old header format.

tools/archive/src/archiver/mod.rs
0xB10C: Fixed this in #450! Thanks!

0xB10C:
> README still mentions the old header format.

Updated the README for the new header format!

0xB10C:
With #450 merged, this is now ready for further review. I dropped the first three commits from 450.

tools/archive/README.md
```diff
@@ -12,19 +12,17 @@ multiple archiver instances to run simultaneously for different recording jobs.
...
 
-Events are stored as sequential length-delimited protobuf messages (using `encode_length_delimited`
-from `prost`), preceded by a 16-byte header:
+Events are stored as sequential length-delimited protobuf events from `protobuf/event.proto`,
+preceded by a protobuf `ArchiveHeader` from `protobuf/archive/header.proto`. The
+protobuf messages are encoded using `encode_length_delimited` from `prost`.
 
 ```
-[ 2B magic "PA" ][ 1B version ][ 4B git commit hash ][ 9B reserved ]
+[ varint length ][ protobuf ArhiveHeader bytes ]
```
octaviolucca:
```suggestion
[ varint length ][ protobuf ArchiveHeader bytes ]
```

octaviolucca (approved):
Much cleaner now with #450 split out. Header rework and iterator reader look good. 
LGTM

PR #452, opened 2026-05-26
Add metrics for IPC extractor

xyzconstant:
Partially addresses #436, based on @ViniciusCestarii 's suggestion in https://github.com/peer-observer/peer-observer/pull/379#pullrequestreview-4154925468.

This PR introduces Prometheus metrics for the newly added ipc-extractor, as well as its corresponding integration test. Mostly mirroring the rpc-extractor code, as this is the reference.

Exposing:
  - `ipcextractor_ipc_fetch_duration_seconds`
  - `ipcextractor_ipc_fetch_errors_total`
  - `nats_publish_errors_total`

Thanks @ViniciusCestarii for the suggestion!

### DIFF
--- a/extractors/ipc/tests/common/mod.rs
+++ b/extractors/ipc/tests/common/mod.rs
@@ -1,0 +1,84 @@
+use shared::{
+    bitcoind,
+    log::{self, info},
+    nats_util::NatsArgs,
+    simple_logger::SimpleLogger,
+};
+
+use std::sync::Once;
+
+use ipc_extractor::Args;
+
+static INIT: Once = Once::new();
+
+// 1 second query interval for fast tests
+pub const QUERY_INTERVAL_SECONDS: u64 = 1;
+
+pub fn setup() {
+    INIT.call_once(|| {
+        SimpleLogger::new()
+            .with_level(log::LevelFilter::Trace)
+            .init()
+            .unwrap();
+    });
+}
+
+pub fn make_test_args(nats_port: u16, ipc_socket_path: String) -> Args {
+    Args {
+        nats: NatsArgs {
+            address: format!("127.0.0.1:{}", nats_port),
+            username: None,
+            password: None,
+            password_file: None,
+        },
+        log_level: log::Level::Trace,
+        ipc_socket_path,
+        query_interval: QUERY_INTERVAL_SECONDS,
+        prometheus_address: "127.0.0.1:0".to_string(),
+    }
+}
+
+fn bitcoin_node_exe_path() -> String {
+    if let Ok(p) = std::env::var("BITCOIN_NODE_EXE") {
+        return p;
+    }
+    let bitcoind_path =
+        std::path::PathBuf::from(bitcoind::downloaded_exe_path().expect("downloaded_exe_path"));
+    bitcoind_path
+        .parent()
+        .and_then(|p| p.parent())
+        .expect("unexpected bitcoind path layout")
+        .join("libexec")
+        .join("bitcoin-node")
+        .to_string_lossy()
+        .into_owned()
+}
+
+fn setup_node(conf: bitcoind::Conf) -> bitcoind::BitcoinD {
+    let exe_path = bitcoin_node_exe_path();
+    info!("Using bitcoin-node at '{}'", exe_path);
+    bitcoind::BitcoinD::with_conf(exe_path, &conf).unwrap()
+}
+
+pub fn configure_node() -> bitcoind::BitcoinD {
+    let mut node_conf = bitcoind::Conf::default();
+    node_conf.args = vec!["-regtest", "-ipcbind=unix"];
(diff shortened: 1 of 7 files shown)


extractors/ipc/tests/metrics_integration.rs
```diff
@@ -0,0 +1,111 @@
...
+#[tokio::test]
+async fn test_integration_metrics_server_basic() {
+    setup();
+    let node = configure_node();
+    let nats_server = NatsServerForTesting::new(&[]).await;
+    let args = make_test_args(nats_server.port, ipc_socket_path(&node));
+
+    let local = LocalSet::new();
+    local
+        .run_until(async move {
```
m4ycon, line 31: Why those "local" code? I see in its docs that this will run in the same thread, is it required to IPC work? Sorry, I don't know too much about IPC yet, just a diff that I've seen between yours and rpc.

extractors/ipc/tests/common/mod.rs
m4ycon, line 1: > Mostly mirroring the rpc-extractor code, as this is the reference.

A very secondary comment for a future task / far from this PR scope. As you said, this mirrors what rpc-extractor does in its own context, that makes me think if it wouldn't be good to reuse code between those cases... Probably it'd be a very generic typed interface but it would decrease a lot of "duplicated" code, and perhaps, make someone else's job easier if they add some similar extractor.

m4ycon (commented):
utACK 5b392ee

I've compared this PR with rpc-extractor relatives, side by side, and it seems good to me.

extractors/ipc/tests/metrics_integration.rs
xyzconstant: This is required by the Cap N' Proto RPC layer. For reference, the `RpcSystem` implementation in Rust states the following:

>An `RpcSystem` is a non-`Send`able `Future` and needs to be driven by a task
executor. A common way accomplish that is to pass the `RpcSystem` to
`tokio::task::spawn_local()`.


And here's the error you get when you run it outside `LocalSet`:

```
2026-05-28T04:42:45.558Z DEBUG [shared::nats_util] Connecting to NATS-server at 127.0.0.1:4222 without authentification
2026-05-28T04:42:45.562Z INFO  [ipc_extractor] Connected to NATS server at 127.0.0.1:4222
2026-05-28T04:42:45.563Z INFO  [ipc_extractor] Connected to IPC socket at /Users/xxx/Library/Application Support/Bitcoin/signet/node.sock

thread 'main' (16689157) panicked at /Users/xxx/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tokio-1.51.0/src/task/local.rs:445:29:
`spawn_local` called from outside of a `task::LocalSet` or `runtime::LocalRuntime`
```

That said, be advised that I'm a Rust newbie, so I'm not 100% sure if this could be done differently. Any suggestions would be appreciated!

extractors/ipc/tests/common/mod.rs
xyzconstant: Yeah, I was thinking about this too, definitely would be helpful for future Prometheus integrations

xyzconstant:
Rebased and fixed conflicts with master

5b392ee5ff873cb2acf8309be78e0c64071db049 -> b55cfcb ([compare](https://github.com/peer-observer/peer-observer/compare/5b392ee5ff873cb2acf8309be78e0c64071db049..b55cfcbdfbb2ed8c9f9069f3708478c12e1d612f))

Thanks for the review @m4ycon

GuiSchet:
Tested this locally against v31 — all tests in the `ipc-extractor` crate pass cleanly (`integration.rs` + the 2 new `metrics_integration.rs` tests).

One small thing, non-blocking:

- Compared to `rpc-extractor`'s metrics tests, as a follow-up, it'd be nice to also test that a failed IPC call (e.g. node killed mid-run) increments `ipc_fetch_errors_total` — and a similar one for `nats_publish_errors_total` by tearing down NATS after startup.

Nice work on this! 👍

GuiSchet (approved):

extractors/ipc/tests/metrics_integration.rs
```diff
@@ -0,0 +1,111 @@
...
+                    .expect("ipc extractor failed");
+            });
+
+            let metrics_port = select! {
+                addr = addr_rx => addr.expect("ipc-extractor should send bound address").port(),
+                result = &mut handle => {
+                    result.unwrap();
+                    unreachable!("ipc-extractor task exited before sending bound address");
+                }
+            };
```
0xB10C, line 47: I was wondering if we need a timeout here. Can there be a case where we wait forever and the CI job won't finish?

0xB10C:
Code looks good to me. I haven't tested this, but plan to do so soon!

0xB10C:
Thanks @m4ycon @GuiSchet for testing! And thanks @xyzconstant for following up with this!

PR #441, opened 2026-05-20
Extend ipc-extractor to test bitcoin/bitcoin#29409

xyzconstant:
This is an extension of the `ipc-extractor` work to address #370.

As stated in https://github.com/peer-observer/peer-observer/issues/370#issuecomment-4500724967, this PR introduces the `ChainNotifications` interface to subscribe to the following chain events:

- `TransactionAddedToMempool`
- `TransactionRemovedFromMempool`
- `BlockConnected`
- `BlockDisconnected`
- `UpdatedBlockTip`
- `ChainStateFlushed`

`capnp` schemas are vendored from bitcoin/bitcoin#29409, so this `ipc-extractor` version works with a compiled binary built from that PR. They'll need to be re-synced if the upstream schemas change.

Note: the polling scaffolding is in place ready to be extended. The tip is now being fetched reactively, triggered by `UpdatedBlockTip` events.

PR #453, opened 2026-05-27
Improve disconnection handling in ipc-extractor

xyzconstant:
This PR cherry-picks 2 commits from #441.

Instead of detecting IPC connection loss via polling, this could be improved by watching the RPC task and using capnp's Disconnector object for clean shutdown.

### DIFF
--- a/extractors/ipc/src/ipc.rs
+++ b/extractors/ipc/src/ipc.rs
@@ -12,7 +12,7 @@ use init_capnp::init::Client as InitClient;
 use mining_capnp::mining::Client as MiningClient;
 use proxy_capnp::thread::Client as ThreadClient;
 
-use capnp_rpc::{RpcSystem, rpc_twoparty_capnp, twoparty};
+use capnp_rpc::{Disconnector, RpcSystem, rpc_twoparty_capnp, twoparty};
 use shared::{
     futures::AsyncReadExt,
     protobuf::ipc_extractor::BlockTip,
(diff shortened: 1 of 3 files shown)


0xB10C:
This looks like a worthwhile, small change. I'm not familiar with capnp's Disconnector, but from the sound of it, it makes sense to use here.

I assume a manual test for this would be to stop bitcoind and see how fast the ipc-extractor stops afterwards.

Do we have an integration test for bitcoind going offline (e.g. `bitcoin-cli stop`) and the ipc-extractor shutting down afterwards? If not, it might make sense to add one here as the first commit.

0xB10C:
> I assume a manual test for this would be to stop bitcoind and see how fast the ipc-extractor stops afterwards.

Tested this manually! It now stops the ipc-extractor immediately, while on `master` it only checks it during the next RPC fetch.

0xB10C:
> Do we have an integration test for bitcoind going offline (e.g. bitcoin-cli stop) and the ipc-extractor shutting down afterwards? If not, it might make sense to add one here as the first commit.

As part of review, I wrote the integration tests and was so free to push it.

This looks good to me, but I'll wait with merging until you had another look.

xyzconstant:
re: https://github.com/peer-observer/peer-observer/pull/453#issuecomment-4552171334

>I assume a manual test for this would be to stop bitcoind and see how fast the ipc-extractor stops afterwards

Yes, that's right! Forgot to mention that

re: https://github.com/peer-observer/peer-observer/pull/453#issuecomment-4552375127

The code looks good to me. Thanks for adding it!!!

0xB10C:
Thank you!

PR #455, opened 2026-06-02
refactor(replayer): use clap and log, return errors instead of unwrap

octaviolucca:
Follow-up from the https://github.com/peer-observer/peer-observer/pull/373 review comments.
- Use clap for replayer arguments and add --log-level
- Log decoded replay output at info level
- Log read/display errors instead of unwrapping
- Treat failures before a valid archive header as errors and exit 1
- Treat `UnexpectedEof` while reading events as a truncated archive error and exit 1
- Update the archive README with the generated replayer help

### DIFF
--- a/tools/archive/src/bin/replayer.rs
+++ b/tools/archive/src/bin/replayer.rs
@@ -1,50 +1,116 @@
 use archive::read::ArchiveReader;
-use shared::protobuf::event::event::PeerObserverEvent;
-use std::env;
-use std::path::Path;
+use shared::clap;
+use shared::clap::Parser;
+use shared::protobuf::event::{event::PeerObserverEvent, Event};
+use shared::{log, simple_logger};
+use std::io::ErrorKind;
+use std::path::PathBuf;
+use std::process::ExitCode;
 
-fn main() {
-    let files: Vec<String> = env::args().skip(1).collect();
-    if files.is_empty() || files.iter().any(|f| f == "--help" || f == "-h") {
-        println!("Read and display peer-observer archive files");
-        println!();
-        println!("Usage: replayer <file.bin[.zst]> [file2.bin[.zst] ...]");
-        return;
+#[derive(Parser, Debug)]
+#[command(version, about = "Read and display peer-observer archive files")]
+struct Args {
+    /// Archive files to read.
+    #[arg(value_name = "FILE", required = true)]
+    files: Vec<PathBuf>,
+
+    /// The log level the tool should run on.
+    #[arg(short, long, default_value_t = log::Level::Info)]
+    log_level: log::Level,
+}
+
+fn main() -> ExitCode {
+    let args = Args::parse();
+
+    if let Err(e) = simple_logger::init_with_level(args.log_level) {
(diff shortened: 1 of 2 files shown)


0xB10C:
Thanks for picking this up!

> Keep UnexpectedEof while reading events as a warning so partially replayable archives still exit 0

Any specific reason for this? I think it might be good to fail if we get an UnexpectedEof? Otherwise, we are might be missing bugs or cases where archiver writes broken archives?

tools/archive/src/bin/replayer.rs
```diff
@@ -1,50 +1,115 @@
...
+            let ebpf_event = e.ebpf_event.as_ref().ok_or("missing ebpf event")?;
+            println!("[{n}] ts={ts} ebpf: {ebpf_event}");
+        }
+        Some(PeerObserverEvent::RpcExtractor(r)) => {
+            let rpc_event = r.rpc_event.as_ref().ok_or("missing rpc event")?;
+            println!("[{n}] ts={ts} rpc: {rpc_event}");
+        }
+        Some(PeerObserverEvent::P2pExtractor(p)) => {
+            let p2p_event = p.p2p_event.as_ref().ok_or("missing p2p event")?;
+            println!("[{n}] ts={ts} p2p: {p2p_event}");
```
0xB10C: We probably want to make these `println!()` into `log::info!()`

0xB10C:
I was thinking about splitting the replayer up into our usual `lib.rs` and `main.rs` way to be able to integration test it, but that doesn't need to happen in this PR.

octaviolucca:
> > Keep UnexpectedEof while reading events as a warning so partially replayable archives still exit 0
> 
> Any specific reason for this? I think it might be good to fail if we get an UnexpectedEof? Otherwise, we are might be missing bugs or cases where archiver writes broken archives?

I was thinking of this as a partially recoverable archive: if the header and previous events decode successfully, the replayer can still show useful data even if the last event was interrupted.

But I agree that an `UnexpectedEof` still means the archive is truncated, and treating it as success could hide archiver bugs. This now logs an error and exits 1 for `UnexpectedEof` while reading events. I also updated the PR description.

tools/archive/src/bin/replayer.rs
octaviolucca: Done. Moved the event output to log::info!, and the header/total/separator lines too for consistency.

octaviolucca:
> I was thinking about splitting the replayer up into our usual `lib.rs` and `main.rs` way to be able to integration test it, but that doesn't need to happen in this PR.

Makes sense. I'll do it as a follow-up

0xB10C:
I was thinking about how this relates to #418, but it seems like we don't do much error handing here. 

I think once #418 is merged, we can do more anyhow error handling in the same PR that splits up into lib.rs and main.rs.

deadmanoz (approved):
LGTM

0xB10C:
This looks good to me.

While testing it, I found a bug in the archive reading code for larger archives. Fixing this in a separate PR.

Thanks @octaviolucca!

PR #463, opened 2026-06-26
fix: correctly read protobuf message varints in archives

0xB10C:
Due to using `fill_buf()`, it could have happened that our read buffer did only contain a part of the varint that marks the size of the next protobuf message. We'd fail reading the varint with:

failed to decode Protobuf message: invalid varint

To avoid this, stop decoding the varint out of the buffered reader's current contents (which may hold only part of it) and instead read it byte-by-byte until the continuation bit clears. We read byte-by-byte because prost only decodes a varint from data already in memory, and we don't know up front how many bytes it'll be.

### DIFF
--- a/tools/archive/src/read.rs
+++ b/tools/archive/src/read.rs
@@ -5,8 +5,7 @@ use shared::protobuf::event::Event;
 use shared::zstd;
 use std::ffi::OsStr;
 use std::fs::File;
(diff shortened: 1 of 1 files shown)


octaviolucca:
LGTM. 
The split-varint test covers the issue, and archive tests pass locally.

PR #464, opened 2026-06-29
refactor(archive): split replayer into library module and use anyhow

octaviolucca:
Follow-up from #455.

This splits the archive replayer logic out of `tools/archive/src/bin/replayer.rs` and into `archive::replayer`. This makes the replay behavior testable without spawning the binary.

Changes:
- move replayer args and replay logic into `tools/archive/src/replayer.rs`
- expose the new module from `tools/archive/src/lib.rs`
- add focused tests for valid archives, missing headers, truncated events, malformed events, display-only errors, and run-level error reporting
- update the new replayer code to use `anyhow` with archive/event context, following #418

The replayer still preserves the existing run-level behavior of processing all requested files and returning a failing exit status if any replay fails.

### DIFF
--- a/tools/archive/src/bin/replayer.rs
+++ b/tools/archive/src/bin/replayer.rs
@@ -1,18 +1,16 @@
 use archive::replayer::{self, Args};
+use shared::anyhow::{Context, Result};
 use shared::clap::Parser;
 use shared::simple_logger;
-use std::process::ExitCode;
 
-fn main() -> ExitCode {
+fn main() -> Result<()> {
     let args = Args::parse();
 
-    if let Err(e) = simple_logger::init_with_level(args.log_level) {
-        eprintln!("replayer tool error: {}", e);
-    }
+    simple_logger::init_with_level(args.log_level).context("could not initialize logger")?;
 
     if replayer::run(&args) {
-        ExitCode::FAILURE
-    } else {
-        ExitCode::SUCCESS
+        std::process::exit(1);
     }
+
+    Ok(())
 }
--- a/tools/archive/src/bin/replayer.rs
+++ b/tools/archive/src/bin/replayer.rs
@@ -1,23 +1,7 @@
-use archive::read::ArchiveReader;
-use shared::clap;
+use archive::replayer::{self, Args};
 use shared::clap::Parser;
-use shared::protobuf::event::{event::PeerObserverEvent, Event};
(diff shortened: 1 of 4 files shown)


0xB10C:
fyi: just merged #418.

octaviolucca:
> fyi: just merged #418.

thanks for letting me know, will rebase on current master and keep this PR focused on the replayer split/testing unless you prefer the anyhow conversion here too.  What do you think?

tools/archive/src/replayer.rs
```diff
@@ -0,0 +1,160 @@
...
+    let mut events_count = 0;
+    for (idx, event) in archive.enumerate() {
+        let n = idx + 1;
+        let event = match event {
+            Ok(event) => event,
+            Err(e) if e.kind() == ErrorKind::UnexpectedEof => {
+                return Err(ReplayFileError::UnexpectedEof {
+                    event_number: n,
+                    source: e,
+                });
```
0xB10C: Yes, I think all new code should use `anyhow`. We don't want to do another `anyhow` refactor in the future.

Here you'd can probably have a anyhow Context with the event number e.g.: `error at event {idx}`

0xB10C: Same for the error below and above

octaviolucca: updated the replayer to use `anyhow` and added archive/event context for replay errors

tools/archive/src/bin/replayer.rs
```diff
@@ -1,115 +1,15 @@
...
-            log::info!("[{n}] ts={ts} log: {log_event}");
-        }
-        Some(PeerObserverEvent::IpcExtractor(i)) => {
-            let ipc_event = i.ipc_event.as_ref().ok_or("missing ipc event")?;
-            log::info!("[{n}] ts={ts} ipc: {ipc_event}");
-        }
-        None => log::info!("[{n}] ts={ts} <unknown>"),
+    if replayer::run(&args) {
+        std::process::exit(1);
     }
```
0xB10C, line 14: I think this is OK, but we could have a cleaner `ctrl+c` handling here similar to the other tools. Not having it isn't problematic as this will just stop printing - there's nothing to cleanly shutdown or stop.

0xB10C:
Looks good! Thanks @octaviolucca!

Issue #421, opened 2026-04-30
docs: Dashboard's description


m4ycon:
With all that many dashboards in [metrics/dashboards](https://github.com/peer-observer/peer-observer/tree/master/tools/metrics/dashboards) folder, I miss some explanation about each one, even a brief one, like done in [[addr] addr rate-limited peers](https://demo.peer.observer/monitoring/d/9ba67dc0-17e5-4333-9b18-10258f9cf7ba/addr-addr-rate-limited-peers?orgId=1&from=now-7d&to=now&timezone=utc).

Few questions that I think should be answered at EVERY dashboard:
- What does it show? -- Some of them are pretty straight foward, like [[fees] Fee Rate Estimates](https://demo.peer.observer/monitoring/d/bcxw6cf/fees-fee-rate-estimates-with-labels?orgId=1&from=2026-04-03T17:46:46.527Z&to=2026-04-23T16:42:03.381Z&timezone=browser&var-datasource=P1809F7CD0C75ACF3&refresh=1m), so maybe not every dashboard should answer that (?).
- What should I worry about it? -- If a value is below or above x? Or is it just to see if anything goes far away from average? I think most of them are the second option, right?
- Any relevant reference? -- Issues, PRs... Like done in [[addr] addr rate-limited peers](https://demo.peer.observer/monitoring/d/9ba67dc0-17e5-4333-9b18-10258f9cf7ba/addr-addr-rate-limited-peers?orgId=1&from=now-7d&to=now&timezone=utc).

And of course, we don't wanna see this explanation (that may be big) everytime we open that dashboard. If grafana has something to "toggle" or a "hover" to see this explanation, it would be better.

0xB10C:
I think you can add a description to the dashboards. I'd need to check if this is show somewhere where we can hover over or with an ℹ️ symbol or similar. 


Generally agree. Most of them are probably only useful to me because I had a specific thing in mind when creating. For some of them there's probably non good explanation as we are looking for unknown unknowns. But this can be documented too.

A good first step would be to evaluate a few methods for showing Infos and then deciding on one. Then we can go through all of them and documented them in the same way.

Sounds like a plan?

m4ycon:
I did some quick tests on a dash and found this:

1. I took a look at the dashboard "description" but didn't find where it shows (if it does);
2. Panel's description create an "i" tooltip, I did a test with a big text and worked well, I guess; It also supports markdown;

[...]

PR #465, opened 2026-06-30
Add dashboards for compact block reconstruction metrics

m4ycon:
This PR is a follow up from #420, it introduces dashboards that shows metrics from it. This was separated from the previous PR to reduce review workload, many comments were made regarding those dashboards on it.

## What this PR does?

- Refactor "Compact block reconstruction without transaction requests" dashboard that actually is an approximation to be an "exact" one ([ref on demo](https://demo.peer.observer/monitoring/d/adtsyzmbdt0cgc/99bfac5?orgId=1&from=now-90d&to=now&timezone=browser));
- Add new dashboards about compact blocks:
  - "Compact block reconstruction in high-bandwidth per day";
  - "Requested transactions stats";
  - "Compact block reconstruction time".

## Dashboards

All screenshots are from my vps running this branch on mainnet.

### Compact block reconstruction without transaction requests per day

This dashboard already existed in the project but it was an approximation using some assumptions from p2p messages, now by watching logs we can be sure of how much of these reconstructions are actually requiring transactions. The yellow line is the "old" approximation, the green one is based on logs -- in the final version, I hide the approximation line. There's quite some difference 👀.

image

### Compact block reconstruction in high-bandwidth per day

Also a simple one, see heuristic in [How do we know if the reconstruction was via low or high-bandwidth connection?](https://github.com/peer-observer/peer-observer/pull/420#how-do-we-know-if-the-reconstruction-was-via-low-or-high-bandwidth-connection).

image

### Requested transactions stats

Some statistics about requested transaction that were not in mempool. Each representing the sum in a "per time" range. Size and number of requested transactions, and also number of transactions in extra pool.

Captura de tela_30-6-2026_145956_5 189 165 114

### Compact block reconstruction time

This dash shows the average time that a block took to make the reconstruction round-trip, it doesn't filter by bandwidth nor with/without transactions.

image

### DIFF
--- a/tools/metrics/dashboards/playlist/compactblock-reconstruction-high-bandwidth.json
+++ b/tools/metrics/dashboards/playlist/compactblock-reconstruction-high-bandwidth.json
@@ -71,12 +71,12 @@
             "mode": "absolute",
             "steps": [
               {
-                "color": "transparent",
+                "color": "green",
                 "value": 0
               },
               {
-                "color": "transparent",
-                "value": 0
+                "color": "red",
+                "value": 80
               }
             ]
           },
@@ -90,7 +90,7 @@
         "x": 0,
         "y": 0
       },
-      "id": 2,
+      "id": 1,
       "options": {
         "legend": {
           "calcs": [],
@@ -100,7 +100,7 @@
         },
         "tooltip": {
           "hideZeros": false,
-          "mode": "multi",
+          "mode": "single",
           "sort": "none"
         }
       },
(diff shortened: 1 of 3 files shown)


m4ycon:
Additionally, I'll probably create another follow up detailing "Compact block reconstruction time" dashboard more, e.g. separated panels for low/high bandwidth. Or maybe I do it here, idk... For now, that's it.

0xB10C:
Do you want to incorporate the link-descriptions as discussed in https://github.com/peer-observer/peer-observer/issues/421 here? Happy to review it, if you end up writing something.

(I guess you have short descriptions in https://github.com/peer-observer/peer-observer/pull/465#issue-4779314570 already, that you can reuse)

m4ycon:
Updated all dashboards with a "link" description, search for `"links"` on changes to check it out. I think that all of them are "unknown unknowns", so in terms of anomalies I didn't write anything. Neither what should be a strange thing to see in them, they all fall in the case of "far from average" would be strange, should we explicitly tell it?

Also, those descriptions which are not real links, it's better to leave with a `#` to avoid user clicking on it and leaving the page.

m4ycon:
Small adjustment to how tooltip shows values, set "multi" and "descending" to all of them.

image

tools/metrics/dashboards/playlist/compactblock-reconstruction-without-tx-req.json
```diff
@@ -120,13 +122,28 @@
...
           "useBackend": false
+        },
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "P1809F7CD0C75ACF3"
+          },
+          "editorMode": "code",
+          "expr": "1 - sum(increase(peerobserver_p2p_message_count{message=\"getblocktxn\", direction=\"outbound\"}[1d])) by (host) / sum(increase(peerobserver_p2p_message_count{message=\"cmpctblock\", direction=\"inbound\"}[1d])) by (host)",
+          "hide": true,
```
0xB10C, line 133: hm this adds a hidden query that we might not need. It's useful in testing, but does not need to be in the merged dashboards.

0xB10C: Image

0xB10C:
For "Compact block reconstruction without transaction requests per day":

PR image

Approximation image

I'm still not sure why the approximation and the one added in this PR are so different. I haven't done a pass over my logs in a while. Maybe that would give us more clarity.

m4ycon:
Rebased with main and removed useless query ([comment](https://github.com/peer-observer/peer-observer/pull/465#discussion_r4050379226)). btw I don't know what I've done that this branch didn't have the description of each dashboard with the link strategy, neither the tooltip sorting, but now it's right.

About comparing approximation and the new version from this PR, taking a look at the expression `cmpctblock` could be sent by many peers to us, but we'd only make one `getblocktxn` per block (if needed). And as it does a `1 - (getblocktxn/cmpctblock)`, if we are overcounting `cmpctblock`, we'll have a higher result. Does it makes sense? Maybe to verify that, we could take a look at one specific block messages and see if this happens, check one that needed to make a request and one that didn't.

PR #467, opened 2026-07-01
fix: CI workflow permissions can be reduced

0xB10C:
This was noted in GH code scanning: https://github.com/peer-observer/peer-observer/security/code-scanning/12


0xB10C:
fewer permission shouldn't hurt, so merging.

PR #479, opened 2026-08-07
test(archive): replace fixed sleeps with NATS synchronization

octaviolucca:
Follow-up for #439.

This replaces the fixed sleeps in the archive integration tests with explicit NATS synchronization

The archiver now sends an optional readiness signal after the NATS subscription is registered and the initial archive file is ready. The tests also use a NATS round-trip probe before shutdown, so they no longer rely on arbitrary delays to wait for published events to be processed.

This also drains the NATS subscription on shutdown, allowing in-flight messages to be yielded before the archive file is finalized

Tested on my fork CI:
https://github.com/octaviolucca/peer-observer/actions/runs/31206841457

### DIFF
--- a/tools/archive/src/archiver/mod.rs
+++ b/tools/archive/src/archiver/mod.rs
@@ -3,7 +3,7 @@
 use std::fs::{self, File};
 use std::io::{BufWriter, Write};
 use std::path::{Path, PathBuf};
-use std::time::{SystemTime, UNIX_EPOCH};
+use std::time::{Duration, SystemTime, UNIX_EPOCH};
 
 use time::macros::format_description;
 use time::OffsetDateTime;
@@ -20,6 +20,7 @@ use shared::protobuf::ebpf_extractor::ebpf;
 use shared::protobuf::event::event::PeerObserverEvent;
 use shared::protobuf::event::Event;
 use shared::tokio::sync::watch;
+use shared::tokio::time::Instant;
 use shared::zstd;
 
 struct TrackingWriter {
@@ -245,6 +246,11 @@ impl Args {
     }
 }
 
+/// How long to wait for the NATS subscription to drain before shutting down
+/// anyway. Draining never completes while the NATS server is unreachable, as
+/// the client reconnects instead of processing the unsubscribe.
+pub const DRAIN_TIMEOUT: Duration = Duration::from_secs(10);
+
 pub async fn run(args: Args, mut shutdown_rx: watch::Receiver<bool>) -> Result<()> {
     if args.archive_all() {
         log::info!("archiving all events: {}", args.archive_all());
@@ -279,6 +285,9 @@ pub async fn run(args: Args, mut shutdown_rx: watch::Receiver<bool>) -> Result<(
             .context("creating the initial archive file")?;
 
     let mut total_files: u64 = 0;
+    let mut draining = false;
+    // Only read once draining is set below.
+    let mut drain_deadline = Instant::now();
 
     loop {
         shared::tokio::select! {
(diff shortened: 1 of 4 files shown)


b10claude:
> From 0xb10c's LLM (Claude Code), not 0xb10c directly.

The `drain()` change can hang shutdown when NATS is unreachable. `Subscriber::drain()` only enqueues a command, and the connection task sits in `handle_disconnect().await` while reconnecting (unlimited by default, since `prepare_connection` uses a bare `ConnectOptions::new()`). The subscription is then never removed, `sub.next()` never yields `None`, and `if !draining` has disabled the other exit. The process needs a SIGKILL (tokio swallows the second Ctrl+C) and `finalize()` never runs — at the default `--compression-level 22` that drops a good chunk of the buffered tail. Before this PR the same case exited cleanly.

Bounding the drain keeps the benefit:

```rust
const DRAIN_TIMEOUT: Duration = Duration::from_secs(10);
let mut drain_deadline = Instant::now(); // only read once draining is set

// after sub.drain():
drain_deadline = Instant::now() + DRAIN_TIMEOUT;

// extra select! arm:
_ = shared::tokio::time::sleep_until(drain_deadline), if draining => {
    log::warn!("draining the NATS subscription timed out. Shutting down anyway.");
    break;
}
```

The deadline is absolute, so `select!` re-creating the future each iteration doesn't extend it.

octaviolucca:
Nice, thank you for spoting that @b10claude, I reproduced it and fixed with a bounded drain (DRAIN_TIMEOUT 10s, absolute deadline). On timeout it warns and breaks, so finalize() still runs

Also added test_shutdown_with_unreachable_nats as a regression test: hangs without the fix, shuts down in ~10s with it

tools/archive/src/archiver/mod.rs
```diff
@@ -307,20 +344,33 @@ pub async fn run(args: Args, mut shutdown_rx: watch::Receiver<bool>) -> Result<(
...
-                        break;
+                        log::warn!("The shutdown notification sender was dropped. Draining the subscription.");
+                        draining = true;
                     }
                 }
+                if draining {
+                    // Unsubscribes, but keeps the subscription open until all
+                    // in-flight messages have been yielded. Afterwards,
+                    // sub.next() returns None and the loop above breaks.
+                    sub.drain().await.context("draining the NATS subscription")?;
```
deadmanoz: The `?` here returns from `run()` without hitting the `finalize()` after the loop. As the zstd frame finalisation is only written in `finalize()` (there's no `Drop` on `ArchiveFile` to compensate), an error here leaves the last archive file unreadable. The likelihood is indeed low, `drain()` only fails when the client's connection task is gone, but this is exactly the "NATS is gone for good" case where we'd want the partial archive to be closed cleanly. So maybe use "the safe way" (log and `break` so `finalize()` still runs), rather than `?`:


```rust
if let Err(e) = sub.drain().await {
    log::warn!("failed to drain the NATS subscription: {}. Shutting down.", e);
    break;
}
```

tools/archive/tests/integration.rs
```diff
@@ -514,3 +540,40 @@ async fn test_no_compression() {
...
+    let mut archiver_handle = tokio::spawn(async move {
+        let args = make_test_args(nats_port, &dir);
+        run(args, shutdown_rx, Some(ready_tx)).await.unwrap();
+    });
+
+    wait_for_archiver_ready(ready_rx, &mut archiver_handle).await;
+
+    // Kill the NATS server while the archiver is connected to it.
+    drop(nats_server);
+    sleep(Duration::from_millis(200)).await;
```
deadmanoz, line 573: So this test exists to prove the archiver still shuts down when the NATS server is unreachable. In that state `drain()` can never finish: the archiver connects with the async-nats default of unlimited reconnects, so after losing the server the client's connection task retries the connection forever and never gets back to completing the drain, meaning `sub.next()` never returns `None`. The only way out is the `DRAIN_TIMEOUT` arm breaking the loop, so the test kills the NATS server, sends the shutdown, and expects the archiver to exit via that timeout.

The intent of this sleep is to give the archiver client time to notice the server is gone before the test shuts the archiver down. Two lines below, the test calls `shutdown_tx.send(true)`; `run()` receives that on its watch channel and reacts by calling `sub.drain()`. If the client hasn't noticed the dead server by then, the drain completes without it, and the test finishes in milliseconds through the normal break. It still passes, but it exercised the ordinary shutdown path rather than the `DRAIN_TIMEOUT` path it was written to cover.

Right now the only sign the intended path ran is that the test takes ~10s. A one-line `assert!(start.elapsed() >= DRAIN_TIMEOUT)` after the timeout(...) join at the end of the test could pin the test to that path and explain to future readers why this test is slow. FWIW, I couldn't trigger the fast (< 10s) pass in practice, but maybe it is worth being explicit?

deadmanoz (commented):
A few minor issues to consider @octaviolucca

tools/archive/tests/integration.rs
octaviolucca: thank you for the review, updated both:

- `drain()` errors now log a warning and break, so the archive file still goes through `finalize()`
- the unreachable NATS test now asserts it actually exits through the drain timeout path

deadmanoz:
Yeah nice one @octaviolucca, the pattern you've adopted matches what has been used elsewhere (tools/metrics, tools/websocket, extractors/p2p, extractors/ipc, extractors/rpc), so LGTM.

One thing I realised in reviewing this is that on the forced-shutdown path we aren't actually testing that the buffers are flushed and the zstd frame is properly finalised. The ordinary tests confirm that implicitly, since they read the archive back and count events, but `test_shutdown_with_unreachable_nats` only asserts that shutdown completes and we don't actually check the file contents.

I don't think this holds up this PR, but it could be worth a follow-up PR assessing archive validity under error conditions more generally? Today the only error-condition tests are on the read side: the replayer suite feeds hand-built broken files, all uncompressed, to the reader. Nothing yet checks what the archiver writes when something goes wrong. Some candidates:

- Extend the unreachable-NATS test to publish + sync() + read the archive back after the join
- Confirm a compressed archive survives a forced shutdown (currently zstd files are only ever read back after clean shutdowns)
- Verify undecodable events are skipped without corrupting the archive

On the read side: check the reader's behaviour on a truncated zstd file (the artefact a missed `finalize()` would leave)

I leave the call up to you on progressing any of this further!

deadmanoz (approved):

0xB10C:
I think this is good to go. Thanks @octaviolucca!

Also thanks to @deadmanoz and @b10claude for reviewing.

PR #482, opened 2026-08-17
add(archive): --low-data mode

octaviolucca:
Follow-up for #439.

This adds a `--low-data` mode to the archive tool. When enabled, the archiver drops large raw transaction/block payloads before writing events to disk:

- `tx` messages keep `txid` and `wtxid`, but drop raw transaction bytes
- `block` messages keep the block header/hash, but drop full transactions
- `blocktxn` messages keep the block hash and transaction ids, but drop raw transaction bytes
- compact block prefilled transactions keep transaction ids, but drop raw transaction bytes

Archives now record the selected mode in `ArchiveHeader.low_data`, so readers can distinguish older archives where the field is unset from new full-data and low-data archives.

I ran local mainnet comparisons on my node:

4h25m window:
- full-data: 88.68 MiB
- low-data: 55.79 MiB
- compressed size reduction: 37.1%

My local run did not capture `blocktxn` or compact block prefilled transactions yet. I’ll run a separate comparison for those and update this PR with the results.

CI passed on my fork:
https://github.com/octaviolucca/peer-observer/actions/runs/32035017742

### DIFF
--- a/shared/src/protobuf/archive.rs
+++ b/shared/src/protobuf/archive.rs
@@ -7,10 +7,10 @@ use std::fmt;
 include!(concat!(env!("OUT_DIR"), "/header.rs"));
 
 impl ArchiveHeader {
-    pub fn new() -> Self {
+    pub fn new(low_data: bool) -> Self {
         Self {
             created: current_timestamp(),
-            low_data: Some(false),
+            low_data: Some(low_data),
         }
     }
 
--- a/shared/src/protobuf/archive.rs
+++ b/shared/src/protobuf/archive.rs
@@ -10,16 +10,46 @@ impl ArchiveHeader {
     pub fn new() -> Self {
         Self {
             created: current_timestamp(),
+            low_data: Some(false),
         }
     }
 
     pub fn to_bytes(&self) -> Vec<u8> {
         self.encode_length_delimited_to_vec()
     }
+
+    /// Unset in archives written before the field existed, which are full-data.
+    pub fn is_low_data(&self) -> bool {
+        self.low_data.unwrap_or(false)
+    }
 }
 
 impl fmt::Display for ArchiveHeader {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "ArchiveHeader(created={})", self.created,)
+        write!(
+            f,
+            "ArchiveHeader(created={}, low_data={})",
+            self.created,
+            self.is_low_data()
+        )
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn unset_low_data_is_full_data() {
+        let header = ArchiveHeader {
+            created: 1,
+            low_data: None,
+        };
+
+        assert!(!header.is_low_data());
+        assert_eq!(
+            header.to_string(),
+            "ArchiveHeader(created=1, low_data=false)"
+        );
     }
 }
--- a/tools/archive/src/archiver/mod.rs
+++ b/tools/archive/src/archiver/mod.rs
@@ -491,3 +491,199 @@ fn format_utc_timestamp(epoch_ms: u64) -> String {
         .format(&fmt)
         .expect("timestamp formatting failed")
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    use shared::protobuf::bitcoin_primitives::{BlockHeader, PrefilledTransaction, Transaction};
+    use shared::protobuf::ebpf_extractor::message::{
+        Block, BlockTxn, CompactBlock, Inv, MessageEvent, Metadata, Tx,
+    };
+    use shared::protobuf::ebpf_extractor::Ebpf;
+
+    const TXID: [u8; 32] = [0x11; 32];
+    const WTXID: [u8; 32] = [0x22; 32];
+    const TXID_2: [u8; 32] = [0x44; 32];
+    const WTXID_2: [u8; 32] = [0x55; 32];
+    const BLOCK_HASH: [u8; 32] = [0x33; 32];
+
+    fn transaction() -> Transaction {
+        transaction_with_ids(TXID, WTXID, 0xff)
+    }
+
+    fn transaction_with_ids(txid: [u8; 32], wtxid: [u8; 32], raw_byte: u8) -> Transaction {
+        Transaction {
+            txid: txid.to_vec(),
+            wtxid: wtxid.to_vec(),
+            raw: Some(vec![raw_byte; 500]),
+        }
+    }
+
+    fn block_header() -> BlockHeader {
+        BlockHeader {
+            version: 1,
+            prev_blockhash: vec![0u8; 32],
+            merkle_root: vec![0u8; 32],
+            time: 1234,
+            bits: 5678,
+            nonce: 42,
+            hash: BLOCK_HASH.to_vec(),
+        }
+    }
+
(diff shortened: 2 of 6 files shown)


tools/archive/src/archiver/mod.rs
```diff
@@ -274,24 +286,32 @@ pub async fn run(args: Args, mut shutdown_rx: watch::Receiver<bool>) -> Result<(
...
-    let mut current_file =
-        ArchiveFile::new(&args.output_dir, &args.base_name, args.compression_level)
-            .context("creating the initial archive file")?;
+    let mut current_file = ArchiveFile::new(
+        &args.output_dir,
+        &args.base_name,
+        args.compression_level,
+        args.low_data,
+    )
+    .context("creating the initial archive file")?;
```
0xB10C, line 306: nit: At some point, **not** here and now, we probably want to have something called an ArchiveFileOption struct for passing these along.

tools/archive/src/archiver/mod.rs
```diff
@@ -338,10 +358,45 @@ pub async fn run(args: Args, mut shutdown_rx: watch::Receiver<bool>) -> Result<(
...
+        Msg::Blocktxn(blocktxn) => blocktxn
+            .transactions
+            .iter_mut()
+            .for_each(|tx| tx.raw = None),
+        Msg::Compactblock(compact_block) => compact_block
+            .transactions
+            .iter_mut()
+            .for_each(|prefilled| prefilled.tx.raw = None),
+        _ => (),
+    }
```
0xB10C, line 451: This looks like it covers everything for now, but it would be good to check on a real file what other messages are potentially large. i.e. have we forgotten anything

0xB10C (commented):
did a light, first review pass. Started testing this against my peer-observer node `alice` with: 

```
$ cargo run --bin archiver -- --messages --base-name lowdata --low-data --output-dir pr482 
and
$ cargo run --bin archiver -- --messages --base-name fulldata --output-dir pr482
```

I think it makes sense to merge #479 first and rebase this.

tools/archive/src/archiver/mod.rs
octaviolucca: I really like that idea! added to the tracking issue https://github.com/peer-observer/peer-observer/issues/439#issuecomment-5332394560

shared/src/protobuf/archive.rs
```diff
@@ -20,6 +21,16 @@ impl ArchiveHeader {
 
 impl fmt::Display for ArchiveHeader {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "ArchiveHeader(created={})", self.created,)
+        let low_data = match self.low_data {
+            Some(true) => "true",
+            Some(false) => "false",
+            // Archives written before the low_data field existed.
+            None => "unknown",
```
deadmanoz: Nit: I know it's just a display method, but because all previous "versions" were false `low_data`, shouldn't we just use "false" instead of "unknown"?

And have a method like:

```Rust
/// Unset in archives written before the field existed, which are always full-data.
pub fn is_low_data(&self) -> bool {
    self.low_data.unwrap_or(false)
}
```

tools/archive/src/archiver/mod.rs
```diff
@@ -391,10 +411,45 @@ pub async fn run(
...
+    let Some(ebpf::EbpfEvent::Message(message_event)) = ebpf.ebpf_event.as_mut() else {
+        return;
+    };
+    let Some(msg) = message_event.msg.as_mut() else {
+        return;
+    };
+
+    match msg {
+        Msg::Tx(tx) => tx.tx.raw = None,
+        Msg::Block(block) => block.transactions.clear(),
```
deadmanoz, line 441: So we're completely removing all TX information from the block and not retaining txid/wtxid?

The following 2 drop only the raw data (and keep the txid/wtxid).

0xB10C: Good point. I think we could keep txid/wtxid on block transactions too, if that doesn't add much code complexity!

octaviolucca: It doesnt add much complexity but I will test for storage space
Im running a comparison now with three modes:
- full-data
- low-data with block header only
- low-data keeping txid/wtxid for block transactions

If keeping the ids does not add much size, I will keep this version

octaviolucca: I compared the three storage modes using the exact same 31,919 block events containing 627,419 transactions with zstd level 22:

- full data: 180.2 MiB
- block header only: 2.4 MiB
- block header + txid/wtxid: 22.8 MiB

Keeping txid/wtxid makes block-event storage 9.6x larger than header-only, but it is still 87.4% smaller than retaining the full raw transactions.
@deadmanoz @0xB10C, would it be more useful to make retaining txid/wtxid for transactions inside block messages optional through a separate flag, such as --keep-block-txids?
Im going to push this version, if you have time, could you please test it?

deadmanoz: I guess my first review assumed that dropping all transaction data was a mistake, but was it intentional?

From the figures you share, seems like retaining the txid/wtxid is not very onerous, so I'd probably retain them, and also not introduce additional complexity with `--keep-block-txids`

deadmanoz: And I hadn't though of this before but I guess block transactions can be "reconstructed" after-the-fact using alternative sources?

octaviolucca: Mostly yes, but only for canonical/historical blocks, If a peer sends us a block with a different transaction set, header-only loses what we actually observed. External sources can reconstruct the canonical block, not the peer’s payload, thats why I think `--keep-block-txids` is useful as an option

tools/archive/src/archiver/mod.rs
```diff
@@ -230,6 +236,24 @@ pub struct Args {
...
     pub compression_level: u32,
+
+    /// If passed, don't archive raw transaction data.
+    /// Requires --messages and cannot be combined with other event filters.
+    /// Transactions keep their txid and wtxid. Blocks keep only their header.
+    #[arg(
+        long,
+        requires = "messages",
+        conflicts_with_all = [
+            "connections",
```
0xB10C: I think I would remove the `conflicts_with_all`part. Currently, running with `--messages --connections --low-data` errors with error: the argument '--connections' cannot be used with '--low-data'` which might be un-wanted (there are cases where recording e.g. connections AND messages make sense). I think just requiring "messages" is fine here. Also needs an update to the comment.

tools/archive/src/archiver/mod.rs
```diff
@@ -503,3 +503,199 @@ fn format_utc_timestamp(epoch_ms: u64) -> String {
...
+        let mut argv = vec!["archiver", "--output-dir", "/tmp"];
+        argv.extend_from_slice(args);
+        Args::try_parse_from(argv)
+    }
+
+    #[test]
+    fn low_data_requires_messages_and_conflicts_with_other_filters() {
+        assert!(parse_args(&["--low-data"]).is_err());
+        assert!(parse_args(&["--low-data", "--messages"]).is_ok());
+        assert!(parse_args(&["--low-data", "--messages", "--rpc"]).is_err());
```
0xB10C: can be dropped after above's change

0xB10C (commented):
Looks mostly good. Just one change I'd like to see regarding using --low-data along with other, non---message data capturing.

PR #483, opened 2026-09-02
Add --connections-with-handshakes & --addr-relay modes to archiver

0xB10C:
As discussed in https://github.com/peer-observer/peer-observer/issues/439#issuecomment-5268281438, this adds both a

- `--connections-with-handshakes` mode: records version handshakes and everything that --connections records
- `--addr-relay` mode: which records address relay

### DIFF
--- a/tools/archive/src/archiver/mod.rs
+++ b/tools/archive/src/archiver/mod.rs
@@ -242,6 +242,18 @@ pub struct Args {
     /// Transactions keep their txid and wtxid. Blocks keep only their header.
     #[arg(long, requires = "messages")]
     pub low_data: bool,
+
+    /// If passed, archive address-relay P2P messages (getaddr, addr, addrv2,
+    /// including empty addrv2).
+    /// Additive: combines freely with the other event filters.
+    #[arg(long)]
+    pub addr_relay: bool,
+
+    /// If passed, archive P2P connections with their version handshakes:
+    /// version messages (both directions) and all P2P connection events.
+    /// Additive: combines freely with the other event filters.
+    #[arg(long)]
+    pub connections_with_handshakes: bool,
 }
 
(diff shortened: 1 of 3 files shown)


tools/archive/src/archiver/mod.rs
```diff
@@ -254,7 +265,40 @@ impl Args {
...
+            addrv2: args.addr_relay,
+            version: args.connections_with_handshakes,
+        }
+    }
+
+    fn matches(&self, msg: &Msg) -> bool {
+        match msg {
+            Msg::Getaddr(_) => self.getaddr,
+            Msg::Addr(_) => self.addr,
+            Msg::Addrv2(_) => self.addrv2,
```
octaviolucca: Should we include `Msg::Emptyaddrv2(_)` here too? The extractor already preserves these events and the metrics count them by peer, so it seems unexpected for `--addr-relay` to drop them.

tools/archive/src/archiver/mod.rs
```diff
@@ -345,7 +410,7 @@ pub async fn run(
                             continue;
                         }
                     };
-                    if should_archive(&event, &args){
+                    if should_archive(&event, &args, msg_filter) {
```
octaviolucca: Small follow-up idea: could we keep the filtering policy in one place? `MessageFilter` is built from `Args`, but `should_archive` still needs both, maybe an `ArchiveFilter::from(&args)` with `filter.matches(&event)` would make this easier to follow and extend.

octaviolucca (commented):
Two comments: one about `Emptyaddrv2` being skipped by `--addr-relay`, and a small follow-up on the filtering structure

octaviolucca (approved):
ACK eb68ee6

PR #486, opened 2026-09-16
Add fuzzing

0xB10C:
This adds a basic fuzzing setup to peer-observer. Fuzzing feeds the code random and slightly broken inputs over and over to find crashes. It covers the places where we read data we do not control: P2P messages from remote peers, ring buffer entries from the eBPF program, archive files, debug.log lines, protobuf events from NATS, and the event handlers of the metrics and alerts tools.

Running the initial harness found four small bugs. They are fixed in this PR:
- a panic when displaying malformed hash fields
- a huge allocation in the p2p-extractor before the payload length was checked
- an unbounded message length in the archive reader
- and a panic on an unknown fee estimate mode

CI builds the fuzz targets on every push and pull request. Every Monday it also fuzzes each target for ten minutes, once on stable and once with AddressSanitizer on nightly, and uploads any crashing inputs.

To run it locally, cargo-fuzz is in shell.nix. From the repository root:

```
cd fuzz
cargo run --example gen_seeds
mkdir -p corpus/p2p_read_message
cargo fuzz run -s none p2p_read_message corpus/p2p_read_message seeds/p2p_read_message
```

The first command generates small valid starting inputs, which are not committed. Replace `p2p_read_message` with any name from `cargo fuzz list`. The run continues until you stop it with Ctrl+C or add `-- -max_total_time=600` for a ten minute run. Crashing inputs end up in `artifacts/<target>/` and can be replayed with the same `cargo fuzz run` command and the file path. `fuzz/README.md` has more, including how to fuzz with large inputs and with AddressSanitizer.


m4ycon:
uh fuzzing! I've always been interested to see how the setup of something like that is made. Any "how to test/fuzz" this PR? This could be valuable to be in readme, not just for me.

I'll take a look at the code at some point, if you aren't in a hurry to merge.

0xB10C:
I just updated the PR description a bit. There's also some stuff in `fuzz/README.md`. Feel free to play around.

> I'll take a look at the code at some point, if you aren't in a hurry to merge.

No, not at all. This is my first attempt at adding something like this. I will need a while to play around with it. Feel free to try it and report back. This is mostly LLM code and hasn't been tested (that's why it's a draft for now).

PR #489, opened 2026-09-17
metrics-tool: Add age, size, and terrible address metrics for the addrman

### LINKED ISSUE #440: metrics idea: median timestamp age per addrman table (new/tried) per node
while going through the addrman data in peer-observer, found it interesting that all nodes except kane (which is a tor/i2p only node) have fresher new table and staler/older tried table. (fresh as in latest timestamp)

kane's new table is much older (30 days median) than its tried table (2.6 days median). ~~so kane hasn't been receiving fresh addr gossip but is still actively connecting to peers which keeps tried fresh.~~ sorry, this was incorrect - it is expected behaviour in a tor/i2p-only node that at max 50% of new table is occupied. https://bnoc.xyz/t/addrman-age-distribution/130

it would be interesting to monitor this on grafana - median timestamp age per addrman table (new/tried) per node 


for snapshot on 2026-05-19, seeing the median timestamp age per addrman table per node as:
(unit is in days in this table)                                                     
                                                                                                                                                                            
  | peer    | new median (d) | tried median (d) |                                                                                                                           
  | ------- | -------------- | ---------------- |                                                                                                                           
  | alice   |           0.69 |             1.75 |                                                                                                                           
  | bob     |           0.62 |             1.05 |                                                                                                                           
  | frank   |           0.64 |             1.06 |                                                                                                                           
  | mike    |           4.56 |            25.96 |                                                                                                                           
  | nico    |           0.61 |             0.75 |                                                                                                                           
  | dave    |           0.49 |             0.76 |                                                                                                                           
  | erin    |           0.69 |             1.08 |                                                                                                                           
  | kane    |          29.69 |             2.65 |

0xB10C:
Adds metrics about the addresses in our addrman, as discussed in #440.

For both the new and the tried table, the metrics tool now reports:

- How old the addresses are. The age of an address is how long ago we last heard about it. Reported as five percentiles in `peerobserver_rpc_getrawaddrman_entry_age_seconds`, so the median tells you the typical age, and the 10th and 90th tell you how far the ages spread.
- How the ages are spread. `peerobserver_rpc_getrawaddrman_entry_age_seconds_bucket` counts how many addresses are younger than an hour, six hours, a day, and so on up to 90 days. This is the same view as the age distribution charts in the issue.
- How many addresses we have. `peerobserver_rpc_getrawaddrman_entries`, split by network.
- How many of them Bitcoin Core would throw away as terrible. `peerobserver_rpc_getrawaddrman_terrible_entries`, split by network. We can only check the two rules that look at the timestamp of an address, because `getrawaddrman` doesn't tell us how often we tried to connect to it. Most terrible addresses are older than 30 days, so this stays close to the real number, but it is a lower bound.
- How many addresses are dated into the future. `peerobserver_rpc_getrawaddrman_future_entries`, split by network. Once for any time in the future, and once for more than ten minutes ahead. Addresses more than ten minutes ahead shouldn't arrive over the network, so seeing any is a hint that something is off, for example a clock that doesn't match.

The ages are measured against the time the RPC extractor asked the node for its addrman, not against our own clock, because the extractor runs next to the node.

There are three new Grafana dashboards: one for the age, one for the size, and one for the terrible and future dated addresses.

Closes #440


tools/metrics/src/metrics.rs
```diff
@@ -106,6 +106,20 @@ pub const BUCKETS_ADDR_ADDRESS_TIMESTAMP_OFFSET: [f64; 26] = [
...
+    3600f64,    // 1 hour
+    21600f64,   // 6 hours
+    86400f64,   // 1 day
+    259200f64,  // 3 days
+    604800f64,  // 7 days
+    1209600f64, // 14 days
+    2592000f64, // 30 days
+    5184000f64, // 60 days
+    7776000f64, // 90 days
+];
```
0xB10C, line 122: @stratospher does this look good to you? We'd learn how man addresses per table are older than this per table. (In addition to the median, 25th, 75th, .. percentile).

tools/metrics/src/metrics.rs
```diff
@@ -106,6 +106,20 @@ pub const BUCKETS_ADDR_ADDRESS_TIMESTAMP_OFFSET: [f64; 26] = [
     16777216f64,
 ];
 
+// Buckets for the age of addrman entries in seconds. The last bucket (30 days) is
```
0xB10C: The last bucket is not 30 days

tools/metrics/src/metrics.rs
```diff
@@ -590,6 +609,11 @@ impl Metrics {
...
+        igv!(rpc_getrawaddrman_entry_age_seconds_bucket, "Number of addrman entries with an age (in seconds) of less than or equal to 'le' by table. Entries with a timestamp in the future are counted as age zero.", ["table", "le"], registry);
+        igv!(rpc_getrawaddrman_entries, "Number of entries in the addrman by table and network.", ["table", "network"], registry);
+        igv!(rpc_getrawaddrman_terrible_entries, "Number of addrman entries Bitcoin Core would consider terrible by table and network. This is a lower bound: only the timestamp based conditions (timestamp more than 10 minutes in the future or older than 30 days) can be checked, as getrawaddrman does not tell us about the connection attempts, the last try, and the last success of an entry.", ["table", "network"], registry);
+        igv!(rpc_getrawaddrman_future_entries, "Number of addrman entries with a timestamp at least 'min_offset_seconds' in the future by table and network. Entries more than 10 minutes (600 seconds) in the future are considered terrible by Bitcoin Core.", ["table", "network", "min_offset_seconds"], registry);
```
0xB10C, line 619: @stratospher anything to add in terms of metrics while we are at it?

0xB10C: I think number of unqiue addresses per network would be interesting too

stratospher: and number of duplicate ip addresses in our addrman (and  ip address:port as well)

octaviolucca (commented):
Concept ACK, this covers #440 
I left two comments, one question about zero counts on the terrible/future metrics and one suggestion for the entry age test

tools/metrics/src/lib.rs
```diff
@@ -917,6 +1026,72 @@ fn handle_rpc_event(e: &rpc::RpcEvent, state_arc: Arc<Mutex<State>>, metrics: me
...
+                }
+
+                for (network, count) in stats.distinct_addresses_per_network.iter() {
+                    metrics
+                        .rpc_getrawaddrman_distinct_addresses
+                        .with_label_values(&[table, network])
+                        .set(*count);
+                }
+
+                for (network, count) in stats.terrible_per_network.iter() {
```
octaviolucca: These loops only write networks with at least one terrible or future entry, so when the count drops to 0 the series disappears, I reproduced it with two snapshots, and the dashboards show gaps or "No data"
Is that intended, or should these read 0 for every network in entries_per_network?

tools/metrics/tests/integration.rs
```diff
@@ -3148,6 +3151,105 @@ async fn test_integration_metrics_rpc_getrawaddrman() {
     .await;
 }
 
+#[tokio::test]
+async fn test_integration_metrics_rpc_getrawaddrman_entry_age() {
+    println!("test that the getrawaddrman entry age, size, terrible and future metrics work");
+
+    let now = current_timestamp() as i64;
```
octaviolucca: `now` here and the event timestamp from `Event::new` come from two separate clock reads, so ages can be off, could we hardcode `now` and set the event timestamp from it?

I tried it locally: the percentiles come out exact, so we can assert the median #440 asks for (6300)
Also, with `now` in the past, the test would catch it if ages were ever computed from the wall clock instead of the event timestamp

PR #491, opened 2026-09-18
ebpf-extractor: perf improvements and bug fixes

0xB10C:
Small changes and fixes to the eBPF extractor, one per commit.  

### Events we were losing  

The ring buffers for connections, mempool and validation were given a size in bytes where the names suggested a page count, so each ended up a single 4 KB page holding a few dozen events. When a block connects, Bitcoin Core removes every transaction in it from the mempool at once, so most of those events were dropped. Nothing reported it, because the return value of the write was never checked.  The BPF programs now count every dropped event and the extractor warns about new ones once a minute. A single message we could not parse also used to cut a read short and make the extractor look idle; it now carries on.  

### Memory 

 The buffer sizes get rounded up to a power of two times the page size. The old values sat just above a boundary, so the four P2P message buffers asked for 580 MB and were given 1.16 GB, all of it reserved up front. They now add up to 57 MB. Buffers for turned off tracepoints are no longer created at all.  

### Correctness  
- Space taken in a ring buffer still holds whatever was written there before.   We never checked whether reading a message out of bitcoind worked, so a   failed read passed on leftovers of an older message. 
- The reason a transaction was rejected is written as 113 bytes but was read   back as 118, five bytes past the end of the event. The two have differed   since they were added. 
- The message type was read out of bitcoind twice for medium sized messages.  

### Speed  
- Events were published by spawning a task each, which could finish in any   order and had no limit on how many piled up. They now go through a queue to   one publisher, which keeps them in order and caps the backlog. 
- A message payload was copied out of the ring buffer, then copied twice more   while putting its header back on, and the decoded message was copied again.   For a block that is four copies of several megabytes. All four are gone. 
- The NATS subject is no longer formatted into a new string per event.


0xB10C:
We don't have CI for the ebpf-extractor at all. So going to work on that first before reviewing and merging this.

